Issue #174, opened 2019-10-21
standalone mode daprd and app startup coordination issue


yglcode:
## Expected Behavior
run go-sdk example client & caller according to Readme; with only exception of compiling code to binaries first.
expect caller will return logs about successful invoke, get, set etc..

## Actual Behavior
many times, caller will fail with "connection refused".

Since dapr fork two processes to run daprd and app (caller,client) concurrently, it happened that caller app starts up faster and try to connect before daprd listens at http & grpc service ports.
Use some hacking code here (https://github.com/yglcode/cli/blob/545558060b5a331050e9d639e3491505098af6ac/cmd/run.go#L107-L144) to verify/fix this scenario.

Need coordination between daprd & app startup sequence?

## Steps to Reproduce the Problem
at ubuntu18.04
check out github.com/dapr/go-sdk
compile example caller client first:
go build -mod=vendor ./example/caller
go build -mod=vendor ./example/client

run example client and caller according to Readme:
dapr run --app-id client --protocol grpc --app-port 4000 ./client
dapr run --app-id caller ./caller

yaron2:
Hey yglcode, thanks for reporting.

We'll get that issue worked out this week.

yaron2:
Actually, I've looked at your code. Do you want to PR this?

yglcode:
Sure, i can add a PR. the current change is only hack fix for standalone mode. Could the issue exist for dapr sidecar and app container in pod? Should i  just PR this change for startup? since my branch has other code for installing dapr locally at linux without sudo.

LMWF:
@yglcode - Thanks for the contribution.  I approved your linked PR for standalone.  Yes the issue could still happen in k8s because the start of containers within the same pod may be different every time.  

The best solution would be for the app to handle and retry appropriate errors.  However, `caller` here is a sample and we want to keep the same code relatively small and clear.

mukundansundar:
@yaron2 is this issue still relevant?

mukundansundar:
@yaron2

PR #177, opened 2019-10-22
quick fix for daprd/app startup coordination in standalone mode. (#174)

yglcode:
# Description
quick fix for daprd/app startup coordination in standalone mode.

## Issue reference
#174 
## Checklist

* [x ] Code compiles correctly
* [ ] Created/updated tests


cmd/run.go
```diff
@@ -107,8 +107,19 @@ var RunCmd = &cobra.Command{
 				errScanner := bufio.NewScanner(stdErrPipe)
 				outScanner := bufio.NewScanner(stdOutPipe)
 				go func() {
+					//pure client app wait till daprd ready for conn
+					start_app := false
```
yaron2: Can this be renamed to `startApp`?

cmd/run.go
```diff
@@ -107,8 +107,19 @@ var RunCmd = &cobra.Command{
...
+					ready_msg := "dapr initialized. Status: Running"
```
yaron2: Can this be renamed to `readyMsg`?

cmd/run.go
```diff
@@ -124,7 +135,11 @@ var RunCmd = &cobra.Command{
 					os.Exit(1)
 				}
 
-				daprRunning <- true
+				//if app exposes service, don't wait for daprd
```
yaron2: Can remove this if block: we still need to wait for daprd even if the app exposes a service. There's a small window between when Dapr detects the app is running and when it starts its gRPC and http server.

cmd/run.go
```diff
@@ -107,8 +107,19 @@ var RunCmd = &cobra.Command{
...
+					//pure client app wait till daprd ready for conn
+					start_app := false
+					ready_msg := "dapr initialized. Status: Running"
 					for errScanner.Scan() {
-						fmt.Printf(print.Yellow(fmt.Sprintf("== DAPR == %s\n", errScanner.Text())))
+						text := errScanner.Text()
+						fmt.Printf(print.Yellow(fmt.Sprintf("== DAPR == %s\n", text)))
+						//stop searching after app started
+						if appPort == -1 && !start_app &&
+							strings.Index(text, ready_msg) != -1 {
```
amanbha: Checking for a string from Dapr output seems fragile. If this string changes in Dapr runtime, this would break. Its hard for anyone to track these string changes as its not an explicit interface.

cmd/run.go
```diff
@@ -124,7 +135,11 @@ var RunCmd = &cobra.Command{
 					os.Exit(1)
 				}
 
-				daprRunning <- true
+				//if app exposes service, don't wait for daprd
+				if appPort != -1 {
```
amanbha: I think a better approach to this would be in Dapr runtime to start http, grpc servers early in its lifecycle.

cmd/run.go
yglcode: sure, i can rename those variables.

cmd/run.go
yglcode: cannot block app which expose services, since daprd initRuntime() will block wait for app service up, cause deadloop.

cmd/run.go
yglcode: i totally agree this is just hack. instead of this, starting http/grpc listeners earlier will be better, increase chances for conn success ( although 100% success can not quaranteed? since daprd & app(without service) start ups are not synced & nondeterministic.  right now, go-sdk, java-sdk just contain protobuf stub code; is this intended minimal design? if in future it will provide wrappers for rest api, then could add conn-retry code at app side in sdk init()/conn()?

yglcode: looks like dapr sidecar startup in pod can be solved by KEP (https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/sidecarcontainers.md). For local startup issue, i saw this dapr issue 725 (https://github.com/dapr/dapr/issues/725) for adding service health check endpoint also required for above-mentioned KEP). After health check endpoint is implemented, in dapr-cli, instead of grepping daprd output text, can use GET http://localhost:daprPort/health to verify daprd up healthy, then start app?

amanbha:
@yglcode Is this PR still valid?

cmd/run.go
artursouza: The health check API seems to be a better approach IMO. This seems to be similar to discussion in https://github.com/dapr/dapr/issues/1288

shalabhms: > The health check API seems to be a better approach IMO. This seems to be similar to discussion in [dapr/dapr#1288](https://github.com/dapr/dapr/issues/1288)

yes, #1288 solves such problems.

amanbha:
@yglcode this PR is very old and I am closing it for now to keep PRs manageable across all our repos. Feel free to reopen when you plan to work on it again.
Looking forward to your contribution, thanks!

Issue #191, opened 2019-10-31
Allow Dapr CLI use from within a Docker container


philliphoff:
Currently, the Dapr CLI makes several assumptions about the environment in which it's run (in standalone mode).  For example, it assumes it can reach the Redis service at `localhost:6379` and the placement service at `localhost:50005`.  This is because `dapr init` spins up the Redis and placement service containers explicitly mapping the container ports back to the local machine, which is fine when running `dapr` on that machine.

However, if `dapr` is being run from within *a container* running on that machine, `localhost` has a very different meaning.  Why might you do that, you ask?  It may be nice to work on a Dapr-related repo from within a [VS Code Dev Container](https://code.visualstudio.com/docs/remote/containers), in order to eliminate the impact of prerequisites needed by the repo on the local machine.  In this scenario, the applications using `dapr` are run inside a container that would be peers to the Redis and placement containers spun up by `dapr init`.  (Since there are limitations on using Docker within a Docker container, typically the container would redirect Docker commands back to the Docker daemon running on the local machine.)

There are a couple things needed to make this work:

 - `dapr init` needs a way to inject its service containers into a specific network and with well-known network aliases
 - `dapr run` needs a way to indicate the host names of the service containers within that network

If the VS Code dev container is then started within a given network, it can spin up its own instance of the Dapr service containers into that same network and, with a slightly augmented `dapr run` command, help Dapr reach those services when running applications.

Specifically, I'm suggesting:

 - Adding an optional `--network string` argument to the `dapr init` command, which represents the (pre-existing) Docker network in which to add the Redis and placement service containers.
 - Adding optional `--redis-host` and `--placement-host` arguments to the `dapr run` command, which represent the in-network host names (aliases) for those respective services.
 - (Optionally) When specifying `--network`, do *not* explicitly map the Redis and placement service container ports to the local machine (the idea being they'll be accessed only from within the network, and thus allow for completely independent and isolated Dapr development).

I have a PoC with these changes that seems to work (e.g. allows me to create a dev container for `dapr/dotnet-sdk` where I can clone/run/develop the SDK/samples without requiring any prerequisites on the local machine (save for Docker itself).  Please let me know if this seems like a reasonable thing to do (or if there are implications I'm not aware of).

amanbha:
@philliphoff yes, it would be a reasonable thing to do. You can create the PR and we can discuss changes in PR. Thanks for the contribution, looking forward to it.
cc @youngbupark @yaron2

PR #192, opened 2019-10-31
Allow installation of Dapr runtime into a specific Docker network

philliphoff:
# Description

Enables installation of the Dapr runtime into a specific Docker network (e.g. for use in [VS Code Dev Container](https://code.visualstudio.com/docs/remote/containers) scenarios). When doing so, the Dapr service containers are given network-specific container names to allow multiple runtime instances as well as distinguish between runtime instances when uninstalling.  The Dapr service containers are also given well-known network aliases to allow following `dapr run` commands made within that network to specify those, now non-`localhost` hosts

Specific changes:

 - Add `.vscode` to `.gitignore`
   - Allows user/platform specific build/debugging configuration
 - Add optional `--network` argument to `dapr init` command
   - If specified, injects the Redis and placement service containers into the specified Docker network and does **not** map container ports to the local machine
   - If omitted, preserves the existing behavior (i.e. the service containers are started without a specific network and container ports **are** mapped to the local machine)
 - Add optional `--redis-host` and `--placement-host` arguments to the `dapr run` command
   - If specified, uses the arguments as the host names for the Redis and placement service addresses, respectively, passed to the underlying `daprd` service (directly or indirectly via the component YAML files)
   - If omitted, preserves the existing behavior (i.e. both host names default to `localhost`)
 - Add optional `--network` argument to `dapr uninstall` command
   - If specified, removes the Redis and placement service containers from the specified Docker network
   - If omitted, preserves the existing behavior (i.e. the service containers not in any specific network are removed)
 - Updates `README.md` to provide examples of using these new arguments

## Issue reference

This PR resolves issue #191.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


yaron2:
From first glance everything looks great. I'll take on a full review later on. Thanks @philliphoff

utils/utils.go
```diff
@@ -67,3 +68,11 @@ func RunCmdAndWait(name string, args ...string) error {
 
 	return nil
 }
+
+func CreateContainerName(serviceContainerName string, dockerNetwork string) string {
+	if (dockerNetwork != "") {
+		return fmt.Sprintf("%s_%s", serviceContainerName, dockerNetwork)
+	} else {
+		return serviceContainerName
```
yaron2: Remove the else block

yaron2:
One comment + resolve conflict and we're good to go.

utils/utils.go
philliphoff: Updated.

yaron2:
Thanks for this @philliphoff!

Issue #193, opened 2019-11-04
dapr always generate `redis.yaml` and `redis_messagebus.yaml` files in standalone mode


flier:
## Expected Behavior

When use `dapr` in standalone mode, `daprd` should load components base on configuration.

## Actual Behavior

In standalone mode, `dapr` will always generate `redis.yaml` and `redis_messagebus.yaml` files in the components folder, but the `daprd` runtime only use the first loaded component with same type.
It means when I add some state store or pubsub with different filename, it maybe be skipped.

## Steps to Reproduce the Problem

- create a new `state store` component, such as `zookeeper.yaml`, in components folder.
- delete the auto genedrated component, `components/redis.yaml`
- run `dapr` in standalone mode, `dapr run --app-id mynode --app-port 3000 --port 3500 node app.js`
- new `redis.yaml` file was generated, and loaded as the state store in `daprd`
- my `zookeeper.yaml` was ignored in silence

philliphoff:
@flier Somewhat related, the `redis.yaml` and `redis_messagebus.yaml` files will not be regenerated if they already exist, even if what *would* be regenerated would be different (e.g. because of different Redis settings passed to the Dapr CLI).  I would hope that the Dapr CLI would at least provide a warning in that case.  This may be a scenario to account for in your related changes.

flier:
@philliphoff agree, I think we need support to load multi components with same type and different name later. For example, we may use one `redis` state store for internal temporary state, and another `etcd` state store for external shared settings.

PR #194, opened 2019-11-04
[WIP] Only generate redis state store/pubsub when missing

flier:
# Description

In standalone mode, `dapr` will always generate `redis.yaml` and `redis_messagebus.yaml` files in the components folder, but the `daprd` runtime only use the first loaded component with same type.
It means when I add some state store or pubsub with different filename, it maybe be skipped.

## Issue reference

dapr/cli#193

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


LMWF:
Thanks for your contribution!

PR #203, opened 2019-11-11
Add VS Code Dev Container configuration to repository

### LINKED ISSUE #202: Simplify development and debugging setup using a dev container.
For the relatively inexperienced Go developer (like me!), getting a development environment setup to build the Dapr CLI is non-trivial.  Making debugging work (in VS Code) is yet another not-so-straightforward task, even given the repository's [development guide](https://github.com/dapr/cli/blob/master/docs/development/development.md).  This setup also varies depending upon the platform.

If the developer is already using VS Code for editing, then one might make use of it's Dev Containers capability to open the repository within a Dev Container that has been pre-setup for building and debugging the Dapr CLI.  This could considerably reduce the barrier to entry for developers wishing to make a contribution but put off by having to first get a working build and debug environment.

philliphoff:
# Description

Adds a VS Code Dev Container definition to serve as the base for development/build/debug of the Dapr CLI.  Adds VS Code task configurations to build debug/release flavors of the CLI as well as adds a launch configuration for debugging the CLI.

The Dev Container image is based on the official Go Docker image, then layers on Docker and Docker Compose to allow testing of the CLI.  The Dev Container also runs in a private Docker network, allowing for an isolated Dapr runtime.

## Issue reference

Resolves #202.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

## Outstanding Questions

 - While one can use the Dev Container to create Go applications for testing Dapr features, is there a common set of stacks used for testing purposes that should be added to the Dev Container image (e.g. Node.js, .NET Core, Python, etc.)? 
 - Would it be preferable to pre-build the image and host it somewhere rather than require VS Code to generate it on each machine, being that it may take less time to download than to build?


yaron2:
If Dapr runs inside the container and needs to communicate with the app running outside of the container, then we need to have the container running with `--net=host` mode, no?

Otherwise the container will not be able to reach 127.0.0.1/localhost when communicating with the app.

philliphoff:
@yaron2 I guess it depends on how one ultimately tests CLI features.  If the Dapr CLI was built within a container, it seems like the simplest way is to test it in that same container (just as if one was developing on one's local machine).  In that case, a private network works well by keeping the modified Dapr isolated from any Dapr usage on the local machine.  It *does* mean that one likely needs other development stacks as part of the Dev Container image (unless one has an affinity for writing test/sample apps in Go).  Adding any or all of the Node.js, Python, or .NET Core stacks to the Dev Container image is reasonably easy.

If one wants to test that new build of the Dapr CLI with an application outside the container (say from a Dev Container running another stack, such as the `dapr-dotnetsdk`), then they could simply make the DevContainers use the same private network.  In that case, however, you'd need to be able to specify the application host in addition to the application port for the Dapr CLI.  Such an option may be a good idea in general.

if one wants to test that new build of the Dapr CLI using an application built on the local machine, then an option is to just run the built application directly from the `dist` folder, as the workspace is simply mapped into the container and therefore available to the host as well.  (Depending on the host platform, you might have to `make build` specifying the target platform.)

> I believe `--net=host` only works for Linux hosts, so it wouldn't be a good choice for a repo-wide Dev Container.

The goal of a Dev Container is to make it as easy as possible to be productive with a repository and build/debug/test its product.  It's also to isolate that development environment from the local machine which may have no or conflicting prerequisites.  It *should* try to cover the basic requirements for testing the product, but it doesn't necessarily have to cover *every* possible test scenario.

.gitignore
```diff
@@ -11,5 +11,3 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
-
-.vscode
```
artursouza: why removing this?

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
artursouza: JSON does not officially support comments. Does this work?

docs/development/development.md
```diff
@@ -3,6 +3,36 @@
...
+
+## Setup
+
+1. Clone the repository:
+
+    ```bash
+    $ git clone https://github.com/dapr/cli.git
+    ```
+
+1. Open the repository in VS Code
```
artursouza: Enumeration is always "1". Is it correct?

More importantly, can you be more specific in the instructions for the las 2 steps here. If not with commands, then with instruction on UI buttons/menus or screenshots.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,30 @@
+// For format details, see https://aka.ms/vscode-remote/devcontainer.json or the definition README at
+// https://github.com/microsoft/vscode-dev-containers/tree/master/containers/go
+{
+	"name": "Go in Docker Compose",
+	"dockerComposeFile": "docker-compose.yml",
+	"service": "docker-in-docker",
+	"workspaceFolder": "/workspace",
+
+	// Use 'settings' to set *default* container specific settings.json values on container create.
```
artursouza: Comments in JSON?

.vscode/launch.json
youngbupark: comment is not described in json spec, but node js and dotnet json parsers treat // as comment.

docs/development/development.md
youngbupark: markdown automatically increases the number :)

youngbupark:
@philliphoff Can you please follow it up this pr? Instead of creating new docker image, I would use the same docker image which dapr uses - https://hub.docker.com/r/daprio/dapr-dev/tags

philliphoff:
@youngbupark Sure, I'll do so as soon as I have a chance.

amanbha:
@philliphoff  Is this PR still valid?

philliphoff:
I haven't been able to get back to this yet, so I'll close the PR and reopen if/when I do.

Issue #200, opened 2019-11-07
No error shown if docker containers aren't running


yaron2:
## Expected Behavior

CLI should show an error if wasn't able to run Docker containers for Redis and Placement Service during `dapr init`.

## Actual Behavior

Dapr finishes the installation successfully without reporting a problem.

## Steps to Reproduce the Problem

Block access to Docker Hub through a firewall.

First reported here: https://github.com/dapr/docs/issues/232

TerraVenil:
I had the same issue on latest version cli
![image](https://user-images.githubusercontent.com/6068068/68960462-e4223980-07d8-11ea-960f-c6fd1a5587ce.png)
![image](https://user-images.githubusercontent.com/6068068/68960440-da003b00-07d8-11ea-9cf9-6ccedf992cfa.png)
When run `docker inspect`
"Error": "failed to create endpoint dapr_placement on network nat: hnsCall failed in Win32: The process cannot access the file because it is being used by another process. (0x20)"

Just follow https://github.com/docker/for-win/issues/1804#issuecomment-371555979 which fix my issue.

PR #205, opened 2019-11-13
Fix issue where cli reports success when it could not start container…

LMWF:
…s #200

# Description

This change surfaces the error and the cli will not report success in this case anymore.

Some edits came in through the formatter when I saved.

## Issue reference

Please reference the issue this PR will close: #_200_

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -137,14 +137,14 @@ func getDaprDir() (string, error) {
 func runRedis(wg *sync.WaitGroup, errorChan chan<- error, dir, version string, dockerNetwork string) {
 	defer wg.Done()
 
-	args := []string {
+	args := []string{
 		"run",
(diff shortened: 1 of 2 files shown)


pkg/standalone/standalone.go
```diff
@@ -156,13 +156,15 @@ func runRedis(wg *sync.WaitGroup, errorChan chan<- error, dir, version string, d
...
 	err := utils.RunCmdAndWait("docker", args...)
+
 	if err != nil {
 		runError := isContainerRunError(err)
 		if !runError {
 			errorChan <- parseDockerError("Redis state store", err)
 			return
+		} else {
+			errorChan <- fmt.Errorf("docker %s failed with: %v", args, err)
 		}
```
youngbupark: @LMWF doesn't it require `return` here?

PR #211, opened 2019-11-21
change send cmd to invoke

yaron2:
The CLI command to invoke a Dapr app is now called "send", which is not used anywhere else in the Dapr terminology.

This PR changes the command to "invoke" to be consistent with other docs and samples.


cmd/invoke.go
```diff
@@ -0,0 +1,46 @@
...
+var invokeAppMethod string
+var invokePayload string
+
+var InvokeCmd = &cobra.Command{
+	Use:   "invoke",
+	Short: "invoke a dapr app with an optional payload",
+	Run: func(cmd *cobra.Command, args []string) {
+		response, err := invoke.InvokeApp(invokeAppID, invokeAppMethod, invokePayload)
+		if err != nil {
+			print.FailureStatusEvent(os.Stdout, fmt.Sprintf("Error invoking app %s: %s", invokeAppID, err))
```
shalabhms: nit: message starts with lowercase

cmd/invoke.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+
+	"github.com/dapr/cli/pkg/invoke"
+	"github.com/dapr/cli/pkg/print"
+	"github.com/spf13/cobra"
+)
+
+var invokeAppID string
+var invokeAppMethod string
+var invokePayload string
```
shalabhms: nit : group them

Issue #79, opened 2019-09-16
Provide an option for non-root install.


brendandburns:
running `actions init` tries to copy something into a privileged location (`/usr/local/bin`) that's fine, but you might want to:

a) print a more useful error message (e.g. 'sudo actions init') instead of just failing the write.

b) provide a command line option for saving to an alternate location so that people w/o root can still install.

yaron2:
How about we change from from `/usr/local/bin` to `$HOME/.dapr/bin` and instruct users to add to path with:

```
export PATH=$PATH:$HOME/.dapr/bin
```

LMWF:
Combining the two suggestions - how about:
1) If `dapr init` fails to copy to the default location (`/usr/local/bin`) print a better message e.g `Cannot write to /usr/local/bin, please run sudo dapr init`.
2) Add an optional arg to `dapr init`, --install-path, to install to a user-provided location instead of `/usr/local/bin` & include output at the end to add the specified location to the path:
```
!!!dapr installed to [user-provided location], please run the following to add it to your path:

    export PATH=$PATH:[user-provided location]
```

LMWF:
Related issue: https://github.com/dapr/cli/issues/168

Issue #168, opened 2019-10-19
On Linux, permission denied even if docker is in sudoers.


chess-equality:
## Expected Behavior

`dapr init` should push through successfully.

In the documentation, only use `sudo dapr init` if you are using `sudo docker...`.

## Actual Behavior

Running `dapr init` I get:

`Error moving dapr binary to path: open /usr/local/bin/daprd: permission denied`

## Steps to Reproduce the Problem

You are using `docker...` commands and not `sudo docker...`.

Install Dapr as per Linux doc, then run `dapr init`.

psibernetic:
I ran into this issue on Solus 4 since it does not ship with a /usr/local/bin directory.

I created the directory and then granted rwx permissions to the bin.

```bash
sudo chmod ugo+rwx bin
```

It worked fine after that, does that help?

gazzu:
create a file
`sudo vim /usr/local/bin/daprd`
give the right permissions
`sudo chown <user>:<group> /usr/local/bin/daprd`
and then
`dapr init`

chess-equality:
It's all good guys, I also did that.

But the point of this issue is all of that should be automated.

LMWF:
If you install to the default /usr/local/bin, then `dapr init` must be run with sudo.  You can now specify alternate install paths if that works better instead: https://github.com/dapr/cli/pull/213

This change documents the new `install-path` arg:
https://github.com/dapr/docs/pull/262

PR #213, opened 2019-11-22
Allow alternate install path

LMWF:
# Description

Add an `--install-path` arg to allow users to install Dapr to a different dir.

Also, add more detailed message if the install to /usr/local/bin (the default) fails because of permissions.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/cli/issues/79, https://github.com/dapr/cli/issues/168

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -137,6 +137,8 @@ func getDaprDir() (string, error) {
 	return p, nil
 }
 
+// installLocation is not used, but it is present because it's required to fit the initSteps func above.  If the number of args
+// increases more, we may consider passing in a struct instead of individual args.
 func runRedis(wg *sync.WaitGroup, errorChan chan<- error, dir, version string, dockerNetwork string, installLocation string) {
 	defer wg.Done()
 
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	"github.com/docker/docker/client"
+	"github.com/fatih/color"
 
 	"github.com/briandowns/spinner"
 	"github.com/dapr/cli/pkg/print"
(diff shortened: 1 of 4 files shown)


pkg/standalone/standalone.go
```diff
@@ -134,7 +137,7 @@ func getDaprDir() (string, error) {
 	return p, nil
 }
 
-func runRedis(wg *sync.WaitGroup, errorChan chan<- error, dir, version string, dockerNetwork string) {
+func runRedis(wg *sync.WaitGroup, errorChan chan<- error, dir, version string, dockerNetwork string, installLocation string) {
```
amanbha: This function doesnt amek use of installLocation, so param can be removed.

pkg/standalone/standalone.go
```diff
@@ -191,7 +194,7 @@ func isContainerRunError(err error) bool {
 	return false
 }
 
-func runPlacementService(wg *sync.WaitGroup, errorChan chan<- error, dir, version string, dockerNetwork string) {
+func runPlacementService(wg *sync.WaitGroup, errorChan chan<- error, dir, version string, dockerNetwork string, installLocation string) {
```
amanbha: This function doesnt amek use of installLocation, so param can be removed.

pkg/standalone/standalone.go
LMWF: Unfortunately this is needed to fit the signature of the `initSteps` func.

pkg/standalone/standalone.go
LMWF: Unfortunately this is needed to fit the signature of the `initSteps` func.

pkg/standalone/standalone.go
youngbupark: nit: It would be good to leave a comment for this.

LMWF: Added

Issue #189, opened 2019-10-29
Proposal: dapr cli to fetch the logs of a running pod.


rajibmitra:
## Describe the proposal
Proposal: dapr cli to fetch the logs of a running pod.
As we discussed on the community call , we have decidede to open this issue and track the work from here. 
@youngbupark @msfussell

yaron2:
What do you envision this to look like?

How about `dapr logs <app-id>`?

rajibmitra:
Yes , this would be great. 
`dapr logs <app-id>`

rajibmitra:
@yaron2 if its ok, I would like to implement that .

yaron2:
That's more than OK, its even better :)

RicardoNiepel:
It would be great if
`dapr logs <app-id>`
could be extended with an option to fetch logs only from the app, only from the dapr sidecar or from both

yaron2:
@rajibmitra, any updates?

yaron2:
Thanks!

PR #227, opened 2020-01-07
[WIP] Added 'logs' command to dapr cli

abhirockzz:
# Description

Added `logs` command to `dapr`

```
dapr logs -h

Gets Dapr sidecar logs for an app in Kubernetes

Usage:
  dapr logs [flags]

Flags:
  -a, --app-id string      The app id for which logs are needed
  -h, --help               help for logs
      --kubernetes         Only works with a Kubernetes cluster (default true)
  -n, --namespace string   (optional) Kubernetes namespace in which your application is deployed. default value is 'default'
  -p, --pod-name string    (optional) Name of the Pod. Use this in case you have multiple app instances (Pods)
```

## Issue reference

Issue #189 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

## Usage

- `dapr logs --app-id nodeapp --kubernetes` - fetch `daprd` logs from `nodeapp`
- `dapr logs --app-id nodeapp --pod-name nodeapp-8544f8bf6b-cvz2x --kubernetes` - fetch `daprd` logs from a specific Pod (if you ahve multiple such Pods)

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -16,6 +16,7 @@ import (
 var logsAppID string
 var podName string
 var namespace string
+var k8s bool
 
 var LogsCmd = &cobra.Command{
 	Use:   "logs",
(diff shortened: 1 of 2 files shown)


yaron2:
Oh this is great!

Thanks.

The `--for` arg is kinda weird, moreover maybe we just remove it completely and provide logs for the sidecar only?

abhirockzz:
> Oh this is great!
> 
> Thanks.
> 
> The `--for` arg is kinda weird, moreover maybe we just remove it completely and provide logs for the sidecar only?

Yeah, I know 😄 Was just trying to get things working and left it as is for now. I guess we can restrict to sidecar logs only. Another scenario is - multiple instances (Pods) for an app

yaron2:
Right. So currently it's breaking after the first found pod. Let's add the ability to give a `--pod-name` to get the sidecar logs for a specific pod. Without a pod option, w'ell take the first one found.

yaron2:
Also, since this command is special to kubernetes, lets add this to enforce the kubernetes flag:

```
LogsCmd.Flags().Bool("kubernetes", true, "")
LogsCmd.MarkFlagRequired("kubernetes")
```

^ Make sure this actually works :) essentially it should force the user to specify `--kubernetes` for `logs`.

msfussell:
The log command needs to have the mandatory --kubernetes option, since this applies only to Kubernetes deployments (for now) and not standalone deployments. In the future there may also be --swarm  or --servicefabric for example

Issue #224, opened 2019-12-20
Can't seem to use invoke command to perform GET request


ryanvolum:
## Expected Behavior

Expect to be specify a GET endpoint when using invoke. For example, I currently cannot run: ```dapr invoke --app-id js -m randomNumber```, where `randomNumber` is the name of a GET endpoint (in this case in a Node app):

```js
app.get('/randomNumber', (_req, res) => {
    res.status(200).send({
        randomNumber: Math.ceil(Math.random() * 100)
    });
});
```

## Actual Behavior

I get an error indicating that it cannot find a POST endpoint by this name: 

```cmd
dapr invoke --app-id js -m randomNumber
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot POST /randomNumber</pre>
</body>
</html>

App invoked successfully
```

vsaroopchand:
Currently it only supports POST operations. Adding GET seems doable however the existing flag 'method' would need to accept the desired HTTP methods, not the actual API action. This of course poses a breaking change. I am happy to send a PR with suggestions on the desired behavior to minimize impact.

yaron2:
> Currently it only supports POST operations. Adding GET seems doable however the existing flag 'method' would need to accept the desired HTTP methods, not the actual API action. This of course poses a breaking change. I am happy to send a PR with suggestions on the desired behavior to minimize impact.

Feel free to propose a PR for this.

vsaroopchand:
You got it, I'll work on this.

vsaroopchand:
With GET added, should I also add DELETE? What about PUT & PATCH?

yaron2:
> With GET added, should I also add DELETE? What about PUT & PATCH?

Delete and Put are required, yes.

Patch is not used.

vsaroopchand:
Ok, I will send a separate PR for Delete and Put.

PR #233, opened 2020-01-13
add invokeDelete operation

vsaroopchand:
# Description

This proposal adds support for HTTP DELETE to invoke command. 

## Issue reference

Please reference the issue this PR will close: #224 

## Checklist

* [ X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/invokeDelete.go
+++ b/cmd/invokeDelete.go
@@ -1,0 +1,44 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/dapr/cli/pkg/invoke"
+	"github.com/dapr/cli/pkg/print"
+	"github.com/spf13/cobra"
+)
+
+var invokeResourceID string
+
+// invokeDeleteCmd represents the invokeDelete command
+var invokeDeleteCmd = &cobra.Command{
+	Use:   "invokeDelete",
+	Short: "Issue HTTP DELETE to Dapr app",
+	Run: func(cmd *cobra.Command, args []string) {
+		err := invoke.Delete(invokeAppID, invokeAppMethod, invokeResourceID)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/invokeDelete.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+
+		print.SuccessStatusEvent(os.Stdout, fmt.Sprintf("HTTP Delete to method %s invoked successfully", invokeAppMethod))
+	},
+}
+
+func init() {
+	invokeDeleteCmd.Flags().StringVarP(&invokeAppID, "app-id", "a", "", "the app id to invoke")
+	invokeDeleteCmd.Flags().StringVarP(&invokeAppMethod, "method", "m", "", "the method to invoke")
+	invokeDeleteCmd.Flags().StringVarP(&invokeResourceID, "id", "", "", "the resource id")
```
LMWF: What is resource id?

vsaroopchand: The ID (key) of the resource to delete. I know Dapr refers to the State ID as Key; however, ID is traditionally used in APIs to designate the target resource.

yaron2: Yeah, but we need to keep consistency and Dapr refers to these as methods..

vsaroopchand: Method is the route, what about the ID of the state to delete?

This was my intention:
http://localhost:PORT/VERSION/invoke/APPID/method/**ID**

For example, if the user wants to delete state with key "99", are you suggesting they should set method as **order/99** and do away with ID?

LMWF: Please see these docs below.  When using service invocation in dapr, only the app-id and the method name are passed in the url.  If there is data to pass it should be in the body.

https://github.com/dapr/docs/blob/master/reference/api/service_invocation.md

https://github.com/dapr/docs/blob/master/howto/invoke-and-discover-services/README.md

vsaroopchand: Sorry, I disagree. Delete REST calls do not have a body. Please answer my question: how do I supply the KEY of the state to be deleted?

vsaroopchand: I am going to close this PR and send a new one with my revisions.

PR #236, opened 2020-01-23
Resolve #221 Installation Failure

### LINKED ISSUE #221: Install fails with "parameter cannot be found that matches parameter name 'DestinationPath'"
## Expected Behavior
Running ./install/install.ps1 it should successfully install without error.

## Actual Behavior

Receive and exception due to multiple modules with the same command 'Expand-Archive'

iex : A parameter cannot be found that matches parameter name 'DestinationPath'.
At line:1 char:83
+ ... //raw.githubusercontent.com/dapr/cli/master/install/install.ps1 | iex
+                                                                       ~~~
    + CategoryInfo          : InvalidArgument: (:) [Invoke-Expression], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,Microsoft.PowerShell.Commands.InvokeExpressionCommand

If both Pscx and Microsoft.Powershell.Archive modules are loaded in powershell and run  ./install/install.ps1 
The Pscx module does not support the DestinationPath parameter and you will receive the exception

ssimontis:
If a user has both PSCX and Microsoft.Powershell.Archive installed on their machine, the Expand-Archive call defaults to PSCX's version of Expand-Archive, with using -OutputPath instead of -DestinationPath. By specifying the module name before the call, we can ensure the proper version is always called and prevent an error for some users.

This will resolve issue #221

# Description

Modified install.ps1 script to change an error that occurs for users who have Powershell Community Extensions installed.

## Issue reference

Please reference the issue this PR will close: #221

## Checklist

* [X] Code compiles correctly
* [N/A] Created/updated tests
* [N/A ] Extended the documentation

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -71,7 +71,7 @@ if (!(Test-Path $zipFilePath -PathType Leaf)) {
 
 # Extract Dapr CLI to c:\dapr
 Write-Output "Extracting $zipFilePath..."
-Expand-Archive -Force -Path $zipFilePath -DestinationPath $DaprRoot
+Microsoft.Powershell.Archive\Expand-Archive -Force -Path $zipFilePath -DestinationPath $DaprRoot
 if (!(Test-Path $DaprCliFilePath -PathType Leaf)) {
     throw "Failed to download Dapr Cli archieve - $zipFilePath"
 }


yaron2:
Thanks for this!

yaron2:
@youngbupark can you take a look?

youngbupark:
@ssimontis thanks for your contribution.

PR #237, opened 2020-01-30
add components, configurations commands

yaron2:
Allows a user to view components and configurations in a Kubernetes cluster.

`dapr components -k`
`dapr configurations -k`

/cc @msfussell


youngbupark:
I want to get rid of vendors in cli :)

yaron2:
> I want to get rid of vendors in cli :)

Please.. lets get rid of it.

PR #261, opened 2020-02-11
enable mTLS support for CLI

yaron2:
This PR reads the configuration file, if supplied, and enables mTLS on `daprd` with the default address for Sentry.


pkg/standalone/mtls_config.go
```diff
@@ -0,0 +1,8 @@
+package standalone
```
youngbupark: license is missing here.

yaron2: added

PR #269, opened 2020-02-14
Fix configuration error when running init

yaron2:


pkg/kubernetes/kubernetes.go
```diff
@@ -43,14 +43,15 @@ func Init() error {
 
 	err = utils.RunCmdAndWait("kubectl", "apply", "-f", daprManifestPath)
 	if err != nil {
-		if s != nil {
-			s.Stop()
+		err = ensureConfig(client)
```
youngbupark: if err happens not from configuration, what happens ?

yaron2: Since this is just applying Kubernetes resources, there's no other condition that can make it error out besides this.

yaron2: So, there's no way to know what caused this.

youngbupark: I think we need to file the issue to figure it out the root cause... but for now, I agree that we can take this to release as a hotfix.

yaron2: root cause of what? we know why it's failing..

yaron2: The error messages we saw were very clear: It was unable to install the configuration object because the CRD resource wasn't registered yet

yaron2: And CRDs are the only dependencies in the installation manifest (and in general..)

youngbupark: Ah ok I was confused. 

I think we need to check if the error is failing from configuration or somewhere else.

yaron2: There's no way we can check that..

youngbupark: @yaron2 we can do with his pr - https://github.com/dapr/cli/commit/8699ed9a8ab688e205657c2dc7c768dc2b627e86 :) his change basically grab all console out :)

yaron2: Its shaky.. basing on output logs..

Issue #248, opened 2020-02-06
Allow dapr cli to run alone without app


RicardoNiepel:
## Describe the proposal

For debugging purpose it's currently necessary to use dummy commands as the app parameter for starting dapr side-by-side with the app process started already from the IDE.

The dummy commands need to different based on the OS.
For Linux and MacOS:
```
dapr run --app-id testapp --app-port 3000 --port 3500 --grpc-port 5001 -- cat
```
For Windows:
```
dapr run --app-id testapp --app-port 3000 --port 3500 --grpc-port 5001 -- waitfor FOREVER
```

**To make this scenario easier it would be better to allow dapr cli to run alone without app.**

See also: dapr/java-sdk#212

philliphoff:
There are multiple platforms which do not allow debugging when forced to start the application via the `dapr` CLI, which then requires tooling to start `daprd` directly and mimic at least some of the behavior of the `dapr` CLI such as scaffolding the standalone components files.  Being able to start Dapr independently of the application (while reusing as much of the `dapr` CLI logic as possible) would be helpful.

How about something something like `dapr start` (as opposed to `dapr run`):

```bash
> dapr start --app-id my-app --app-port 5000
```

yaron2:
> There are multiple platforms which do not allow debugging when forced to start the application via the `dapr` CLI, which then requires tooling to start `daprd` directly and mimic at least some of the behavior of the `dapr` CLI such as scaffolding the standalone components files. Being able to start Dapr independently of the application (while reusing as much of the `dapr` CLI logic as possible) would be helpful.
> 
> How about something something like `dapr start` (as opposed to `dapr run`):
> 
> ```shell
> > dapr start --app-id my-app --app-port 5000
> ```

I like `dapr start`, but would like to avoid adding new commands unless we absolutely have to.
Do you think it'd be natural to just allow `dapr run` without providing the app arguments? It seems natural to me; If I provide app arguments, Dapr will launch my app. If I don't, then only daprd will execute.

bureado:
@yaron2 yes, `dapr run` was the first thing I tried and expected it to stay in the foreground with logs to `stdout`. Something like `dapr run -d` or `--detach` would mean to me run this as a service.

philliphoff:
I'm fine with folding this into `dapr run`, too, but I think it needs an explicit flag (e.g. `--detached`, `-d` as suggested above) to help distinguish between the two cases.

artursouza:
[...]

PR #276, opened 2020-02-20
Allows run without app

artursouza:
# Description

Allows run without app and daemon mode for Unix-based OS.

Run without app:
dapr run --app-id myid --port 3005 --grpc-port 50001
dapr run

## Issue reference

Please reference the issue this PR will close: #248 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -37,13 +37,28 @@ var protocol string
 
 var RunCmd = &cobra.Command{
 	Use:   "run",
-	Short: "Launches Dapr and your app side by side",
-	Args:  cobra.MinimumNArgs(1),
+	Short: "Launches Dapr and (optionally) your app side by side",
+	Long: `Runs Dapr's sidecar and (optionally) an application.
+
+Run a Java application:
+  dapr run --app-id myapp -- java -jar myapp.jar
+Run a NodeJs application that listens to port 3000:
+  dapr run --app-id myapp --app-port 3000 -- node myapp.js
+Run a Python application in the background (daemon):
+  dapr run --app-id myapp -d -- python myapp.py
+Run sidecar only:
+  dapr run --app-id myapp
+	`,
+	Args: cobra.MinimumNArgs(0),
 	PreRun: func(cmd *cobra.Command, args []string) {
 		viper.BindPFlag("placement-host", cmd.Flags().Lookup("placement-host"))
 		viper.BindPFlag("redis-host", cmd.Flags().Lookup("redis-host"))
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			fmt.Println(print.WhiteBold("WARNING: no application command found."))
+		}
+
 		uuid, err := uuid.NewRandom()
 		if err != nil {
 			print.FailureStatusEvent(os.Stdout, err.Error())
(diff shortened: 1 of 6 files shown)


yaron2:
cc @msfussell

artursouza:
I recommend we remove the detached option because it would require us to have logs persisted somewhere and rotated - it is a bigger scope (IMO) and not worth it. Users can simply invoke it with "&" and redirect stdout and stderr to their chosen destination.

cmd/run.go
```diff
@@ -229,6 +257,10 @@ func init() {
 	RunCmd.Flags().StringVarP(&protocol, "protocol", "", "http", "tells Dapr to use HTTP or gRPC to talk to the app. Default is http")
 	RunCmd.Flags().String("redis-host", "localhost", "the host on which the Redis service resides")
 	RunCmd.Flags().String("placement-host", "localhost", "the host on which the placement service resides")
+	if runtime.GOOS != "windows" {
+		// Deamon feature does not work on Windows.
+		RunCmd.Flags().BoolVarP(&isDaemon, "daemon", "d", false, "starts dapr and app in the background")
+	}
```
LMWF: New usage should be added in this method to describe new ways dapr run can be called.

artursouza: Added "Long" attribute. See new version.

cmd/run.go
```diff
@@ -34,16 +36,42 @@ var image string
...
-	Args:  cobra.MinimumNArgs(1),
+	Long: `Runs Dapr's sidecar and (optionally) an application.
+
+Run a Java application:
+  dapr run --app-id myapp -- java -jar myapp.jar
+Run a NodeJs application that listens to port 3000:
+  dapr run --app-id myapp --app-port 3000 -- node myapp.js
+Run a Python application in the background (daemon):
+  dapr run --app-id myapp -d -- python myapp.py
+  Run sidecar only:
```
philliphoff: Nit: indentation

cmd/run.go
```diff
@@ -191,7 +224,11 @@ var RunCmd = &cobra.Command{
 				PID:          os.Getpid(),
 			})
 
-			print.SuccessStatusEvent(os.Stdout, "You're up and running! Both Dapr and your app logs will appear here.\n")
+			if output.AppCMD != nil {
+				print.SuccessStatusEvent(os.Stdout, "You're up and running! Both Dapr and your app logs will appear here.\n")
+			} else {
+				print.SuccessStatusEvent(os.Stdout, "You're up and running! Dapr logs will appear here.\n")
```
philliphoff: This output makes the fact that Dapr is running in "sidecar only" mode a little subtle here.  Perhaps something more explicit in case the user simply forgot to add the application command (which I have done from time to time).  (Also a argument for an explicit "sidecar only" switch.)

cmd/run.go
```diff
@@ -34,16 +36,42 @@ var image string
 var enableProfiling bool
 var logLevel string
 var protocol string
+var isDaemon bool
 
 var RunCmd = &cobra.Command{
 	Use:   "run",
 	Short: "Launches Dapr and your app side by side",
```
philliphoff: `Launches Dapr and (optionally) your app`?

philliphoff:
Agree on removing the "detached" option; since it wouldn't be supported on Windows, we wouldn't be able to take advantage of the feature in any cross-platform tooling anyway.

cmd/run.go
artursouza: Done.

cmd/run.go
artursouza: Done.

cmd/run.go
artursouza: I've added a warning in bold white in the beginning of the run.

If we add a switch, then we need to document the case where use passes the switch and the application command. Users might get confused the other way around: way is my app not running? Oh, I forgot to remove the switch. I think we can avoid a new switch for this use case. The warning is the first thing users will see as the output for this command.

Issue #277, opened 2020-02-21
Problem with dapr kubernetes and config when delimiter found in OS path


CarlosLanderas:
## Expected Behavior

Executing the command:

dapr list --kubernetes

Should list dapr instances in kubernetes

## Actual Behavior

`multiple kubeconfigs in KUBECONFIG environment variable - C:\Users\Carlos Landeras\.kube\config
`

## Steps to Reproduce the Problem

After debugging the CLI, it looks the **:** delimiter used is the one causing the problem. This demiliter is thought to split different kube config files but Windows hard drive paths are not compatible in it's current form:

![image](https://user-images.githubusercontent.com/16475649/75016484-62067880-548b-11ea-886b-60b7f42076bc.png)

yaron2:
Closed with #279

PR #279, opened 2020-02-21
Fix client initialization when KUBECONFIG OS path has : delimiter (#277)

CarlosLanderas:
# Description

I am unable to execute --kubernetes commands in windows as it splits wrong the home path.

image

With this change, if only one delimiter is found and it belongs to the user home path and also the KUBECONFIG env var is the same as the file path join of the user home kube config, no split is executed

## Issue reference

#277 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests (I've create tests that run locally, but they have not been commited because they would access local kube files in the CI)

### DIFF
--- a/pkg/kubernetes/client.go
+++ b/pkg/kubernetes/client.go
@@ -25,7 +25,8 @@ const kubeConfigDelimiter = ":"
 
 func getConfig() (*rest.Config, error) {
 	var kubeconfig *string
-	if home := homeDir(); home != "" {
+	var home string
+	if home = homeDir(); home != "" {
 		kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
 	} else {
 		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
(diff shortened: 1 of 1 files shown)


yaron2:
Thanks for this @CarlosLanderas. Do you have a way to verify this doesn't break a unix based system by trying this on Linux?

CarlosLanderas:
> Thanks for this @CarlosLanderas. Do you have a way to verify this doesn't break a unix based system by trying this on Linux?

Sure. I have dual boot with Manjaro Linux. Let me check it ;)

CarlosLanderas:
Working in Windows and Linux

clanderas@Baal cli]$ uname -ar
Linux Baal 5.3.18-1-MANJARO #1 SMP PREEMPT Wed Dec 18 18:34:35 UTC 2019 x86_64 GNU/Linux

[clanderas@Baal cli]$ go run main.go list --kubernetes
  APP ID              APP PORT  AGE  CREATED              
  bindings-nodeapp    3000      19h  2020-02-20 17:46.25  
  bindings-pythonapp            19h  2020-02-20 17:45.42  
[clanderas@Baal cli]$

Issue #249, opened 2020-02-07
Zombie processes


RicardoNiepel:
## Expected Behavior

No dapr app zombi processes should exist and the usage of dapr-cli should be predictable.

## Actual Behavior

Sometimes some zombie processes seem to be there but not shown and coming up at a later stage.

Also the usage of ```dapr list``` returns different results without scheduling something new.

## Steps to Reproduce the Problem

There was no other terminal or process open, running ```dapr``` cli calls or something else.

As you can see ```dapr list``` does not show up the app *subscriber* and I have not scheduled it.

```ps
PS C:\Sources\dapr-java-azure-sample> dapr list
No Dapr instances found.

PS C:\Sources\dapr-java-azure-sample> dapr run --app-id invokedemo --app-port 3000 --port 3005 -- mvn exec:java -Dexec.mainClass=com.dapr.azure.sample.invoke.DemoService -Dexec.args="-p 3000"
there's already a Dapr instance running with http port 3005

PS C:\Sources\dapr-java-azure-sample> dapr list
  APP ID      HTTP PORT  GRPC PORT  APP PORT  COMMAND               AGE  CREATED              PID   
  subscriber  3005       56464      3000      mvn exec:java -D ...  1d   2020-02-05 14:17.56  9412  

PS C:\Sources\dapr-java-azure-sample> dapr stop --app-id subscriber
app stopped successfully

PS C:\Sources\dapr-java-azure-sample> dapr list
No Dapr instances found.

PS C:\Sources\dapr-java-azure-sample> dapr --version
CLI version: 0.3.0 
Runtime version: 0.3.0
```

yaron2:
Looking at the logs, this seems to work as expected.

The fact you were able to stop the Dapr app and then didn't see it on list means there was an instance running. This might happen if you close the terminal window without exiting properly, and is why we have the stop command there.

If you were unable to stop the app due to it not running but showing up in the list, then I'd say this is a zombie process.

yaron2:
Can you reproduce this on every try?

If so, can you try it with a non Java app (like one of our J's samples)?

RicardoNiepel:
@yaron2 no - it comes up unpredictable, do we have any log switch for the cli I can enable on each call for getting more information when it happens?

LMWF:
There's no switch today for the CLI process' logging, but there ought to be.  If you see it again before we have a fix can you paste the contents of %temp%\dapr-run-data.ldj ?

RicardoNiepel:
[...]

PR #289, opened 2020-03-04
Stateless version of `dapr list`.

artursouza:
# Description

This change DEPENDS on the API created on this other PR: https://github.com/dapr/dapr/pull/1184

Removed dependency on state file.
Kept delete function, so file is unused deleted on uninstall.
Use process list and command line args for daprd to display values.
Use new metadata API to keep track of App Command linked to side car (daprd).

## Issue reference

Please reference the issue this PR will close: #249 #273 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


pkg/standalone/list.go
```diff
@@ -28,31 +33,78 @@ type ListOutput struct {
...
+			}
+
+			appPort, err := strconv.Atoi(argumentsMap["--app-port"])
+			if err != nil {
+				appPort = 0
+			}
+
+			appID := argumentsMap["--app-id"]
+			appCmd := ""
+			appMetadata, err := metadata.Get(httpPort)
```
amanbha: since we are parsing the cmdline arguments used to launch daprd process. Instead of adding a http get. put on daprd, do you think adding an extra param to daprd like --comment will solve this. The comment can be anything cli wants to put there. This avoids extra api on daprd and extra ipc during the dapr list.

amanbha: > we need more info later one, we would need to keep adding more and more of these "dummy" arguments.

Sure, I was usggesting a --comment param, you decide what to put in it. No future changes needed.

> The flag parsing library in Go only works for command line arguments into Go program, it does not parse an arbitrary string (not big deal, since I assume we can find a solution to this).

daprd is a go program, so it would work.

Are there any other use cases for metadata endpoint? If there are, I am fine with this approach.

yaron2: @amanbha the metadata endpoint is currently giving an actors count for a runtime. I expect the metadata to grow with time.

amanbha: Thats for the get, To enable this, here we are adding a put for metadata, are you saying there could be more scenarios for putting metadata over http

artursouza: > > we need more info later one, we would need to keep adding more and more of these "dummy" arguments.
> 
> Sure, I was usggesting a --comment param, you decide what to put in it. No future changes needed.
> 
> > The flag parsing library in Go only works for command line arguments into Go program, it does not parse an arbitrary string (not big deal, since I assume we can find a solution to this).
> 
> daprd is a go program, so it would work.
> 
Sorry if I was not clear: it works for the command line passed into the program. Here, we are parsing an arbitrary string - is not really the same thing. I've spent some decent amount of time on this.
> Are there any other use cases for metadata endpoint? If there are, I am fine with this approach.

yaron2: > Thats for the get, To enable this, here we are adding a put for metadata, are you saying there could be more scenarios for putting metadata over http

Yes. I can see operator-like things for non k8s-platforms using this in the future.

amanbha: OK, gentlemen, I am convinced, resolving this conversation. Thanks

artursouza: Thanks! This discussion was very clarifying for myself as well.

artursouza:
Updated to use a map for all "extra" metadata entries. I called the attribute "extended".

artursouza:
Any concerns on this PR?

Issue #285, opened 2020-02-27
dapr uninstall --kubernetes show errors even if it is succeeded


youngbupark:
## Expected Behavior

dapr uninstall --kubernetes shows the error

## Actual Behavior

dapr uninstall --kubernetes must be succeeded

## Steps to Reproduce the Problem

1. dapr install --kubernetes
2. make sure dapr is installed in kubernetes
3. dapr uninstall --kuberentes

pruthvidhodda:
**Summary:** 
The issue is we are deleting a Custom Resource Definition(CRD) first and then it's associated Custom Resource (CR). But when a CRD is deleted all it's associated CR's will be deleted as well. So when we try to delete the CR later kubectl throws an error saying it couldn't find the requested resource.

**Details:**
When we do `dapr uninstall --kubernetes` we are internally calling <code>kubectl delete -f [daprManifestYaml](https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml)</code>

In our daprManifestYaml along with other resources we are creating a custom resource definition named configurations.dapr.io and also it's associated custom resource of kind Configuration (name:default).

When we do <code>kubectl delete -f [daprManifestYaml](https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml)</code> with a `-v=4` verbose flag we can see that kubectl successfully deletes the CRD `configurations.dapr.io` 
 but throws an error when deleting the resource `default`

```bash
C:\Users\prdhodda>kubectl delete -f https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml -v=4
customresourcedefinition.apiextensions.k8s.io "components.dapr.io" deleted
customresourcedefinition.apiextensions.k8s.io "configurations.dapr.io" deleted  << CRD deleted
secret "dapr-trust-bundle" deleted
secret "dapr-sidecar-injector-cert" deleted
serviceaccount "dapr-operator" deleted
clusterrole.rbac.authorization.k8s.io "secret-reader" deleted
clusterrolebinding.rbac.authorization.k8s.io "dapr-operator" deleted
clusterrolebinding.rbac.authorization.k8s.io "dapr-secret-reader" deleted
service "dapr-api" deleted
service "dapr-placement" deleted
service "dapr-sentry" deleted
service "dapr-sidecar-injector" deleted
deployment.apps "dapr-operator" deleted
deployment.apps "dapr-placement" deleted
deployment.apps "dapr-sentry" deleted
[...]

PR #291, opened 2020-03-09
Make changes for programmatically installing a custom resource

pruthvidhodda:
# Description

Make changes for programmatically installing a custom resource

## Issue reference

Please reference the issue this PR will close: [#_[285]_](https://github.com/dapr/cli/issues/285)

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


yaron2:
The default configuration is already installed in from https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml, isn't it?

youngbupark:
> The default configuration is already installed from https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml, isn't it?

this is the cherry-pick of your change in 0.4.0 release. I forgot to cherry pick it from release-0.4.0 to master :)

pruthvidhodda:
> The default configuration is already installed in from https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml, isn't it?

The default configuration won't be present in that dapr-operator.yaml anymore because of the changes made in [this PR](https://github.com/dapr/dapr/pull/1218/files).

yaron2:
> > The default configuration is already installed in from https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml, isn't it?
> 
> The default configuration won't be present in that dapr-operator.yaml anymore because of the changes made in [this PR](https://github.com/dapr/dapr/pull/1218/files).

Yes, I just caught up on the issue and the conversation there.

My only issue here is that if we remove the configuration CRD from the manifest YAML, it will break users that do `dapr init --kubernetes` with a CLI version < 0.5.0.

@youngbupark what do you think?

youngbupark:
> > > The default configuration is already installed in from https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml, isn't it?
> > 
> > 
> > The default configuration won't be present in that dapr-operator.yaml anymore because of the changes made in [this PR](https://github.com/dapr/dapr/pull/1218/files).
> 
> Yes, I just caught up on the issue and the conversation there.
> 
> My only issue here is that if we remove the configuration CRD from the manifest YAML, it will break users that do `dapr init --kubernetes` with a CLI version < 0.5.0.
> 
> @youngbupark what do you think?

The better solution is deploying yaml to github release and then cli refers to github release one. then each dapr version has its own yaml. how do you think ?

yaron2:
> > > > The default configuration is already installed in from https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml, isn't it?
> > > 
> > > 
> > > The default configuration won't be present in that dapr-operator.yaml anymore because of the changes made in [this PR](https://github.com/dapr/dapr/pull/1218/files).
> > 
> > 
> > Yes, I just caught up on the issue and the conversation there.
> > My only issue here is that if we remove the configuration CRD from the manifest YAML, it will break users that do `dapr init --kubernetes` with a CLI version < 0.5.0.
> > @youngbupark what do you think?
> 
> The better solution is deploying yaml to github release and then cli refers to github release one. then each dapr version has its own yaml. how do you think ?

If we start from this release, that would indeed be backward compatible.

yaron2:
@youngbupark if we merge this, do you have bandwidth for it?

youngbupark:
> @youngbupark if we merge this, do you have bandwidth for it?

@pruthvidhodda will do :)

Issue #292, opened 2020-03-10
Get the manifest file dapr-operator.yaml from releases


pruthvidhodda:
## Describe the proposal

Right now the dapr-operator.yaml file is getting pulled in from https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml

In order to allow different manifest files for different release versions it's decided to place the file in release assets as discussed in this [thread](https://github.com/dapr/cli/pull/291). To work along with the proposed change the CLI has to be updated to pull from correct location.

yaron2:
lgtm

PR #293, opened 2020-03-10
Update the URL to pull dapr-operator.yaml

pruthvidhodda:
# Description

Updated the URL to pull dapr-operator.yaml

## Issue reference

Please reference the issue this PR will close: [#_[292]_](https://github.com/dapr/cli/issues/292)

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -12,13 +12,11 @@ import (
 	"time"
 
 	"github.com/dapr/cli/pkg/print"
-	"github.com/dapr/cli/pkg/version"
+	"github.com/dapr/cli/pkg/standalone"
 	"github.com/dapr/cli/utils"
 
 	"github.com/briandowns/spinner"
 )
-
-var daprManifestPath string = "https://github.com/dapr/dapr/releases/download/v" + version.GetCLIVersion() + "/dapr-operator.yaml"
 
 // Init deploys the Dapr operator
 func Init() error {
@@ -26,6 +24,13 @@ func Init() error {
 	if err != nil {
 		return fmt.Errorf("can't connect to a Kubernetes cluster: %v", err)
 	}
+
+	version, err := standalone.GetLatestRelease(standalone.DaprGitHubOrg, standalone.DaprGitHubRepo)
+	if err != nil {
+		return fmt.Errorf("cannot get the manifest file: %s", err)
+	}
+
+	var daprManifestPath string = "https://github.com/dapr/dapr/releases/download/" + version + "/dapr-operator.yaml"
 
 	msg := "Deploying the Dapr Operator to your cluster..."
 	var s *spinner.Spinner
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/dapr/cli/pkg/print"
-	"github.com/dapr/cli/pkg/standalone"
 	pkg_version "github.com/dapr/cli/pkg/version"
 	"github.com/dapr/cli/utils"
 
@@ -28,7 +27,7 @@ func Init() error {
 		return fmt.Errorf("can't connect to a Kubernetes cluster: %v", err)
 	}
 
-	version, err := pkg_version.GetLatestRelease(standalone.DaprGitHubOrg, standalone.DaprGitHubRepo)
+	version, err := pkg_version.GetLatestRelease(pkg_version.DaprGitHubOrg, pkg_version.DaprGitHubRepo)
 	if err != nil {
 		return fmt.Errorf("cannot get the manifest file: %s", err)
 	}
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -12,12 +12,13 @@ import (
 	"time"
 
 	"github.com/dapr/cli/pkg/print"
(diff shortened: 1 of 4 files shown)


pkg/version/version.go
```diff
@@ -24,4 +25,26 @@ func GetRuntimeVersion() string {
 		return "n/a\n"
 	}
 	return string(out)
+
+}
+
+// GetCLIVersion returns the version for the Dapr cli.
+func GetCLIVersion() string {
+	cliName := ""
```
yaron2: I think you just need to use `RootCmd.Version`. This is what returns the `dapr --version`. You already have it inside your code..

pruthvidhodda: It's causing a cyclic dependency when I tried that.

pkg/kubernetes/kubernetes.go
```diff
@@ -12,12 +12,13 @@ import (
...
 
 	"github.com/dapr/cli/pkg/print"
+	"github.com/dapr/cli/pkg/version"
 	"github.com/dapr/cli/utils"
 
 	"github.com/briandowns/spinner"
 )
 
-const daprManifestPath = "https://daprreleases.blob.core.windows.net/manifest/dapr-operator.yaml"
+var daprManifestPath string = "https://github.com/dapr/dapr/releases/download/v" + version.GetCLIVersion() + "/dapr-operator.yaml"
```
yaron2: Try to call `RootCmd.Version` from here instead of `version.GetCLIVersion()`.

pruthvidhodda: Even that's causing cyclic dependency. We import dapr/cli/pkg/kubernetes package in dapr/cli/cmd (init.go) so we can't really import dapr/cli/cmd in dapr/cli/pkg/kubernetes package. This is based on my understanding so far but correct me if I am wrong.

yaron2: Best way is to give the version as a parameter to Init(), and then put it in the URL. Then you can use RootCmd.Version from the init command.

youngbupark: daprManifestPath needs to use the latest version of github release. GetCLIVersion() will not return what you need.

if --runtime-version is set, you need to use this version. otherwise, you need to get the latest version of cli by using below way.

https://github.com/dapr/cli/blob/eb76cbf2200058f849814649f23212b3247a952c/pkg/standalone/standalone.go#L282

pruthvidhodda: Actually we are not required to match the CLI version, we should get the latest version as suggested by @youngbupark below.

yaron2: Indeed.

pruthvidhodda: As discussed offline, we don't support --runtime-version for kubernetes mode, as of now we always pull the latest release. Made the changes.

pkg/standalone/standalone.go
```diff
@@ -112,6 +114,44 @@ func Init(runtimeVersion string, dockerNetwork string, installLocation string) e
...
+
+	for _, release := range githubRepoReleases {
+		if !strings.Contains(release.TagName, "-rc") {
+			return release.TagName, nil
+		}
+	}
+
+	return "", fmt.Errorf("no releases")
+}
+
```
youngbupark: This one is shared in k8s and standalone modes. Then, I would move this function to https://github.com/dapr/cli/blob/master/pkg/version/version.go

pruthvidhodda: That's better, moved it.

pkg/kubernetes/kubernetes.go
```diff
@@ -28,7 +27,7 @@ func Init() error {
 		return fmt.Errorf("can't connect to a Kubernetes cluster: %v", err)
 	}
 
-	version, err := pkg_version.GetLatestRelease(standalone.DaprGitHubOrg, standalone.DaprGitHubRepo)
+	version, err := pkg_version.GetLatestRelease(pkg_version.DaprGitHubOrg, pkg_version.DaprGitHubRepo)
```
youngbupark: nit: pkg_version seems weird naming :) I know someone renamed it. isn't `cli_ver` more readable ? how do you think?

pruthvidhodda: I agree with that, changed it.

pkg/kubernetes/kubernetes.go
```diff
@@ -12,22 +12,28 @@ import (
...
 	if err != nil {
 		return fmt.Errorf("can't connect to a Kubernetes cluster: %v", err)
 	}
 
+	version, err := cli_ver.GetLatestRelease(cli_ver.DaprGitHubOrg, cli_ver.DaprGitHubRepo)
+	if err != nil {
+		return fmt.Errorf("cannot get the manifest file: %s", err)
+	}
+
+	var daprManifestPath string = "https://github.com/dapr/dapr/releases/download/" + version + "/dapr-operator.yaml"
```
shalabhms: nit : Go way => daprManifestPath := "xxxxxxx"

pkg/version/version.go
```diff
@@ -6,8 +6,20 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
 	"os/exec"
 	"runtime"
+	"strings"
+)
+
+const (
+	// DaprGitHubOrg is the org name of dapr on GitHub
```
shalabhms: nit : please use "Dapr" in comment text instead of "dapr"

Issue #255, opened 2020-02-08
During dapr init, only installDaprBinary if it's actually necessary


bureado:
## Describe the proposal

In `Init()`, the `initSteps` always include `installDaprBinary`, and the logic of `installDaprBinary()` will almost always involve some form of privilege escalation in Linux since it involves creating, replacing or making `daprd` executable in a traditionally protected path such as `/usr/local/bin`.

This seems excessive, particularly for repeated executions of `dapr init`. It also prods the user to run `sudo dapr init` which in turn can trigger issues with some other commands that run during `Init()`.

The desired behavior would be for `installDaprBinary` to only be in `initSteps` if `daprd` doesn't exist or a new release is found, and `sudo` to be used only for the needed operations.

(For further details, if I switch from `docker` to `podman` and I hit this issue, then my `conmon` processes all run as `root`, even though they are perfectly capable not to.)

yaron2:
This is undocumented (will fix that soon), but there's an optional `install-path` flag to `dapr init` to have the `daprd` binary be installed to a custom location.

In any case, it would be preferable to only install the binary if its a newer version or its a first time install.

LMWF:
`--install-path` was supposed to be documented with 0.3 in https://github.com/dapr/docs/pull/262, but I accidentally closed it instead of merging it and didn't realize until now.  Reopening..

PR #299, opened 2020-03-17
Install daprd binary only if it's actually necessary

abhirockzz:
# Description

Updated to ensure `dapr init` installs `daprd` only if:

- it's a fresh install (`daprd` binary does not exist in the target location)
- the requested version is different than what's installed (requested version can be an older one)

## Issue reference

Issue #255 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -75,7 +75,7 @@ func isInstallationRequired(installLocation, requestedVersion string) bool {
 	v, err := utils.RunCmdAndWait(daprdBinaryPath, "--version")
 	if err != nil {
 		msg = fmt.Sprintf("unable to determine installed Dapr version at %s. installation will continue", destDir)
-		print.InfoStatusEvent(os.Stdout, msg)
+		fmt.Println(msg)
 		return true
 	}
 	installedVersion := strings.TrimSpace(v)
@@ -85,13 +85,13 @@ func isInstallationRequired(installLocation, requestedVersion string) bool {
 		latestVersion, err := cli_ver.GetLatestRelease(cli_ver.DaprGitHubOrg, cli_ver.DaprGitHubRepo)
 		if err != nil {
 			msg = fmt.Sprintf("latest Dapr version information could not be found - %s", err.Error())
-			print.FailureStatusEvent(os.Stdout, msg)
+			fmt.Println(msg)
 			return false
 		}
 		latestVersion = latestVersion[1:]
 		if installedVersion == latestVersion {
 			msg = fmt.Sprintf("required version %s is the same as installed version at %s", latestVersion, destDir)
-			print.InfoStatusEvent(os.Stdout, msg)
+			fmt.Println(msg)
 			return false
 		}
 	}
@@ -99,7 +99,7 @@ func isInstallationRequired(installLocation, requestedVersion string) bool {
 	// if daprd exists, need to confirm if the intended version is same as the current one
 	if installedVersion == requestedVersion {
 		msg = fmt.Sprintf("required version %s is the same as installed version at %s", requestedVersion, destDir)
-		print.InfoStatusEvent(os.Stdout, msg)
+		fmt.Println(msg)
 		return false
 	}
 
@@ -108,10 +108,6 @@ func isInstallationRequired(installLocation, requestedVersion string) bool {
 
 // Init installs Dapr on a local machine using the supplied runtimeVersion.
 func Init(runtimeVersion string, dockerNetwork string, installLocation string) error {
-	// confirm if installation is needed
-	if !isInstallationRequired(installLocation, runtimeVersion) {
-		return nil
-	}
 	dockerInstalled := isDockerInstalled()
 	if !dockerInstalled {
 		return errors.New("could not connect to Docker. Docker may not be installed or running")
@@ -331,6 +327,11 @@ func runPlacementService(wg *sync.WaitGroup, errorChan chan<- error, dir, versio
 
 func installDaprBinary(wg *sync.WaitGroup, errorChan chan<- error, dir, version string, dockerNetwork string, installLocation string) {
 	defer wg.Done()
+
+	// confirm if installation is required
+	if !isInstallationRequired(installLocation, version) {
+		return
+	}
 
 	archiveExt := "tar.gz"
 	if runtime.GOOS == daprWindowsOS {
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -20,6 +20,7 @@ import (
 	"os/exec"
 	"os/user"
 	"path"
+	"path/filepath"
 	path_filepath "path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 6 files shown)


pkg/standalone/standalone.go
```diff
@@ -50,8 +51,64 @@ const (
...
+	}
+
+	var msg string
+
+	//what's the installed version?
+	v, err := utils.RunCmdAndWait(daprdBinaryPath, "--version")
+	if err != nil {
+		msg = fmt.Sprintf("unable to determine installed Dapr version at %s", destDir)
+		print.FailureStatusEvent(os.Stdout, msg)
+		return false
```
amanbha: return true, It would be better to copy the binary in this case.

pkg/standalone/standalone.go
```diff
@@ -50,8 +51,64 @@ const (
...
+	if requestedVersion == daprLatestVersion {
+		latestVersion, err := getLatestRelease(daprGitHubOrg, daprGitHubRepo)
+		if err != nil {
+			msg = fmt.Sprintf("latest Dapr version information could not be found - %s", err.Error())
+			print.FailureStatusEvent(os.Stdout, msg)
+			return false
+		}
+		latestVersion = latestVersion[1:]
+		if installedVersion == latestVersion {
+			requestedVersion = latestVersion
```
amanbha: changing requestedVersion here doesnt seem correct and complicates the logic,, setting a bool flag here will make code easy to follow

pkg/standalone/standalone.go
```diff
@@ -50,8 +51,64 @@ const (
...
+	}
+
+	return true
+}
+
 // Init installs Dapr on a local machine using the supplied runtimeVersion.
 func Init(runtimeVersion string, dockerNetwork string, installLocation string) error {
+	//confirm if installation is needed
+	if !isInstallationRequired(installLocation, runtimeVersion) {
+		return errors.New("installation will not proceed")
```
amanbha: Rather than error, Should it just say somthing along the lines of "latest version is already installed, it will not be overwritten"

pkg/standalone/standalone.go
```diff
@@ -20,6 +20,7 @@ import (
 	"os/exec"
 	"os/user"
 	"path"
+	"path/filepath"
 	path_filepath "path/filepath"
```
pruthvidhodda: Looks like duplicate import, can we please remove this?

pkg/standalone/standalone.go
```diff
@@ -48,8 +49,64 @@ const (
...
+	destDir := daprDefaultLinuxAndMacInstallPath
+	if installLocation == "" {
+		if runtime.GOOS == daprWindowsOS {
+			destDir = daprDefaultWindowsInstallPath
+		}
+	}
+
+	daprdBinaryPath := filepath.Join(destDir, daprRuntimeFilePrefix) //e.g. /usr/local/bin/daprd or c:\\daprd
+
+	//First time install?
```
youngbupark: Nit: please add space between // and First :)

pkg/standalone/standalone.go
```diff
@@ -48,8 +49,64 @@ const (
 	DaprRedisContainerName = "dapr_redis"
 )
 
+func isInstallationRequired(installLocation, requestedVersion string) bool {
+	destDir := daprDefaultLinuxAndMacInstallPath
+	if installLocation == "" {
+		if runtime.GOOS == daprWindowsOS {
```
youngbupark: nit: I would not use nested if statement if possible :) 

```
if installLocation == "" && runtime.GOOS == daprWindowsOS {
..
```

pkg/standalone/standalone.go
```diff
@@ -48,8 +49,64 @@ const (
...
+	if err != nil {
+		msg = fmt.Sprintf("unable to determine installed Dapr version at %s", destDir)
+		print.FailureStatusEvent(os.Stdout, msg)
+		return false
+	}
+	installedVersion := strings.TrimSpace(v)
+
+	//"latest" version requested. need to check the corresponding version
+	if requestedVersion == daprLatestVersion {
+		latestVersion, err := getLatestRelease(daprGitHubOrg, daprGitHubRepo)
```
youngbupark: We moved daprGitHubOrg and daprGitHubRepo constants to pkg/version/version.go. please import version pkg.

```
import cli_ver "github.com/dapr/cli/pkg/version"
```

pkg/standalone/standalone.go
abhirockzz: Agree. It returns true and along with the `msg` as an `InfoStatusEvent`

pkg/standalone/standalone.go
abhirockzz: yes. to make this simpler, it just returns `false` along with an `Info` message

pkg/standalone/standalone.go
abhirockzz: updated: simply returns `nil` (no error). the `isInstallationRequired` function will already provide the info message to the user so explicit logging is not needed here

pkg/standalone/standalone.go
abhirockzz: thanks! fixed

pkg/standalone/standalone.go
abhirockzz: fixed :) (along with other comments as well)

pkg/standalone/standalone.go
abhirockzz: updated

pkg/standalone/standalone.go
abhirockzz: done!

pkg/standalone/standalone.go
```diff
@@ -48,8 +48,63 @@ const (
 	DaprRedisContainerName = "dapr_redis"
 )
 
+func isInstallationRequired(installLocation, requestedVersion string) bool {
+	destDir := daprDefaultLinuxAndMacInstallPath
+	if installLocation == "" && runtime.GOOS == daprWindowsOS {
+		destDir = daprDefaultWindowsInstallPath
+	}
+
+	daprdBinaryPath := path_filepath.Join(destDir, daprRuntimeFilePrefix) //e.g. /usr/local/bin/daprd or c:\\daprd
```
pruthvidhodda: What if an `installLocation` is given and it's not the default path?

amanbha: > What if an `installLocation` is given and it's not the default path?

@abhirockzz This function is not using the installLocation, should it use installLocation as destDir if provided.

abhirockzz: destDir should suffice since it was derived from the installLocation in previous step

amanbha: > destDir should suffice since it was derived from the installLocation in previous step

@abhirockzz may be I am missing something here, but DestDir is set to default daprDefaultLinuxAndMacInstallPath on line 52.

abhirockzz: Got it. Thank you! Updated to handle this

pkg/standalone/standalone.go
```diff
@@ -48,8 +48,70 @@ const (
...
+	}
+
+	return true
+}
+
 // Init installs Dapr on a local machine using the supplied runtimeVersion.
 func Init(runtimeVersion string, dockerNetwork string, installLocation string) error {
+	// confirm if installation is needed
+	if !isInstallationRequired(installLocation, runtimeVersion) {
+		return nil
```
artursouza: Init also creates the redis cluster (line 165). Should we really abort everything or just the dapr binary downloads?

amanbha: > Init also creates the redis cluster (line 165). Should we really abort everything or just the dapr binary downloads?

Good catch, yes,  its short circuiting little too early, it should just skip binary overwriting and continue with rest of the containers setup.

abhirockzz: updated to account for this

PR #303, opened 2020-03-23
Use the full container name when checking redis

### LINKED ISSUE #302: Init fails when using a specific network
## Expected Behavior

The instructions at https://github.com/dapr/cli/tree/157a87f1cd7584de3941c79a8ea82718064d2473#install-to-a-specific-docker-network describe how it is possible to configure `dapr init` to use a specific network.

The expectation is that `dapr init` would succeed when following those instructions.

## Actual Behavior

The following command shows the output with a newly-created network:

$ docker network create foobar && dapr init --network foobar
6363db0545f7dcbf56452a316a472564ecf203ff0f09ebabb2ed592f5c437f59
WARNING: could not delete run data file
↑  Downloading binaries and setting up components... Installing Dapr to /usr/local/bin
❌  container dapr_redis is not running

As a pre-requisite, check that dapr hasn't already been initialized (or rm all related docker containers) and then run:

```
$ docker network create foobar && dapr init --network foobar
```

## Additional information

This issue was observed when using the vscode devcontainer because it sets a specific network (see https://github.com/microsoft/vscode-dev-containers/blob/bc25eb057620b1299daaf5afbb0d6c310537960b/containers/dapr-dotnetcore-3.1/.devcontainer/docker-compose.yml#L22).

vjrantal:
# Description

When the redis container is created, the potential custom docker
network is appended to the container name. This change uses the
same convention when checking that the redis container is running.

## Issue reference

Fixes #302.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -100,7 +100,7 @@ func Init(runtimeVersion string, dockerNetwork string, installLocation string) e
 
 	if s != nil {
 		s.Stop()
-		err = confirmContainerIsRunning(DaprRedisContainerName)
+		err = confirmContainerIsRunning(utils.CreateContainerName(DaprRedisContainerName, dockerNetwork))
 		if err != nil {
 			return err
 		}


youngbupark:
@vjrantal Have you tested the uninstall scenario with docker network option? I saw that there is failure when we uninstall with network option.

pruthvidhodda:
> @vjrantal Have you tested the uninstall scenario with docker network option? I saw that there is failure when we uninstall with network option.

I tried uninstall with network option, it succeeded with `Dapr has been removed successfully` message.

But I see that the dapr_redis_<networkname> container still running, are you talking about that?

youngbupark:
> > @vjrantal Have you tested the uninstall scenario with docker network option? I saw that there is failure when we uninstall with network option.
> 
> I tried uninstall with network option, it succeeded with `Dapr has been removed successfully` message.
> 
> But I see that the dapr_redis_ container still running, are you talking about that?

I am talking about `dapr uninstall --all`. After that, if you still see redis container, then it is a bug.

pruthvidhodda:
That's right, I see the error with `--all` option.

vjrantal:
@youngbupark @pruthvidhodda The uninstall works but one has to give the custom network also to the uninstall command:

```
$ dapr init --network foobar
WARNING: could not delete run data file
↘  Downloading binaries and setting up components... Installing Dapr to /usr/local/bin
✅  Downloading binaries and setting up components...
✅  Success! Dapr is up and running. To get started, go here: https://aka.ms/dapr-getting-started

$ dapr uninstall --network foobar --all
ℹ️  Removing Dapr from your cluster...
WARNING: could not delete run data file
✅  Dapr has been removed successfully
```

The way this is currently implemented, it might be challenging to "guess" the custom network in the uninstall unless the end user passes that.

If uninstall must work for custom network *without* passing the custom network, what about raising a new issue and tracking the implementation of that separately from this PR?

pruthvidhodda:
> @youngbupark @pruthvidhodda The uninstall works but one has to give the custom network also to the uninstall command:

Ah, I didn't realize that. Thanks for pointing that.

> If uninstall must work for custom network _without_ passing the custom network, what about raising a new issue and tracking the implementation of that separately from this PR?

I think we shouldn't allow that as there would be a conflict when we have dapr initialized on multiple networks and we can't really default it to any particular network when the name is not provided. Thoughts @youngbupark ?

amanbha:
> > @youngbupark @pruthvidhodda The uninstall works but one has to give the custom network also to the uninstall command:
> 
> Ah, I didn't realize that. Thanks for pointing that.
> 
> > If uninstall must work for custom network _without_ passing the custom network, what about raising a new issue and tracking the implementation of that separately from this PR?
> 
> I think we shouldn't allow that as there would be a conflict when we have dapr initialized on multiple networks and we can't really default it to any particular network when the name is not provided. Thoughts @youngbupark ?

Yes, use the network name only when provided.

youngbupark:
> > I think we shouldn't allow that as there would be a conflict when we have dapr initialized on multiple networks and we can't really default it to any particular network when the name is not provided. Thoughts @youngbupark ?
> 
> Yes, use the network name only when provided.

Yes, right. I wanted to validate init and uninstall behavior with this change.

Issue #171, opened 2019-10-20
Create Homebrew package for macOS


andmos:
## Describe the proposal
👋 Hi guys, 
Creating a formula for [homebrew](https://docs.brew.sh/Formula-Cookbook) would be nice, so the CLI package can be easily kept up-to-date with `brew upgrade`, and the installer itself as easy as `brew install dapr-cli` 👍

yaron2:
Indeed!

amanbha:
On the similar lines, I would like to suggest chocolatey for Windows and Snap for Linux.

yaron2:
Yep, agreed.

andmos:
Ok so I ran
`brew create https://github.com/dapr/cli/releases/download/v0.3.0/dapr_darwin_amd64.tar.gz --set-name dapr-cli`

And got this formula generated as `/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/dapr-cli.rb`.

```ruby
class DaprCli < Formula
  desc ""
  homepage "https://dapr.io/"
  url "https://github.com/dapr/cli/releases/download/v0.3.0/dapr_darwin_amd64.tar.gz"
  sha256 "c79fd18ca04059e2c65621d9db1cc102004e3f3a4f05af96004e8da3f7634332"

  def install
    system "./configure", "--disable-debug",
                          "--disable-dependency-tracking",
                          "--disable-silent-rules",
                          "--prefix=#{prefix}"
  end

  test do

    system "false"
  end
end
```

Running it in the interactive sandbox, the binary was there and functional. Anywhere I should PR this to, or is the information enough for someone on the team to get started?

yaron2:
> Ok so I ran
> `brew create https://github.com/dapr/cli/releases/download/v0.3.0/dapr_darwin_amd64.tar.gz --set-name dapr-cli`
> 
> And got this formula generated as `/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/dapr-cli.rb`.
> 
> ```ruby
> class DaprCli < Formula
>   desc ""
>   homepage "https://dapr.io/"
>   url "https://github.com/dapr/cli/releases/download/v0.3.0/dapr_darwin_amd64.tar.gz"
>   sha256 "c79fd18ca04059e2c65621d9db1cc102004e3f3a4f05af96004e8da3f7634332"
> 
>   def install
>     system "./configure", "--disable-debug",
>                           "--disable-dependency-tracking",
[...]

PR #304, opened 2020-03-31
Adds tar.gz for Homebrew.

artursouza:
# Description

Adds tar.gz for Homebrew on Catalina. It will allow installation via binary instead of source code on Catalina macOS.

## Issue reference

Please reference the issue this PR will close: #171 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,9 @@ archive-$(CLI_BINARY).tar.gz:
 	chmod +x $(BINS_OUT_DIR)/$(CLI_BINARY)$(BINARY_EXT)
 	tar czf "$(ARCHIVE_OUT_DIR)/$(CLI_BINARY)_$(GOOS)_$(GOARCH)$(ARCHIVE_EXT)" -C "$(BINS_OUT_DIR)" "$(CLI_BINARY)$(BINARY_EXT)"
 endif
+ifeq ($(GOOS),darwin)
+	cp "$(ARCHIVE_OUT_DIR)/$(CLI_BINARY)_$(GOOS)_$(GOARCH)$(ARCHIVE_EXT)" "$(ARCHIVE_OUT_DIR)/dapr-cli-$(CLI_VERSION).catalina.bottle.tar.gz"
+endif
 
 ################################################################################
 # Target: release                                                              #


Makefile
```diff
@@ -112,6 +112,9 @@ archive-$(CLI_BINARY).tar.gz:
 	chmod +x $(BINS_OUT_DIR)/$(CLI_BINARY)$(BINARY_EXT)
 	tar czf "$(ARCHIVE_OUT_DIR)/$(CLI_BINARY)_$(GOOS)_$(GOARCH)$(ARCHIVE_EXT)" -C "$(BINS_OUT_DIR)" "$(CLI_BINARY)$(BINARY_EXT)"
 endif
+ifeq ($(GOOS),darwin)
+	cp "$(ARCHIVE_OUT_DIR)/$(CLI_BINARY)_$(GOOS)_$(GOARCH)$(ARCHIVE_EXT)" "$(ARCHIVE_OUT_DIR)/dapr-cli-$(CLI_VERSION).catalina.bottle.tar.gz"
```
shalabhms: How the tar is generated , wont we need to explicitly run "tar'' command ?

shalabhms: How about bottle metadata ? Should we need to add metadata DSL in the bottle ?

artursouza: TarGz is already generated in the command above. This copy is just because the name needs to match what Homebrew expects.

artursouza: No. That is done in a separate repo. I will create that once we merge this.

amanbha: Which separate repo, will it be checked into this repo as well?

artursouza: It will be a new repo under Dapr. It will be called homebrew-tap. This is as per the instruction from Homebrew. See artursouza/Homebrew-tap as example. It contains a rb file with all metadata needed for Homebrew to install it.

amanbha: Could you explain how the release process would look like?

artursouza: For 0.6, this change will automate the binary upload. The metadata will be manual for 0.6, which means there will be a PR that updates the version in the metadata file after a cli release.

For 0.7, I plan to have an automation to automatically update the metadata after cli release.

amanbha: Sounds good, I have marked the issue as 0.7.0 for now to track the automation as well.

PR #314, opened 2020-04-06
Add dapr control plane status for Kubernetes

### LINKED ISSUE #312: dapr status CLI command
The "dapr status" CLI command would show the status of the Dapr system services including  
1) Name of system service. e.g. Sentry service
2) Health status of the service
3) Number of service instances running 
4) Service version info

yaron2:
Closes #312 

This PR allows a user to see the Dapr control plane status when running on Kubernetes.

Command: `dapr status -k` or `dapr status --kubernetes`.

Output example: (2 services running, 2 failing):

```
  NAME                   NAMESPACE    HEALTHY  STATUS                      VERSION  AGE  CREATED              
  dapr-operator          dapr-system  False    Waiting (CrashLoopBackOff)  0.6.0    2d   2020-04-04 11:19.39  
  dapr-placement         dapr-system  True     Running                     0.6.0    2d   2020-04-04 11:19.39  
  dapr-sentry            dapr-system  False    Terminated                  0.6.0    2d   2020-04-04 11:19.39  
  dapr-sidecar-injector  dapr-system  True     Running                     0.6.0    2d   2020-04-04 11:19.39 
```

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -1,0 +1,41 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package cmd
+
+import (
+	"os"
+
+	"github.com/dapr/cli/pkg/kubernetes"
+	"github.com/dapr/cli/pkg/print"
+	"github.com/dapr/cli/utils"
+	"github.com/gocarina/gocsv"
+	"github.com/spf13/cobra"
+)
+
+var StatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Shows status and health of the Dapr control plane on Kubernetes",
+	Run: func(cmd *cobra.Command, args []string) {
+		status, err := kubernetes.Status()
+		if err != nil {
+			print.FailureStatusEvent(os.Stdout, err.Error())
+			os.Exit(1)
+		}
+		table, err := gocsv.MarshalString(status)
+		if err != nil {
+			print.FailureStatusEvent(os.Stdout, err.Error())
+			os.Exit(1)
+		}
+
+		utils.PrintTable(table)
+	},
+}
+
+func init() {
+	StatusCmd.Flags().BoolVarP(&k8s, "kubernetes", "k", true, "Only works with a Kubernetes cluster")
+	StatusCmd.MarkFlagRequired("kubernetes")
+	RootCmd.AddCommand(StatusCmd)
+}
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -35,7 +35,7 @@ var StatusCmd = &cobra.Command{
 }
 
 func init() {
-	StatusCmd.Flags().BoolVarP(&k8s, "kubernetes", "k", true, "Only works with a Kubernetes cluster")
+	StatusCmd.Flags().BoolVarP(&k8s, "kubernetes", "k", true, "only works with a Kubernetes cluster")
 	StatusCmd.MarkFlagRequired("kubernetes")
 	RootCmd.AddCommand(StatusCmd)
 }
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -17,7 +17,7 @@ import (
 
 var StatusCmd = &cobra.Command{
 	Use:   "status",
-	Short: "Shows the Dapr control plane health status.",
+	Short: "Shows the Dapr system services (control plane) health status.",
 	Run: func(cmd *cobra.Command, args []string) {
 		status, err := kubernetes.Status()
 		if err != nil {
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -17,7 +17,7 @@ import (
 
 var StatusCmd = &cobra.Command{
 	Use:   "status",
-	Short: "Shows status and health of the Dapr control plane on Kubernetes",
+	Short: "Shows the Dapr control plane health status.",
 	Run: func(cmd *cobra.Command, args []string) {
 		status, err := kubernetes.Status()
 		if err != nil {
--- a/pkg/kubernetes/status.go
+++ b/pkg/kubernetes/status.go
@@ -1,0 +1,98 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package kubernetes
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/dapr/cli/pkg/age"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+var (
+	controlPlaneLabels = []string{"dapr-operator", "dapr-sentry", "dapr-placement", "dapr-sidecar-injector"}
+)
+
+// StatutOutput represents the status of a named Dapr resource.
+type StatusOutput struct {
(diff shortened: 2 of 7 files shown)


cmd/status.go
```diff
@@ -0,0 +1,41 @@
...
+			print.FailureStatusEvent(os.Stdout, err.Error())
+			os.Exit(1)
+		}
+
+		utils.PrintTable(table)
+	},
+}
+
+func init() {
+	StatusCmd.Flags().BoolVarP(&k8s, "kubernetes", "k", true, "Only works with a Kubernetes cluster")
```
shalabhms, line 38: nit: starts with lowercase with verb first..."works only with a Kubernetes cluster"

pkg/kubernetes/status.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/dapr/cli/pkg/age"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+var (
+	controlPlaneLabels = []string{"dapr-operator", "dapr-sentry", "dapr-placement", "dapr-sidecar-injector"}
+)
+
+// StatutOutput represents the status of a named Dapr resource.
```
shalabhms, line 22: typo : should be StatusOutput

pkg/kubernetes/status.go
```diff
@@ -0,0 +1,98 @@
...
+
+	var wg sync.WaitGroup
+	wg.Add(len(controlPlaneLabels))
+
+	m := sync.Mutex{}
+	statuses := []StatusOutput{}
+
+	for _, lbl := range controlPlaneLabels {
+		go func(label string) {
+			p, err := client.CoreV1().Pods(v1.NamespaceAll).List(v1.ListOptions{
```
shalabhms, line 48: As I know dapr control plane will always be in dapr-system, so we can just pass namespace filter "dapr-system" and optimize..

yaron2: It's configurable, we can't expect it to always be on `dapr-system`.

pkg/kubernetes/status.go
yaron2: fixed

cmd/status.go
yaron2: fixed

pkg/kubernetes/status.go
```diff
@@ -0,0 +1,98 @@
...
+				LabelSelector: labels.FormatLabels(
+					map[string]string{
+						"app": label,
```
shalabhms, line 51: Could we please separate out in helper method?
p, err := getList(...)

pkg/kubernetes/status.go
shalabhms: ok. In that case, can we take the namespace as optional parameter and use that filter if provided else use NameSpace.All... ?

cmd/status.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/dapr/cli/pkg/kubernetes"
+	"github.com/dapr/cli/pkg/print"
+	"github.com/dapr/cli/utils"
+	"github.com/gocarina/gocsv"
+	"github.com/spf13/cobra"
+)
+
+var StatusCmd = &cobra.Command{
+	Use:   "status",
```
shalabhms, line 19: If it works only in Kubernetes, then I think we can avoid passing "--kubernetes " flag explicitly..

pkg/kubernetes/status.go
yaron2: The API server query is very cheap, I think the user experience here not having to specify the namespace is better here, for devs or people who are not aware of Kubernetes specific things.

cmd/status.go
yaron2: /cc @msfussell 

We took a decision a while back to allow support for other platforms in the future, and that's why we have explicit flags for platforms.

pkg/kubernetes/status.go
yaron2: Done

cmd/status.go
```diff
@@ -0,0 +1,41 @@
...
+	Short: "Shows status and health of the Dapr control plane on Kubernetes",
+	Run: func(cmd *cobra.Command, args []string) {
+		status, err := kubernetes.Status()
```
shalabhms, line 22: If we need to open this for other platforms then I think we should add a check here and for other flags(platforms) give some common response message ( status command not yet available ...? ) 
if flag=="kubernetes" { status, err := kubernetes.Status()}

yaron2: Very well, but since this is implemented in other CLI commands, this PR isn't the place to do this.

pkg/kubernetes/status.go
shalabhms: sounds good.

cmd/status.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/dapr/cli/pkg/kubernetes"
+	"github.com/dapr/cli/pkg/print"
+	"github.com/dapr/cli/utils"
+	"github.com/gocarina/gocsv"
+	"github.com/spf13/cobra"
+)
+
+var StatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Shows status and health of the Dapr control plane on Kubernetes",
```
amanbha, line 20: The command will not be kubernetes specific (its for now) so remove kubernetes from it.
placement is for local development as well, so allow it.

yaron2: We have kubernetes specific flags for many other commands. If we want to remove the flag entirely it should be done on a PR with that goal.

amanbha: yes, that can be done but this PR can do the right thing and not add Kubernetes here.
The text can just read:
"Shows status and health of the Dapr control plane."

yaron2: But that won't be the case: This PR is about Dapr control plane status on Kubernetes.

yaron2: Currently, Kubernetes is the only platform where the control plane is running.
If there are other platforms in the future, we will enable platform specific flags for those.

amanbha: yes, i understand that, and at that time, you would change this text to be:
"Shows status and health of the Dapr control plane."
You will not be adding all the platforms to it to be: "Shows status and health of the Dapr control plane on Kubernetes, Service Fabric, IoT Edge". The platform would be enforced by the flag.
So I am just asking the text to be and nothing more here:
"Shows status and health of the Dapr control plane."

yaron2: Ah ok I thought you meant something different for some reason.

yaron2: Done

msfussell: My feedback
1)  We have now invented two terms - Dapr system services and Dapr control plane and we should stick with one.
2) Does this work for dapr init -k where this is deployed to Kubernetes default namespace?
3) Will there be a PR to do this command on self hosted?

yaron2: I think we should stick with Control Plane as its where the rest of the ecosystem is in terms of terminology.

Currently we only have the Control Plane for Kubernetes: Injector and Operator are k8s only, placement is not really part of the control plane, and for sentry and placement the amount of details (like healthy/unhealthy) you can get in self-hosted mode are very scarce as opposed to Kubernetes which is designed to supply that info.

yaron2: And yes, this also works with `dapr init -k`.

yaron2:
Updated text to `system services (control plane)`. Coordinated and signed off by @msfussell

PR #316, opened 2020-04-07
Adding support for --runtime-version flag for dapr init k8s

### LINKED ISSUE #307: Support --runtime-version flag for dapr init k8s
## In what area(s)?

> /area runtime

## Describe the feature

As of today we don't have support for `--runtime-version` flag when we are installing Dapr on Kubernetes. If no --runtime-version flag is provided it should download the latest release without filtering rc releases.

shalabhms:
# Description

To support honoring the --runtime-version flag for kubernetes init

## Issue reference

Closes #307

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -20,16 +20,24 @@ import (
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// Init deploys the Dapr operator
-func Init() error {
+const (
+	daprLatestVersion = "latest"
+)
+
+// Init deploys the Dapr operator using the supplied runtime version.
+func Init(version string) error {
 	client, err := DaprClient()
 	if err != nil {
 		return fmt.Errorf("can't connect to a Kubernetes cluster: %v", err)
 	}
 
-	version, err := cli_ver.GetLatestRelease(cli_ver.DaprGitHubOrg, cli_ver.DaprGitHubRepo)
-	if err != nil {
-		return fmt.Errorf("cannot get the manifest file: %s", err)
+	if version == daprLatestVersion {
+		v, err := cli_ver.GetLatestRelease(cli_ver.DaprGitHubOrg, cli_ver.DaprGitHubRepo)
+		if err != nil {
+			return fmt.Errorf("cannot get the manifest file: %s", err)
+		}
+
+		version = v
 	}
 
 	var daprManifestPath string = "https://github.com/dapr/dapr/releases/download/" + version + "/dapr-operator.yaml"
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -20,16 +20,25 @@ import (
...
 	if err != nil {
 		return fmt.Errorf("can't connect to a Kubernetes cluster: %v", err)
 	}
 
-	version, err := cli_ver.GetLatestRelease(cli_ver.DaprGitHubOrg, cli_ver.DaprGitHubRepo)
-	if err != nil {
-		return fmt.Errorf("cannot get the manifest file: %s", err)
+	if version == daprLatestVersion {
+		var v string
+		v, err = cli_ver.GetLatestRelease(cli_ver.DaprGitHubOrg, cli_ver.DaprGitHubRepo)
```
amanbha, line 36: can it be refactored, this call should happen only when version is not latest, in all other cases it should just use the user provided version

shalabhms: > can it be refactored, this call should happen only when version is not latest, in all other cases it should just use the user provided version

The default version is always "latest" if not provided by user, so this block will get the latest release version number and use it to fetch respective latest versioned dapr-operator.yaml..

amanbha: in this case, will this call to "cli_ver.GetLatestRelease" be made when user provides a version?

pkg/kubernetes/kubernetes.go
```diff
@@ -20,16 +20,25 @@ import (
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// Init deploys the Dapr operator
-func Init() error {
+const (
+	daprLatestVersion = "latest"
```
yaron2, line 24: Can this be exported? be made into `DaprLatestVersion` and then used from `init.go`:

`InitCmd.Flags().StringVarP(&runtimeVersion, "runtime-version", "", kubernetes.DaprLatestVersion, "The version of the Dapr runtime to install. for example: v0.1.0")`

pkg/kubernetes/kubernetes.go
shalabhms: > in this case, will this call to "cli_ver.GetLatestRelease" be made when user provides a version?

no. user provided version will directly be used in fetching dapr_operator.yaml at Line 44.

pkg/kubernetes/kubernetes.go
shalabhms: @yaron2 , if we export, then standalone package need to be changed as well. Also I prefer that it is better to keep init variable (runtimeVersion) separately and get the variable value ( default ) and then pass it over in packages and define "latest" as const in kubernetes.go which is as per current PR.

yaron2: ok

Issue #300, opened 2020-03-19
Improve the error message and make uninstall work when Docker is not running


msfussell:
Docker was not running on my machine and I went to uninstall Dapr and got this confusing error message.
C:\MVP\daprapp>dapr uninstall
Removing Dapr from your cluster...
WARNING: could not delete run data file
Error removing Dapr: could not delete (dapr_placement,daprio/dapr)

 Either 
1) Provide a hint that Docker should be enabled to uninstall Dapr (like the init command below) or
2) Can you remove Dapr's containers when docker is not running (probably not)

C:\MVP\daprapp>dapr init
Making the jump to hyperspace...
WARNING: could not delete run data file
could not connect to Docker. Docker may not be installed or running

artursouza:
Alternative 1 can be added to the error message.
Alternative 2 (I assume you mean not to try to delete containers) should remain as is. The user might not be aware that docker is not running and the uninstallation would give the impression of success if we skip the container removal. Giving a clear error message (alternative 1) seems to be a better message to customers.

ksashikumar:
@msfussell @artursouza  This seems to be related to https://github.com/dapr/cli/issues/330. As I've worked on the issue, I've raised another PR for this. Could you please review? 🙂

pruthvidhodda:
Thanks for your contribution @ksashikumar :)

PR #333, opened 2020-05-14
More verbose log on failure of container uninstall

ksashikumar:
# Description

This PR adds a more verbose log on the failure of container uninstall. 

The error message would look like this:
```
➜  dapr uninstall
ℹ️  Removing Dapr from your cluster...
WARNING: could not delete run data file
❌  Error removing Dapr: uninstall failed 
 could not remove dapr_placement container: Error: No such container: dapr_placement
 
 could not remove dapr_redis container: Error: No such container: dapr_redis
```

## Issue reference

This PR will close: https://github.com/dapr/cli/issues/330

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

---
This change is [Reviewable](https://reviewable.io/reviews/dapr/cli/333)

PR #334, opened 2020-05-14
Check if docker is running on uninstall

ksashikumar:
# Description

This PR adds a check if docker is running while doing `dapr uninstall` similar to `dapr init`. 
If docker is not running, the error message would look like this: 
```
➜  dapr uninstall
ℹ️  Removing Dapr from your cluster...
WARNING: could not delete run data file
❌  Error removing Dapr: uninstall failed 
 could not connect to Docker. Docker may not be installed or running

```

If docker is running, the error message would look like this: 
```
➜  dapr uninstall
ℹ️  Removing Dapr from your cluster...
WARNING: could not delete run data file
❌  Error removing Dapr: uninstall failed 
 could not remove dapr_placement container: Error: No such container: dapr_placement
 
 could not remove dapr_redis container: Error: No such container: dapr_redis
```

## Issue reference

This PR will close: https://github.com/dapr/cli/issues/300

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

---
This change is [Reviewable](https://reviewable.io/reviews/dapr/cli/334)

### DIFF
--- a/pkg/standalone/uninstall.go
+++ b/pkg/standalone/uninstall.go
@@ -8,49 +8,53 @@ import (
 	"github.com/dapr/cli/utils"
 )
 
+func removeContainers(uninstallAll bool, dockerNetwork string) []error {
+	var containerErrs []error
+
+	_, err := utils.RunCmdAndWait(
+		"docker", "rm",
+		"--force",
+		utils.CreateContainerName(DaprPlacementContainerName, dockerNetwork))
+
+	if err != nil {
+		containerErrs = append(
+			containerErrs,
+			fmt.Errorf("could not remove %s container: %s", DaprPlacementContainerName, err))
+	}
+
+	_, err = utils.RunCmdAndWait(
+		"docker", "rmi",
+		"--force",
+		daprDockerImageName)
+
+	if err != nil {
+		containerErrs = append(
+			containerErrs,
+			fmt.Errorf("could not remove %s image: %s", daprDockerImageName, err))
+	}
+
+	if uninstallAll {
+		_, err = utils.RunCmdAndWait(
+			"docker", "rm",
+			"--force",
+			utils.CreateContainerName(DaprRedisContainerName, dockerNetwork))
+		if err != nil {
+			containerErrs = append(
+				containerErrs,
+				fmt.Errorf("could not remove %s container: %s", DaprRedisContainerName, err))
+		}
+	}
+
+	return containerErrs
+}
+
 // Uninstall deletes all installed containers
 func Uninstall(uninstallAll bool, dockerNetwork string) error {
 	var containerErrs []error
 
 	dockerInstalled := utils.IsDockerInstalled()
-	if !dockerInstalled {
-		containerErrs = append(
-			containerErrs,
-			fmt.Errorf("could not connect to Docker. Docker may not be installed or running"))
-	} else {
-		_, err := utils.RunCmdAndWait(
-			"docker", "rm",
-			"--force",
-			utils.CreateContainerName(DaprPlacementContainerName, dockerNetwork))
-
-		if err != nil {
-			containerErrs = append(
-				containerErrs,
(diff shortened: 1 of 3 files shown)


pkg/standalone/uninstall.go
```diff
@@ -10,6 +11,11 @@ import (
 
 // Uninstall deletes all installed containers
 func Uninstall(uninstallAll bool, dockerNetwork string) error {
+	dockerInstalled := utils.IsDockerInstalled()
+	if !dockerInstalled {
+		return errors.New("could not connect to Docker. Docker may not be installed or running")
```
artursouza, line 16: nit: end with punctuation.

utils/utils.go
```diff
@@ -106,3 +108,12 @@ func CreateDirectory(dir string) error {
 	}
 	return os.Mkdir(dir, 0777)
 }
+
```
artursouza, line 111: Add comment for exported method.

pkg/standalone/uninstall.go
pruthvidhodda: nit: While we are here can you please change "could not..." -> "Could not..."

LMWF: The linter doesn't like an errors.New that starts with a capital letter.

pruthvidhodda: That's right just checked the lint error. Neither should it end with punctuation. Sorry about the trouble @ksashikumar

pkg/standalone/uninstall.go
```diff
@@ -10,6 +11,11 @@ import (
 
 // Uninstall deletes all installed containers
 func Uninstall(uninstallAll bool, dockerNetwork string) error {
+	dockerInstalled := utils.IsDockerInstalled()
+	if !dockerInstalled {
```
LMWF, line 15: I think this should be best-effort in that even if docker is not there it should still perform parts of uninstall that are related to that.  In this case part unrelated is deleting a rundatafile.  So if docker is not there is skips the docker container related items, tries the other ones, then at the end fails and reports docker is not installed etc.  What do you think?

pkg/standalone/uninstall.go
```diff
@@ -10,6 +11,11 @@ import (
...
+		return errors.New("could not connect to Docker. Docker may not be installed or running.")
```
LMWF, line 16: There's a linter error here:

```
golangci-lint run --fix
pkg/standalone/uninstall.go:16:21  golint  error strings should not be capitalized or end with punctuation or a newline
make: *** [lint] Error 1
Makefile:98: recipe for target 'lint' failed
##[error]Process completed with exit code 2.
```

pkg/standalone/uninstall.go
ksashikumar: My bad! 🙁 I didn't run lint before pushing, I'll update in the next commit! 👍

pkg/standalone/uninstall.go
ksashikumar: Yeah, it makes sense! I missed `rundata.DeleteRunDataFile()`. I think this change would be even easier if https://github.com/dapr/cli/pull/333 is merged!

pkg/standalone/uninstall.go
```diff
@@ -10,47 +10,60 @@ import (
...
+			"docker", "rmi",
 			"--force",
-			utils.CreateContainerName(DaprRedisContainerName, dockerNetwork))
+			daprDockerImageName)
+
 		if err != nil {
-			errs = append(errs, fmt.Errorf("could not remove %s container: %s", DaprRedisContainerName, err))
+			containerErrs = append(
+				containerErrs,
+				fmt.Errorf("could not remove %s container: %s", daprDockerImageName, err))
```
pruthvidhodda, line 40: nit: should this message say image instead of container

ksashikumar: Good catch 👍 I've made the change and also moved the removal of containers into a method so that it's more readable.

PR #347, opened 2020-05-25
Implementing #344

cmendible:
# Description

Now it´s possible to run `dapr run` with the following flag: `--enable-dev-secretstore` which will enable the new json secret store: https://github.com/dapr/components-contrib/pull/328

## Issue reference

Please reference the issue this PR will close: #344

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/docs/reference/dapr-run.md
+++ b/docs/reference/dapr-run.md
@@ -29,4 +29,4 @@ dapr run [flags] [command]
 | `--profile-port` | | `-1` | The port for the profile server to listen on |
 | `--protocol` | | `http` | Tells Dapr to use HTTP or gRPC to talk to the app. Valid values are: `http` or `grpc` |
 | `--redis-host` | `DAPR_REDIS_HOST` | `localhost` | The host on which the Redis service resides |
-| `--enable-dev-secretstore` |  | false | Enable Developer Secret Store |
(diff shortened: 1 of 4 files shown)


docs/reference/dapr-run.md
```diff
@@ -29,3 +29,4 @@ dapr run [flags] [command]
 | `--profile-port` | | `-1` | The port for the profile server to listen on |
 | `--protocol` | | `http` | Tells Dapr to use HTTP or gRPC to talk to the app. Valid values are: `http` or `grpc` |
 | `--redis-host` | `DAPR_REDIS_HOST` | `localhost` | The host on which the Redis service resides |
+| `--enable-dev-secretstore` |  | false | Enable Developer Secret Store |
```
LMWF, line 32: Could you please add more here?  What it is, how to use it.  It's fine to link this somewhere else or have a separate issue it's going into samples/docs repo.

cmendible: @LMWF I'm going to open a PR to https://github.com/dapr/docs/tree/master/howto/setup-secret-store to explain the feature and the update with the link here. Would that be enough or do we add something else here too?

LMWF: Thanks that's good.

PR #359, opened 2020-06-04
Update README.md

yaron2:
Update with `components-path` example for `dapr run`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,12 @@ Example of launching Dapr with a node app:
 $ dapr run --app-id nodeapp node app.js
 ```
 
+Example of launching Dapr with a custom components dir:
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,6 +154,12 @@ Example of launching Dapr with a node app:
 $ dapr run --app-id nodeapp node app.js
 ```
 
+Example of launching Dapr with a custom components dir:
+
+```
+$ dapr run --app-id nodeapp --components-path ./components node app.js
```
LMWF, line 160: This should be clear that the yamls should already be where --components-path points if you use that arg.  It's easy to mistakenly interpret as "the cli will generate yamls as it used to and this will be the location".

mukundansundar:
Available in doc https://github.com/dapr/cli#use-non-default-components-path

PR #365, opened 2020-06-06
Components rc fix

vinayada1:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pkg/standalone/standalone.go
```diff
@@ -187,7 +187,12 @@ func getDownloadDest(installLocation string) (string, error) {
 		p = path.Join(usr.HomeDir, ".dapr")
 	}
 
-	err := os.MkdirAll(p, 0700)
```
vinayada1, line 190: The correct fix as suggested by @LMWF is to move copying of the default yaml files to dapr init. But the code looks too involved (package dependencies with components loader) for a last minute fix. I can address that in the next release

Issue #366, opened 2020-06-06
cli should pass components-path to daprd.


amanbha:

mukundansundar:
Would it make sense here to pass the argument all the time from the cli to runtime or only when it is overridden by the user?

amanbha:
> Would it make sense here to pass the argument all the time from the cli to runtime or only when it is overridden by the user?

it will be all the time.

yaron2:
It needs to pass the default path every time and override it with the user supplied path. That's what we also discussed on the other issue.

PR #367, opened 2020-06-06
passing components folder to daprd

amanbha:
Fixes https://github.com/dapr/cli/issues/366

### DIFF
--- a/pkg/standalone/run.go
+++ b/pkg/standalone/run.go
@@ -259,6 +259,11 @@ func Run(config *RunConfig) (*RunOutput, error) {
 		appID = strings.Replace(sillyname.GenerateStupidName(), " ", "-", -1)
 	}
 
+	_, err := os.Stat(config.ComponentsPath)
+	if err != nil {
+		return nil, err
+	}
+
 	dapr, err := List()
 	if err != nil {
 		return nil, err
@@ -270,12 +275,7 @@ func Run(config *RunConfig) (*RunOutput, error) {
 		}
 	}
 
-	componentsPath, err := getComponentsPath(config)
-	if err != nil {
-		return nil, err
-	}
-
-	componentsLoader := components.NewStandaloneComponents(modes.StandaloneConfig{ComponentsPath: componentsPath})
+	componentsLoader := components.NewStandaloneComponents(modes.StandaloneConfig{ComponentsPath: config.ComponentsPath})
 	components, err := componentsLoader.LoadComponents()
 	if err != nil {
 		return nil, err
@@ -293,20 +293,20 @@ func Run(config *RunConfig) (*RunOutput, error) {
 	}
 
 	if stateStore == "" {
-		err = createRedisStateStore(config.RedisHost, componentsPath)
+		err = createRedisStateStore(config.RedisHost, config.ComponentsPath)
 		if err != nil {
 			return nil, err
 		}
 	}
 
 	if pubSub == "" {
-		err = createRedisPubSub(config.RedisHost, componentsPath)
+		err = createRedisPubSub(config.RedisHost, config.ComponentsPath)
 		if err != nil {
 			return nil, err
 		}
 	}
 
-	daprCMD, daprHTTPPort, daprGRPCPort, metricsPort, err := getDaprCommand(appID, config.HTTPPort, config.GRPCPort, config.AppPort, config.ConfigFile, config.Protocol, config.EnableProfiling, config.ProfilePort, config.LogLevel, config.MaxConcurrency, config.PlacementHost, componentsPath)
+	daprCMD, daprHTTPPort, daprGRPCPort, metricsPort, err := getDaprCommand(appID, config.HTTPPort, config.GRPCPort, config.AppPort, config.ConfigFile, config.Protocol, config.EnableProfiling, config.ProfilePort, config.LogLevel, config.MaxConcurrency, config.PlacementHost, config.ComponentsPath)
 	if err != nil {
 		return nil, err
 	}
@@ -342,14 +342,4 @@ func Run(config *RunConfig) (*RunOutput, error) {
 		DaprHTTPPort: daprHTTPPort,
 		DaprGRPCPort: daprGRPCPort,
 	}, nil
-}
-
-func getComponentsPath(config *RunConfig) (string, error) {
-	if config.ComponentsPath == "" {
-		componentsPath := getDefaultComponentsFolder()
-		return componentsPath, nil
-	}
-
-	_, err := os.Stat(config.ComponentsPath)
-	return config.ComponentsPath, err
 }
--- a/pkg/standalone/run.go
+++ b/pkg/standalone/run.go
@@ -73,7 +73,7 @@ type componentMetadataItem struct {
 	Value string `yaml:"value"`
 }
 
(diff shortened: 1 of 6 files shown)


yaron2:
https://github.com/dapr/cli/blob/737a4e6e776466e772db9b3a1c91dd4bbdbad1bd/pkg/standalone/run.go#L47

yaron2:
Need to make sure that field gets populated with the default dir path if not supplied by the user.

pkg/standalone/run.go
```diff
@@ -73,7 +73,7 @@ type componentMetadataItem struct {
 	Value string `yaml:"value"`
 }
 
-func getDaprCommand(appID string, daprHTTPPort int, daprGRPCPort int, appPort int, configFile, protocol string, enableProfiling bool, profilePort int, logLevel string, maxConcurrency int, placementHost string) (*exec.Cmd, int, int, int, error) {
+func getDaprCommand(appID string, daprHTTPPort int, daprGRPCPort int, appPort int, configFile, protocol string, enableProfiling bool, profilePort int, logLevel string, maxConcurrency int, placementHost string, componentsPath string) (*exec.Cmd, int, int, int, error) {
```
youngbupark, line 76: is componentsPath always current directory ?

amanbha:
> Need to make sure that field gets populated with the default dir path if not supplied by the user.

Thats what the code is doing, its getting the default path 

> There's a components path field on the Config struct

Thats only set when user overrides it. Here its using the default path when not overridden by user, see here https://github.com/dapr/cli/pull/367/files#diff-184e85420a4841e42d77c23247296911R273

pkg/standalone/run.go
amanbha: its set here https://github.com/dapr/cli/pull/367/files#diff-184e85420a4841e42d77c23247296911R273

yaron2: I don't see where it is set, @amanbha. Trace the call from from cmd/run

youngbupark:
> > Need to make sure that field gets populated with the default dir path if not supplied by the user.
> 
> Thats what the code is doing, its getting the default path
> 
> > There's a components path field on the Config struct
> 
> Thats only set when user overrides it. Here its using the default path when not overridden by user, see here https://github.com/dapr/cli/pull/367/files#diff-184e85420a4841e42d77c23247296911R273

link may be wrong. https://github.com/dapr/cli/blob/c37118166b1b6c869cd2795863eb99f4960fe100/cmd/run.go#L248

this is not the current directly. 

the code always set componentsPath.. But I cannot find where componentsPath is set by default.

pkg/standalone/run.go
yaron2: We need to make sure `componentsPath` is set to the default path when it's passed to this method.

yaron2: This is missing a default:

https://github.com/dapr/cli/blob/c37118166b1b6c869cd2795863eb99f4960fe100/cmd/run.go#L248

yaron2: We're telling the user there's a default  path, but its not being set.

yaron2:
> > > Need to make sure that field gets populated with the default dir path if not supplied by the user.
> > 
> > 
> > Thats what the code is doing, its getting the default path
> > > There's a components path field on the Config struct
> > 
> > 
> > Thats only set when user overrides it. Here its using the default path when not overridden by user, see here https://github.com/dapr/cli/pull/367/files#diff-184e85420a4841e42d77c23247296911R273
> 
> link may be wrong.
> 
> https://github.com/dapr/cli/blob/c37118166b1b6c869cd2795863eb99f4960fe100/cmd/run.go#L248
> 
> this is not the current directory.
> 
> the code always set componentsPath.. But I cannot find where componentsPath is set by default.

Indeed. See my comments above.

pkg/standalone/run.go
yaron2: The default dir `getDefaultComponentsFolder()` should be set as the default value for the flag here.

amanbha:
> > > > Need to make sure that field gets populated with the default dir path if not supplied by the user.
> > > 
> > > 
> > > Thats what the code is doing, its getting the default path
> > > > There's a components path field on the Config struct
> > > 
> > > 
> > > Thats only set when user overrides it. Here its using the default path when not overridden by user, see here https://github.com/dapr/cli/pull/367/files#diff-184e85420a4841e42d77c23247296911R273
> > 
> > 
> > link may be wrong.
> > https://github.com/dapr/cli/blob/c37118166b1b6c869cd2795863eb99f4960fe100/cmd/run.go#L248
> > 
> > this is not the current directory.
> > the code always set componentsPath.. But I cannot find where componentsPath is set by default.
> 
> Indeed. See my comments above.

Adding the correct link, componentsPath is getting set here https://github.com/dapr/cli/blob/fix_366/pkg/standalone/run.go#L273

which is calling func at https://github.com/dapr/cli/blob/fix_366/pkg/standalone/run.go#L347

yaron2:
> > > > > Need to make sure that field gets populated with the default dir path if not supplied by the user.
> > > > 
> > > > 
> > > > Thats what the code is doing, its getting the default path
> > > > > There's a components path field on the Config struct
> > > > 
> > > > 
> > > > Thats only set when user overrides it. Here its using the default path when not overridden by user, see here https://github.com/dapr/cli/pull/367/files#diff-184e85420a4841e42d77c23247296911R273
> > > 
> > > 
> > > link may be wrong.
> > > https://github.com/dapr/cli/blob/c37118166b1b6c869cd2795863eb99f4960fe100/cmd/run.go#L248
> > > 
> > > this is not the current directory.
> > > the code always set componentsPath.. But I cannot find where componentsPath is set by default.
> > 
> > 
> > Indeed. See my comments above.
> 
> Adding the correct link, componentsPath is getting set here https://github.com/dapr/cli/blob/fix_366/pkg/standalone/run.go#L273
> 
> which is calling func at https://github.com/dapr/cli/blob/fix_366/pkg/standalone/run.go#L347

Can we use the CLI flags to set default values? I want to keep it consistent.

amanbha:
> > > > > > Need to make sure that field gets populated with the default dir path if not supplied by the user.
> > > > > 
> > > > > 
> > > > > Thats what the code is doing, its getting the default path
> > > > > > There's a components path field on the Config struct
> > > > > 
> > > > > 
> > > > > Thats only set when user overrides it. Here its using the default path when not overridden by user, see here https://github.com/dapr/cli/pull/367/files#diff-184e85420a4841e42d77c23247296911R273
> > > > 
> > > > 
> > > > link may be wrong.
> > > > https://github.com/dapr/cli/blob/c37118166b1b6c869cd2795863eb99f4960fe100/cmd/run.go#L248
> > > > 
> > > > this is not the current directory.
> > > > the code always set componentsPath.. But I cannot find where componentsPath is set by default.
> > > 
> > > 
> > > Indeed. See my comments above.
> > 
> > 
> > Adding the correct link, componentsPath is getting set here https://github.com/dapr/cli/blob/fix_366/pkg/standalone/run.go#L273
> > which is calling func at https://github.com/dapr/cli/blob/fix_366/pkg/standalone/run.go#L347
> 
> Can we use the CLI flags to set default values? I want to keep it consistent.

Sure, we can

yaron2:
I think you forgot to paste the function to cmd/run.go

Issue #341, opened 2020-05-19
New CLI command `dashboard`


orizohar:
Add a `dapr dashboard` command that spins up the dapr dashboard:
https://github.com/dapr/dashboard

gdhuper:
If this issue is still up for grabs, I'd like to take a stab at it. Thanks

orizohar:
> If this issue is still up for grabs, I'd like to take a stab at it. Thanks

Thanks @gdhuper for jumping on this! Can I assign this to you so we can keep track on who is working on this?

gdhuper:
Np. Yes, you can go ahead and assign this to me.

yaron2:
Thanks @gdhuper for jumping on this.

Essentially we'd like to do what the linkerd CLI does with `linkerd dashboard`, so it might be worth while to get a pointer from them :)

yaron2:
Reference: https://github.com/linkerd/linkerd2/blob/master/cli/cmd/dashboard.go

gdhuper:
Thanks for the link @yaron2. This is a great starting point. I will post questions / discussions as I develop a better understanding of this issue/cli code.

gdhuper:
@yaron2  @orizohar @youngbupark Here is a draft of what the help prompt might look like for `dapr dashboard -h`: 
```bash
$ dapr dashboard -h
Launches Dapr dashboard service to view information about running applications

Usage:
  dapr dashboard [flags]

Flags:
  -a, --address string   the address at which to serve the dashboard (default "127.0.0.1")
  -h, --help             help for dashboard
  -o, --open             opens dashboard in the browser
  -p, --port int         the HTTP port for the dashboard to listen on (default 8080)

```
After doing initial research on `linkerd dashboard` and `kubectl proxy`, these were some of the common flags provided with the dashboard command. However, if I am not mistaken dapr/dashboard does not currently support --port or --address. So I was wondering what flags do we wanna support for the first iteration or if there are any plans to extend dapr/dashboard to allow configuring port and address?

orizohar:
Thanks for doing this research @gdhuper . I think for this initial implementation of the command we can stick to what is already supported (default port and address). Moving forward if we need to ad additional functionality to the dashboard we can do so.

gdhuper:
> Thanks for doing this research @gdhuper . I think for this initial implementation of the command we can stick to what is already supported (default port and address). Moving forward if we need to ad additional functionality to the dashboard we can do so.

[...]

PR #383, opened 2020-06-25
Added new CLI command dapr dashboard

gdhuper:
# Description

Extended dapr cli to launch dapr dashboard in kubernetes.
```
Runs the Dapr dashboard on a Kubernetes cluster

Usage:
  dapr dashboard [flags]

Flags:
  -h, --help         help for dashboard
  -k, --kubernetes   Start Dapr dashboard in local browser
  -p, --port int     The local port on which to serve dashboard (default 8080)
```

## Issue reference

This PR will close: #341 

## Checklist

* [x] Code compiles correctly
* [x] Added code comments
* [ ] Created/updated tests


pkg/kubernetes/portforward.go
```diff
@@ -0,0 +1,139 @@
...
+	"time"
+
+	core_v1 "k8s.io/api/core/v1"
+	k8s "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+)
+
+type PortForward struct {
```
shalabhms, line 24: Please add comments for exported methods. It is applicable for other methods.

pkg/kubernetes/portforward.go
```diff
@@ -0,0 +1,139 @@
...
+	ReadyCh    chan struct{}
+}
+
+func NewPortForward(
+	config *rest.Config,
+	namespace, deployName string,
+	host string, localPort, remotePort int,
+	emitLogs bool,
+) (*PortForward, error) {
+	time.Sleep(10 * time.Second)
```
shalabhms, line 42: I think this sleep is not needed here, almost it is quite long 10 secs.

pkg/kubernetes/dashboard.go
```diff
@@ -0,0 +1,50 @@
...
+	"time"
+
+	"github.com/briandowns/spinner"
+	"github.com/dapr/cli/pkg/print"
+	"github.com/dapr/cli/utils"
+)
+
+func InitDashboard() error {
+
+	var dashboardManifestPath string = "https://raw.githubusercontent.com/dapr/dashboard/master/deploy/dashboard.yaml"
```
shalabhms, line 21: It would be better to create const and use that const.

cmd/dashboard.go
```diff
@@ -0,0 +1,104 @@
...
+			config,
+			meta_v1.NamespaceDefault,
+			dashboardSvc,
+			defaultHost,
+			defaultPort,
+			defaultPort,
+			false,
+		)
+		if err != nil {
+			print.FailureStatusEvent(os.Stderr, "Failed to initialize port forwarding: %s\n", err)
```
shalabhms, line 65: It would be better if either you can change the error message here to "creating.." or change Line 70 to "error in port forwarding.."
because both messages conveying the same message for 2 different related things..

cmd/dashboard.go
```diff
@@ -0,0 +1,104 @@
...
+			}
+		}
+
+		<-portForward.GetStop()
+	},
+}
+
+func init() {
+	DashboardCmd.Flags().BoolVarP(&kubernetesMode, "kubernetes", "k", false, "Deploy Dapr dashboard to a Kubernetes cluster")
+	DashboardCmd.Flags().BoolVarP(&open, "open", "o", false, "Open Dapr dashboard in a browser")
```
shalabhms, line 100: Change "open" to "start" .

shalabhms:
@gdhuper , will you also be considering in your changes for invoking dashboard in standalone mode ?

pkg/kubernetes/portforward.go
gdhuper: Agree, 10 secs feel a bit long. I had to introduce this 10 secs delay because portforwarding would fail since the pod takes couple of seconds to be up and running. Any ideas how to do this the right way using client API?

willdavsmith: Just looked into it, does this help?

https://gianarb.it/blog/programmatically-kube-port-forward-in-go

yaron2: Yes, that's the way to do it. Sleep is not needed.

Generally, we should only connect if the dashboard pod is in Ready status.

yaron2:
> @gdhuper , will you also be considering in your changes for invoking dashboard in standalone mode ?

We discussed this in the issue, this PR will tackle Kubernetes first. Once we agree on a packaging and distribution format of the dashboard for self-hosted mode, we can later enable it in the CLI.

yaron2:
This overall looks great!

Yesterday in a maintainers meeting, the following decisions were made:

1. The dashboard will be installed by default with Helm
2. The dashboard will be installed by default with `dapr init`.

Thus, I think we can remove the deployment of the dashboard from this PR, and have `dapr dashboard -k` open the browser locally IF the dashboard pod is located in the cluster and is in a Ready state.

gdhuper:
> @gdhuper , will you also be considering in your changes for invoking dashboard in standalone mode ?

yeah so that was the plan initially to support both standalone and kubernetes mode, but turns out standalone mode is a bit tricky with [containers](https://github.com/dapr/dashboard/issues/13#issuecomment-635637923). However, as Yaron mentioned in this [comment](https://github.com/dapr/cli/issues/341#issuecomment-648310794), we can download dashboard as a zip and spin it up in a local dir. IIRC, dapr/dashboard does not have a release pipeline. Any plans on setting one up for dashboard so standalone mode could be supported?

shalabhms:
> This overall looks great!
> 
> Yesterday in a maintainers meeting, the following decisions were made:
> 
> 1. The dashboard will be installed by default with Helm
> 2. The dashboard will be installed by default with `dapr init`.
> 
> Thus, I think we can remove the deployment of the dashboard from this PR, and have `dapr dashboard -k` open the browser locally IF the dashboard pod is located in the cluster and is in a Ready state.
> 
> /cc @gdhuper @shalabhms @willdavsmith

@yaron2 , that's right. We can remove deployment from `dapr dashboard --k" but we need to have similar changes in `dapr init` for kubernetes also for development mode. I think @gdhuper can still bring changes directly in `dapr init` code path.
Thoughts ?

shalabhms:
> > This overall looks great!
> > Yesterday in a maintainers meeting, the following decisions were made:
> > 
> > 1. The dashboard will be installed by default with Helm
> > 2. The dashboard will be installed by default with `dapr init`.
> > 
> > Thus, I think we can remove the deployment of the dashboard from this PR, and have `dapr dashboard -k` open the browser locally IF the dashboard pod is located in the cluster and is in a Ready state.
> > /cc @gdhuper @shalabhms @willdavsmith
> 
> @yaron2 , that's right. We can remove deployment from `dapr dashboard --k` but we need to have similar changes in `dapr init` for kubernetes also for development mode. I think @gdhuper can still bring changes directly in `dapr init` code path.
> Thoughts ?

@gdhuper ,  `dapr init -k` will take care of deploying dashboard.yaml. So yes, you can remove dashboard deployment from this PR. 

Thanks @yaron2 for clarification.

Issue #175, opened 2019-10-21
Initialize Dapr without docker


youngbupark:
## Describe the proposal

This is the issue to track the [Gitter conversation](https://gitter.im/Dapr/community?at=5dab95a50ac62f4acda2873b). 

This is the proposal to run `dapr init` without docker or provide the document to run dapr without docker.

orizohar:
Per recent discussion in the dapr core team, this will be implemented via an addition of an optional flag to allow `dapr init` to run while skipping docker container init steps, by so removing dependency to docker for users who wish to avoid it.

mukundansundar:
@orizohar @youngbupark @yaron2 

This is my idea for implementing this: 
1. Introduce a command line flag for `dapr init` such as `--without-docker` or `--no-docker`
2. Update docs with a how-to that describes how to install without docker running 
  -- run the placement service as part of the installation steps https://github.com/dapr/dapr/releases/download/v0.8.0/placement_darwin_amd64.tar.gz
3. Update samples/1.hello-world to include the example where dapr is run without docker.

LMWF:
I don't think the design for this was finished.  Both sample 1 and actors need a state store.  If we're going to have a "no docker" mode either it should have either:
1. only service invocation and no other building blocks or
2. the same things as today, but the state store and placement don't run in containers.

Option 1 makes for reduced functionality, but it allows the user to see one key feature of dapr.

Option 2 implies that the state store either be
a. local, but not containerized or
b. somewhere else (e.g. in a cloud) or
c. a emulated/in-memory state store/local shared file - meant for this scenario only and not supported in any other way.

Option a tends to have the problems that it is difficult to set up and may have a lot of env dependencies.  This would probably take longer and have more resulting problems than setting up docker.

Option b requires the user have an account and set up the component themselves, which might a bit of a hassle to run a local sample.

Option c seems like a reasonable direction - we can achieve the aims of the issue without increasing  env problems or user set up.

mukundansundar:
@LMWF good point. This also brings into the discussion on whether this "no docker" mode should have tracing enabled as part of it or not. https://github.com/dapr/cli/issues/342 

For tracing also we have the same issue of either emulating a zipkin service(or other exporters) or run it locally or run it in cloud.

yaron2:
> 1. Introduce a command line flag for dapr init such as --without-docker or --no-docker

I would opt to use for some more standardized terms to signify a reduced set of capabilities.
[...]

PR #386, opened 2020-06-29
Initialize dapr without docker dependency

mukundansundar:
# Description

Added a `--slim` flag to initialize dapr without docker. 

## Issue reference

Please reference the issue this PR will close: #175 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Extended the documentation - https://github.com/dapr/docs/issues/664, https://github.com/dapr/samples/issues/247

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,9 +154,7 @@ $ dapr init --redis-host 10.0.0.1
 
 ### Uninstall Dapr in a standalone mode
 
-#### With Docker
-
-Uninstalling will remove the placement container and the daprd binary installed in either the provided `--install-path` on `dapr init` or the default path `/usr/local/bin` for Linux/MacOS or `C:\dapr` for Windows. 
+Uninstalling will remove the placement container if installed with Docker or the placement binary if not, and the daprd binary installed in either the provided `--install-path` on `dapr init` or the default path `/usr/local/bin` for Linux/MacOS or `C:\dapr` for Windows. 
 
 
 ```bash
@@ -169,20 +167,11 @@ The command above won't remove the redis or zipkin containers by default in case
 $ dapr uninstall --all
 ```
 
-You should always run a `dapr uninstall` before running another `dapr init`.	
+The above command can also be run when Dapr has been installed in a non-docker environment, it will only remove the installed binaries and the default dapr folder in that case. 
 
-#### Without Docker
+> NB: The `dapr uninstall` command will always try to remove the placement binary/service and will throw an error is not able to.
 
-If Dapr has been installed in self-hosted mode with docker, run the following command to remove the installed binaries(daprd, placement): 
-
-```bash
-$ dapr uninstall --slim
-ℹ️  Removing Dapr from your machine...
-removing binary:  /usr/local/bin/daprd
-removing binary:  /usr/local/bin/placement
-WARNING: could not delete run data file
-✅  Dapr binaries have been removed successfully
-```
+**You should always run a `dapr uninstall` before running another `dapr init`.**
 
 #### Uninstall Dapr from a specific install path 
 
--- a/README.md
+++ b/README.md
@@ -8,10 +8,11 @@ The Dapr CLI allows you to setup Dapr on your local dev machine or on a Kubernet
 ## Getting started
 
 ### Prerequisites
+If not following the instructions for installing Dapr [without containers](#Without-Containers)
 
 * Install [Docker](https://docs.docker.com/install/)
 
-*__Note: On Windows, Docker must be running in Linux Containers mode__*
+>__Note: On Windows, Docker must be running in Linux Containers mode__
 
 ### Installing Dapr CLI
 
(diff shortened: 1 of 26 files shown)


README.md
```diff
@@ -135,6 +171,19 @@ $ dapr uninstall --all
 
 You should always run a `dapr uninstall` before running another `dapr init`.	
 
+#### Without Docker
+
+If Dapr has been installed in self-hosted mode with docker, run the following command to remove the installed binaries(daprd, placement): 
+
+```bash
+$ dapr uninstall --slim
```
amanbha, line 179: Does uninstall have to be separate, dapr unintall can take care of it with some checks in the code.

README.md
```diff
@@ -83,6 +90,34 @@ This step creates the following defaults:
...
+removing extracted binary ~/.dapr/daprd
+✅  Downloaded binaries and completed components set up.
+ℹ️  daprd binary has been installed.
+
+ℹ️  placement binary has been installed.
+
+✅  Success! Dapr is up and running. To get started, go here: https://aka.ms/dapr-getting-started
+```
+
+>Note: Only the Dapr runtime binary and the placement service binary are installed. An empty components folder, which is later used during `dapr run` unless the `--components-path` option is provided, is created. For Linux/MacOS, the default components folder path is `$HOME/.dapr/components` and for Windows it is `%USERPROFILE%\.dapr\components`.
```
amanbha, line 119: What is the scenario for creating am empty component folder?

mukundansundar: we set `standalone.DefaultComponentsDirPath()` as default and without an empty folder, it will throw an exception.

README.md
mukundansundar: it can but we will not be able to differentiate between not being able to connect to docker vs docker not being needed.

amanbha: Sure, we will now have 3 options:
dapr uninstall (removes dapr binaries and placement container)
dapr uninstall --all (removes dapr binaries, placement container, redis etc)
dapr uninstall --slim (removes dapr binaries.)

amanbha:
+ @orizohar for the behavior.

README.md
mukundansundar: Changed logic for uninstall and removed the `--slim` flag for uninstall.

orizohar:
> * @orizohar for the behavior of dapr init --slim

What part of the behavior is under discussion?

README.md
```diff
@@ -51,11 +51,18 @@ Each release of Dapr CLI includes various OSes and architectures. These binary v
    * For Linux/MacOS - `/usr/local/bin`
    * For Windows, create a directory and add this to your System PATH. For example create a directory called `c:\dapr` and add this directory to your path, by editing your system environment variable.
 
-### Install Dapr on your local machine (standalone)
+### Install Dapr on your local machine (self-hosted)
+
+In self-hosted mode, dapr can be installed with docker(recommended) or without a docker dependency.
```
yaron2, line 56: this suggests that dapr installs Docker which is not true.

README.md
```diff
@@ -51,11 +51,18 @@ Each release of Dapr CLI includes various OSes and architectures. These binary v
...
+
+#### With Docker(recommended)
```
yaron2, line 58: Phrasing needs to change here. it's "With Redis Container" much more than "With Docker".

README.md
```diff
@@ -83,6 +90,34 @@ This step creates the following defaults:
 2. component files in the components folder called `pubsub.yaml`, `statestore.yaml` and `zipkin.yaml`. 
 3. default config file `$HOME/.dapr/config.yaml` for Linux/MacOS or for Windows at `%USERPROFILE%\.dapr\config.yaml` to enable tracing on `dapr init` call. Can be overridden with the `--config` flag on `dapr run`.
 
+#### Without Docker
```
yaron2, line 93: same as above

cmd/init.go
```diff
@@ -53,6 +58,7 @@ var InitCmd = &cobra.Command{
 
 func init() {
 	InitCmd.Flags().BoolVarP(&kubernetesMode, "kubernetes", "k", false, "Deploy Dapr to a Kubernetes cluster")
+	InitCmd.Flags().BoolVarP(&slimMode, "slim", "s", false, "Initialize dapr in self-hosted mode without docker.")
```
yaron2, line 61: Rephrase "without docker" to "without Redis and Actor Placement containers" or something similar

amanbha:
> > * @orizohar for the behavior of dapr init --slim
> 
> What part of the behavior is under discussion?

I added you for the PM approval for the overall behavior

youngbupark:
lgtm

README.md
orizohar: For this one, since we are using --slim, we should use that term, something like "Slim init" or "Minimal installation"

amanbha:
> > * @orizohar for the behavior of dapr init --slim
> 
> What part of the behavior is under discussion?

For the overall feature behavior and doc changes.

README.md
orizohar, line 58: I would phrase this as default initialization.

README.md
orizohar, line 56: In essence, this is not about with docker or without docker, or with these containers or without. The init command provides a "developer environment" that allows a user to get started easily. It does so using Docker so there is a dependency in the default installation. If a user does not want to use docker or does not wish to have that dev env setup for them, Dapr offers a minimal init (or slim). I would add this kind of explanation here to better articulate it

PR #387, opened 2020-07-01
Added documentation for dapr dashboard

gdhuper:
# Description

Added docs for `dapr dashboard` command, and fixed a grammatical error.

## Issue reference

This PR will close: #341 

## Checklist

* [x] Code compiles correctly

### DIFF
--- a/docs/reference/dapr-dashboard.md
+++ b/docs/reference/dapr-dashboard.md
@@ -1,0 +1,19 @@
+# dapr dashboard
+
+## Description
+
+Starts Dapr dashboard in a Kubernetes cluster.
+
+## Usage
+
+```bash
+dapr dashboard [flags]
+```
+
+## Flags
+
+| Name | Environment Variable | Default | Description
+| --- | --- | --- | --- |
+| `--help`, `-h` | | | Help for dashboard |
+| `--kubernetes`, `-k` | | `false` | Start Dapr dashboard in local browser |
+| `--port`, `-p` | | `8080` | The local port on which to serve dashboard |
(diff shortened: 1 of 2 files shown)


docs/reference/dapr-dashboard.md
```diff
@@ -0,0 +1,19 @@
+# dapr dashboard
+
+## Description
+
+Starts Dapr dashboard in a Kubernetes cluster.
```
shalabhms, line 5: `dapr dashboard ` is for standalone and kubernetes.  Only when invoked with flag -k , it starts for kubernetes.
So you might need to change "Start Dapr dashboard" for now. Since with current changes, it is not supported for standalone.

shalabhms: Since the dashboard deployment has to be done before running dashboard command. 
It would be better to state that in the doc and command help about this pre-reqs:

Prerequisites : Dapr dashboard should be deployed in the Kubernetes cluster. 
You can deploy the dashboard in the Kubernetes cluster as follows:
kubectl apply -f https://raw.githubusercontent.com/dapr/dashboard/master/deploy/dashboard.yaml "

gdhuper: Gotcha! I'll go ahead and update the docs. For dashboard deployment, I remember Yaron mentioning that `dapr init` would take care of it. Is that issue going to be tackled in this milestone?

shalabhms: Yes the plan is  that `dapr init` will take care of,  yes it is targeted to this milestone but might come later. Due to that, till it comes, it is better to add the prereqs in the doc, later it can be removed from doc once `dapr init` takes care of.

PR #390, opened 2020-07-09
Dashboard integration: Updating user requirements and port-forwarding correct namespace

### LINKED ISSUE #389: Dashboard CLI command is hard coded to do port-forwarding in default namespace
Dashboard CLI command currently does port forwarding using default namespace.  Dashboard could be in dapr-system namespace.  

So need to fix the code by  :
  1. First try in "dapr-system" namespace
  2. if not, then fallback to "default" namespace
  3. otherwise show the correct error message to the user.

Also add the flag `namespace` as parameter.

willdavsmith:
# Description

* Removed user requirement to apply dashboard to their cluster, rely on 'dapr init -k' or helm install
* Added search for dashboard namespace: searches namespaces dapr-system and default for dashboard pod
* Added method to check if pod exists in namespace
* Added -n flag for dapr dashboard command to allow the user to supply a namespace
* Updated documentation for dashboard command

## Issue reference

closes #389 

References: https://github.com/dapr/dapr/pull/1752

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/dashboard.go
+++ b/cmd/dashboard.go
@@ -14,7 +14,6 @@ import (
 	"github.com/dapr/cli/pkg/print"
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -27,10 +26,14 @@ const (
 	// defaultLocalPort is the default local port used for port forwarding for `dapr dashboard`
 	defaultLocalPort = 8080
 
+	// defaultNamespace is the default namespace where the dashboard is deployed
+	defaultNamespace = "dapr-system"
+
 	// remotePort is the port dapr dashboard pod is listening on
 	remotePort = 8080
 )
 
+var dashboardNamespace string
 var localPort int
 
 var DashboardCmd = &cobra.Command{
@@ -43,11 +46,10 @@ var DashboardCmd = &cobra.Command{
 			localPort = port
 		}
 
-		config, err := kubernetes.GetKubeConfig()
-
+		config, client, err := kubernetes.GetKubeConfigClient()
 		if err != nil {
 			print.FailureStatusEvent(os.Stdout, "Failed to initialize kubernetes client")
-			return
+			os.Exit(1)
 		}
 
 		// manage termination of port forwarding connection on interrupt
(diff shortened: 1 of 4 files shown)


willdavsmith:
Hi all,

Originally, this issue included the ability for the user to pass a -n flag to dapr dashboard, allowing them to set the preferred namespace to run the dashboard. This was removed because it seemed unnecessary. If anyone has any strong opinions about adding/ keeping this feature removed, please let me know.

shalabhms:
> Hi all,
> 
> Originally, this issue included the ability for the user to pass a -n flag to dapr dashboard, allowing them to set the preferred namespace to run the dashboard. This was removed because it seemed unnecessary. If anyone has any strong opinions about adding/ keeping this feature removed, please let me know.

We discussed further and we are adding that capability as Dapr control plane could be running in other than dapr-system or default name.

cmd/dashboard.go
```diff
@@ -43,11 +46,39 @@ var DashboardCmd = &cobra.Command{
 			localPort = port
 		}
 
-		config, err := kubernetes.GetKubeConfig()
-
+		config, client, err := kubernetes.GetKubeConfigClient()
 		if err != nil {
 			print.FailureStatusEvent(os.Stdout, "Failed to initialize kubernetes client")
```
shalabhms, line 51: You can print the error also using  err.Error()

cmd/dashboard.go
```diff
@@ -43,11 +46,39 @@ var DashboardCmd = &cobra.Command{
...
+			os.Exit(1)
+		}
+
+		// search for dashboard service namespace in order:
+		// user-supplied namespace, dapr-system, default
+		namespaces := []string{dashboardNamespace}
+		if dashboardNamespace != defaultNamespace {
+			namespaces = append(namespaces, defaultNamespace)
+		}
+		namespaces = append(namespaces, "default")
```
amanbha, line 61: if users provides -n default.
This will add "default" again. It should be added inside if similar to the previous one.

cmd/dashboard.go
```diff
@@ -27,10 +26,14 @@ const (
 	// defaultLocalPort is the default local port used for port forwarding for `dapr dashboard`
 	defaultLocalPort = 8080
 
+	// defaultNamespace is the default namespace where the dashboard is deployed
+	defaultNamespace = "dapr-system"
```
amanbha, line 30: rename this var to daprSystemNamespace and introduce another variable defaultNamespace for "default"

cmd/dashboard.go
```diff
@@ -43,11 +46,39 @@ var DashboardCmd = &cobra.Command{
...
+		foundNamespace := ""
+		for _, namespace := range namespaces {
+			ok, _ := kubernetes.CheckPodExists(client, namespace, nil, dashboardSvc)
+			if ok {
+				foundNamespace = namespace
+				break
+			}
+		}
+
+		// if the service is not found, try to search all pods
```
shalabhms, line 77: after some thoughts, I think we can remove this code path . This will give user wrong expectation that dashboard is running in namespace which user provided which could be false information . Exception is only for default and dapr-system.

Instead,  the correct experience would be , print the message that "Dapr dashboard found to be running in the namespace. Please try with namespace"

cmd/dashboard.go
```diff
@@ -99,6 +136,7 @@ var DashboardCmd = &cobra.Command{
 func init() {
 	DashboardCmd.Flags().BoolVarP(&kubernetesMode, "kubernetes", "k", false, "Start Dapr dashboard in local browser")
 	DashboardCmd.Flags().IntVarP(&port, "port", "p", defaultLocalPort, "The local port on which to serve dashboard")
+	DashboardCmd.Flags().StringVarP(&dashboardNamespace, "namespace", "n", defaultNamespace, "The namespace where Dapr dashboard is running")
```
amanbha, line 139: after rename of vars, does it needs to changed to daprSystemNamespace

willdavsmith: Yes, good catch. I'll fix it

PR #407, opened 2020-07-16
Integrate Dashboard with `Dapr init` and able to launch in standalone mode

willdavsmith:
# Description
* Integrate Dapr dashboard with dapr init
* Added ability to launch dashboard in standalone
* Added --dashboard-path (-d) flag to allow the user to specify dashboard install location

## Issue reference

Closes #238 
Dependent on #417 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Extended the documentation

Issue #238, opened 2020-01-31
Install Dapr dashboard when using the dapr init command


msfussell:
The dapr dashboard provides a visualization of when is happening in the cluster or local environment. This should be installed by default using the dapr init command which set up development environment

PR #417, opened 2020-07-24
Setup for addition of dashboard self-hosted command

willdavsmith:
# Description

Making necessary changes to current CLI to allow for later pull request adding `dapr dashboard` command in self-hosted mode.

* Updated pkg/standalone.go/untar function to untar all files in tarball rather than just binary
* Added check for kubernetes mode in pkg/dashboard.go
* Added 'githubRepo' as a parameter to installBinary function
* Removed extra error message

## Issue reference

References #407 
References #238 

Will close in next PR

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/dashboard.go
+++ b/cmd/dashboard.go
@@ -34,10 +34,6 @@ const (
 
 	// remotePort is the port dapr dashboard pod is listening on
 	remotePort = 8080
-
-	// default install locations for Dapr executables
-	daprDefaultLinuxAndMacInstallPath = "/usr/local/bin"
-	daprDefaultWindowsInstallPath     = "c:\\dapr"
 )
 
 var dashboardNamespace string
@@ -138,8 +134,6 @@ var DashboardCmd = &cobra.Command{
 			}
 
 			<-portForward.GetStop()
-		} else {
-			// Standalone mode
 		}
 	},
 }
--- a/cmd/dashboard.go
+++ b/cmd/dashboard.go
@@ -29,7 +29,7 @@ const (
 	// daprSystemNamespace is the namespace "dapr-system" (recommended Dapr install namespace)
 	daprSystemNamespace = "dapr-system"
 
-	// defaultNamespace is the default namespace (from dapr init -k installation)
+	// defaultNamespace is the default namespace (dapr init -k installation)
 	defaultNamespace = "default"
 
 	// remotePort is the port dapr dashboard pod is listening on
--- a/cmd/dashboard.go
+++ b/cmd/dashboard.go
@@ -153,7 +153,6 @@ var DashboardCmd = &cobra.Command{
 					binaryName = "dashboard"
 				}
 			}
-			fmt.Println(dashboardPath)
 			cmdDashboardStandalone := &exec.Cmd{
 				Path:   path_filepath.Join(dashboardPath, binaryName),
 				Dir:    dashboardPath,
--- a/cmd/dashboard.go
+++ b/cmd/dashboard.go
@@ -14,7 +14,6 @@ import (
 	"github.com/dapr/cli/pkg/print"
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -27,10 +26,17 @@ const (
 	// defaultLocalPort is the default local port used for port forwarding for `dapr dashboard`
 	defaultLocalPort = 8080
 
+	// daprSystemNamespace is the namespace "dapr-system" (recommended Dapr install namespace)
+	daprSystemNamespace = "dapr-system"
+
+	// defaultNamespace is the default namespace (dapr init -k installation)
+	defaultNamespace = "default"
+
 	// remotePort is the port dapr dashboard pod is listening on
 	remotePort = 8080
 )
 
+var dashboardNamespace string
 var localPort int
 
 var DashboardCmd = &cobra.Command{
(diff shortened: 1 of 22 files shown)


pkg/standalone/standalone.go
```diff
@@ -612,41 +612,45 @@ func untar(filepath, targetDir, binaryFilePrefix string) (string, error) {
...
 			}
+			continue
+		}
 
-			// #nosec G110
-			if _, err := io.Copy(f, tr); err != nil {
-				return "", err
-			}
-			f.Close()
+		f, err := os.OpenFile(path, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, info.Mode())
```
shalabhms, line 637: these permissions are different than earlier change,,os.O_RDWR missing..? Do we need this additional permissions ?

pkg/standalone/standalone.go
```diff
@@ -612,41 +612,45 @@ func untar(filepath, targetDir, binaryFilePrefix string) (string, error) {
...
 
+	foundBinary := ""
 	for {
 		header, err := tr.Next()
 
-		switch {
-		case err == io.EOF:
-			return "", fmt.Errorf("file is empty")
-		case err != nil:
+		if err == io.EOF {
```
shalabhms, line 619: We can still have switch block...no need to change to if/else blocks...

willdavsmith: Actually, it's kind of surprising but this does not work with the switch statement. the 'break' statement will break out of the switch statement rather than the for. My options were to either do this or create a label for the outer for loop and break out of that, and I went with this solution.

shalabhms:
@mukundansundar , could you please review this PR..changes around install/uninstall ?

pkg/standalone/standalone.go
```diff
@@ -612,41 +612,45 @@ func untar(filepath, targetDir, binaryFilePrefix string) (string, error) {
...
+		defer f.Close()
+
+		// #nosec G110
+		if _, err = io.Copy(f, tr); err != nil {
+			return "", err
+		}
 
-			return extractedFilePath, nil
+		// If the found file is the binary that we want to find, save it and return later
+		if strings.HasSuffix(header.Name, binaryFilePrefix) || strings.HasSuffix(header.Name, binaryFilePrefix+".exe") {
```
mukundansundar, line 649: runtime.GOOS is windows is handled by unzip method

pkg/standalone/standalone.go
```diff
@@ -612,41 +612,45 @@ func untar(filepath, targetDir, binaryFilePrefix string) (string, error) {
 
 	tr := tar.NewReader(gzr)
 
+	foundBinary := ""
```
mukundansundar, line 615: no need for foundBinary ... can be refactored

pkg/standalone/uninstall.go
```diff
@@ -97,7 +96,6 @@ func Uninstall(uninstallAll bool, dockerNetwork string) error {
 		}
 	}
 
-	err = errors.New("uninstall failed")
```
mukundansundar, line 100: Why is this removed?

pkg/standalone/standalone.go
```diff
@@ -429,29 +429,29 @@ func runPlacementService(wg *sync.WaitGroup, errorChan chan<- error, dir, versio
...
 		version = version[1:]
 	}
-	// https://github.com/dapr/dapr/releases/download/v0.8.0/daprd_darwin_amd64.tar.gz
-	// https://github.com/dapr/dapr/releases/download/v0.8.0/placement_darwin_amd64.tar.gz
+
 	fileURL := fmt.Sprintf(
 		"https://github.com/%s/%s/releases/download/v%s/%s_%s_%s.%s",
 		cli_ver.DaprGitHubOrg,
-		cli_ver.DaprGitHubRepo,
+		githubRepo,
```
mukundansundar, line 454: why is this a local variable while the Org one is not?

pkg/standalone/standalone.go
willdavsmith: The current artifacts published for Dashboard are tarfiles for both windows and linux, So I had to make adjustments to this

pkg/standalone/standalone.go
willdavsmith: I added foundBinary to ensure that all files in the tarball are extracted, even if the reader finds the binary first. If I returned as soon as I found the binary, then the function may not extract everything from the tarball.

pkg/standalone/standalone.go
willdavsmith: I left the Org as such because all binaries that we install should be from dapr org. I can change this though, just let me know.

shalabhms:
@mukundansundar , could you please look into the comments reply and resolve if looks good?

pkg/standalone/standalone.go
willdavsmith: The artifact for dashboard on windows is now a zip archive, so I updated this.

cmd/dashboard.go
```diff
@@ -49,90 +49,92 @@ var DashboardCmd = &cobra.Command{
...
-		// search for dashboard service namespace in order:
-		// user-supplied namespace, dapr-system, default
-		namespaces := []string{dashboardNamespace}
-		if dashboardNamespace != daprSystemNamespace {
-			namespaces = append(namespaces, daprSystemNamespace)
-		}
-		if dashboardNamespace != defaultNamespace {
-			namespaces = append(namespaces, defaultNamespace)
-		}
+			// search for dashboard service namespace in order:
```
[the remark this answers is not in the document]
mukundansundar: This still looks like tabs. Can you recheck this part?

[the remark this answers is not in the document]
shalabhms: Figured out that Github provides the setting to hide view space change..Once we have that setting , it does not show .

PR #431, opened 2020-08-12
Add certificate export command

### LINKED ISSUE #419: Add a certificate export command
When upgrading Dapr to a new version, it's recommended to carry over the existing certificates to avoid downtime.
The process of carrying over the certs is a little lengthy and requires the following:

1. `kubectl get secret dapr-trust-bundle -o yaml -n <NAMESPACE>`
2. base64 decoding the values for `issuer.crt`, `issuer.key` and `ca.crt`
3. Saving the decoded values in a 3 files (root, issuer certs and issuer key)

While doable, it would be helpful if there was a CLI command to do that on behalf of the user:

```bash
dapr mtls export -o ./certs
```

/cc @msfussell @orizohar

yaron2:
This PR does the following:

1. Adds a command that helps users export their TLS certificates from a kubernetes cluster
2. Fixes race condition bugs for the k8s/dapr clients

Closes #419

### DIFF
--- a/pkg/kubernetes/mtls.go
+++ b/pkg/kubernetes/mtls.go
@@ -1,28 +1,92 @@
 package kubernetes
 
 import (
+	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/dapr/dapr/pkg/apis/configuration/v1alpha1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	defaultConfigName = "default"
+	systemConfigName      = "daprsystem"
+	trustBundleSecretName = "dapr-trust-bundle"
 )
 
 func IsMTLSEnabled() (bool, error) {
-	client, err := DaprClient()
+	c, err := getSystemConfig()
 	if err != nil {
 		return false, err
+	}
+	return c.Spec.MTLSSpec.Enabled, nil
+}
+
+func getSystemConfig() (*v1alpha1.Configuration, error) {
+	client, err := DaprClient()
+	if err != nil {
+		return nil, err
 	}
 
 	configs, err := client.ConfigurationV1alpha1().Configurations(meta_v1.NamespaceAll).List(meta_v1.ListOptions{})
 	if err != nil {
-		return false, err
+		return nil, err
 	}
 
 	for _, c := range configs.Items {
-		if c.GetName() == defaultConfigName {
-			return c.Spec.MTLSSpec.Enabled, nil
+		if c.GetName() == systemConfigName {
+			return &c, nil
 		}
 	}
-	return false, nil
+	return nil, errors.New("system configuration not found")
+}
+
+func ExportTrustChain(outputDir string) error {
+	_, err := os.Stat(outputDir)
+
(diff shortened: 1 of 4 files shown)


pkg/kubernetes/mtls.go
```diff
@@ -1,28 +1,92 @@
...
+	if err != nil {
+		return err
+	}
+	res, err := client.CoreV1().Secrets(c.GetNamespace()).List(meta_v1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, i := range res.Items {
+		if i.GetName() == trustBundleSecretName {
```
shalabhms, line 70: Will it be only 1 trustBundleSecretName (dapr-trust-bundle) or expected to be >=1 ?

yaron2: There can ever only be one

pkg/kubernetes/mtls.go
```diff
@@ -1,28 +1,92 @@
...
+
+			err = ioutil.WriteFile(filepath.Join(outputDir, "issuer.crt"), issuerCert, 0600)
+			if err != nil {
+				return err
+			}
+
+			err = ioutil.WriteFile(filepath.Join(outputDir, "issuer.key"), issuerKey, 0600)
+			if err != nil {
+				return err
+			}
```
shalabhms, line 88: If expected only one bundle,  then we could possibly break here...

pkg/kubernetes/mtls.go
```diff
@@ -1,28 +1,92 @@
...
-		if c.GetName() == defaultConfigName {
-			return c.Spec.MTLSSpec.Enabled, nil
+		if c.GetName() == systemConfigName {
+			return &c, nil
+		}
+	}
+	return nil, errors.New("system configuration not found")
+}
+
+func ExportTrustChain(outputDir string) error {
```
shalabhms, line 45: nit : comment for exported methods...

pkg/kubernetes/mtls.go
yaron2: Added a break

pkg/kubernetes/mtls.go
yaron2: Added. I wonder why the linter didn't warn about this.

pkg/kubernetes/mtls.go
```diff
@@ -1,28 +1,92 @@
...
+			issuerKey := i.Data["issuer.key"]
+
+			err := ioutil.WriteFile(filepath.Join(outputDir, "ca.crt"), ca, 0600)
+			if err != nil {
+				return err
+			}
+
+			err = ioutil.WriteFile(filepath.Join(outputDir, "issuer.crt"), issuerCert, 0600)
+			if err != nil {
+				return err
```
shalabhms, line 83: If it fails at this step of exporting say issuer.crt , will re-run of the same command could possible fail at first step of exporting ca.cert which was successfully exported at first time as the file already exists..

Should we remove the directory or remove the content files before exporting...?

yaron2: No, it will overwrite existing files.

Issue #156, opened 2019-10-16
Remove certs from remote install manifest


yaron2:
## Describe the proposal

Currently, `dapr init --kubernetes` pulls an install manifest from Azure Blob Storage which is generated by Helm and has the Sidecar injector tls certs hardcoded.

We should change that to generate the certs at runtime.

PR #432, opened 2020-08-13
Refactor init on kubernetes to use Helm

### LINKED ISSUE #281: Ability to define namespace to deploy dapr on init
## Describe the proposal

Today when you run `dapr init --kubernetes` it installs the operator in the default namespace. Would be nice if you could also pass in some flag to define another namespace.  `dapr init --kubernetes --namespace dapr`

yaron2:
This PR does the following:

1. Uses Helm programmatically to install Dapr
2. Removes the need for `operator.yaml` and it's associated CI actions
3. Allows the user to use `dapr init -k` with the following options:
   a. Custom namespace
   b. mTLS enabled/disabled
  c. HA mode enabled/disabled
4. Fixes the security issue of having `operator.yaml` install the same trust chain certificates per Dapr version

Essentially, this PR enables the Dapr CLI to provide Helm-like installations that are consistent with our charts, without requiring the users to actually have Helm installed.

This provides a much more robust installation path and opens the door to many features.

Closes #281 and #156.


pkg/kubernetes/kubernetes.go
```diff
@@ -7,42 +7,40 @@ package kubernetes
...
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	helm "helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/cli"
 )
 
 const (
+	daprReleaseName   = "dapr"
+	daprHelmRepo      = "https://daprio.azurecr.io/helm/v1/repo"
```
amanbha, line 31: we should change helm repo to https://github.com/dapr/dapr-helm-charts.io

pkg/kubernetes/chart_versions.go
```diff
@@ -0,0 +1,19 @@
+package kubernetes
+
+// This can be deprecated in the future (>= 1.0) when chart versions align to runtime versions.
+var chartVersionsMap = map[string]string{
+	"0.7.0": "0.4.0",
+	"0.7.1": "0.4.1",
+	"0.8.0": "0.4.2",
+	"0.9.0": "0.4.3",
```
amanbha, line 8: lets add 0.10.0 version here as its coming out next week, else it would need another change to cli.

pkg/kubernetes/uninstall.go
```diff
@@ -6,30 +6,17 @@
...
-		}
-
-		version = v
-	}
-
-	var daprManifestPath string = fmt.Sprintf("https://github.com/dapr/dapr/releases/download/%s/dapr-operator.yaml", version)
-
-	_, err := utils.RunCmdAndWait("kubectl", "delete", "-f", daprManifestPath)
+// Uninstall removes Dapr from a Kubernetes cluster
+func Uninstall(namespace string) error {
```
amanbha, line 13: We need to put a note in breaking changes that, cli installs in k8s till 0.9.0 must be uninstalled with cli 0.9.0 and then get the new cli.

pkg/kubernetes/kubernetes.go
youngbupark: jFYI - this is the correct url - https://dapr.github.io/dapr-helm-charts.io/ :)

pkg/kubernetes/uninstall.go
yaron2: Sure, planning to do that.

pkg/kubernetes/kubernetes.go
yaron2: Done

pkg/kubernetes/chart_versions.go
yaron2: We plan to start versioning the chart versions to the Dapr runtime versions beginning 0.10.0

amanbha: sounds good!

PR #433, opened 2020-08-13
Adding dashboard version command

willdavsmith:
# Description

* Added command: `dapr dashboard --version`, which prints the version of the installed dashboard binary

## Issue reference

dapr/dashboard#71, dapr/dashboard#56

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pkg/kubernetes/kubernetes.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	daprLatestVersion = "latest"
+	daprLatestVersion = "vlatest"
```
shalabhms, line 24: what is "vlatest" ?, is this right ?

willdavsmith: yes, in current CLI, if a user runs `dapr init -k`, cli tries to get the 'vlatest' version of dapr. This change makes it so that the user can run `dapr init -k` without supplying a runtime version.

shalabhms: looks like it is changed. please take latest changes and update the PR and then this is not needed.

Issue #439, opened 2020-08-18
Dapr CLI version flag not working


mukundansundar:
## Expected Behavior
CLI version: 0.10.0-rc.0
Runtime version : 0.10.0-rc.0

## Actual Behavior

flag provided but not defined: -version
Usage of dapr:
  -kubeconfig string
    	(optional) absolute path to the kubeconfig file (default "/Users/msundar/.kube/config")

## Steps to Reproduce the Problem
with the new rc cli

`dapr --version`

PR #440, opened 2020-08-18
Parse flags when needed in getConfig

mukundansundar:
# Description

Fixing flag parsing only when required.

## Issue reference

Please reference the issue this PR will close: #439 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/kubernetes/client.go
+++ b/pkg/kubernetes/client.go
@@ -31,10 +31,10 @@ func init() {
 	} else {
 		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
 	}
-	flag.Parse()
 }
 
 func getConfig() (*rest.Config, error) {
+	flag.Parse()
 	kubeConfigEnv := os.Getenv("KUBECONFIG")
 	delimiterBelongsToPath := strings.Count(*kubeconfig, kubeConfigDelimiter) == 1 && strings.EqualFold(*kubeconfig, kubeConfigEnv)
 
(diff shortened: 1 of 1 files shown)


mukundansundar:
The flags should be parsed only when needed. Small change made from #431 .

pkg/kubernetes/client.go
```diff
@@ -31,10 +31,10 @@ func init() {
 	} else {
 		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
 	}
-	flag.Parse()
 }
 
 func getConfig() (*rest.Config, error) {
+	flag.Parse()
```
yaron2, line 37: This might mess up both `dapr mtls expiry` and `dapr mtls export`. Please verify those two commands work after this change.

mukundansundar: Tested working ... since it also calling `GetKubeConfigClient()` which calls `getConfig()` which calls for `flag.Parse()`

pruthvidhodda (dismissed):

pkg/kubernetes/client.go
mukundansundar: $./dist/darwin_amd64/release/dapr mtls expiry
Root certificate expires in 8736 hours. Expiry date: 2021-08-17 22:13:29 +0000 UTC
$ ./dist/darwin_amd64/release/dapr mtls export
✅  Trust certs successfully exported to <pwd>

amanbha (dismissed):

pkg/kubernetes/client.go
yaron2: For safety, we should use DoOnce to make sure we run it only once:

```go
doOnce.Do(func() {
flag.Parse()
})
```

mukundansundar: done

PR #446, opened 2020-08-20
Generate CLI docs on make command

mukundansundar:
Add gen docs and diff checking to workflow

# Description

Generate docs automatically on make command. 

Error out if the docs are different or new files are added in workflow. Similar to go mod diff.  

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


cmd/run.go
```diff
@@ -54,6 +54,12 @@ Run sidecar only:
 		viper.BindPFlag("placement-host", cmd.Flags().Lookup("placement-host"))
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		if configFile == "" {
```
mukundansundar, line 57: This is checked here and not added in defaults so that the automatic docs generated do not have any platform dependent defaults.

mukundansundar: https://github.com/dapr/cli/runs/1009642534?check_suite_focus=true#step:8:28

.github/scripts/docs.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/dapr/cli/cmd"
+	"github.com/spf13/cobra/doc"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		panic("Requires a path to generate docs in.")
+	}
+	// Tag contains date. Need to disable generation to avoid date generation errors.
+	cmd.RootCmd.DisableAutoGenTag = true
```
mukundansundar, line 20: Example of error: https://github.com/dapr/cli/pull/446/checks?check_run_id=1010866665#step:8:89

docs/development/development.md
```diff
@@ -70,6 +70,12 @@ make build GOOS=linux GOARCH=amd64
 make test
 ```
 
+### Generate CLI docs
+
+```bash
```
amanbha, line 75: Can we add the steps here on what is needed to checkin the docs when new command is added or existing one is changed?

PR #461, opened 2020-09-15
Waits for dapr sidecar before running app without port.

artursouza:
# Description

If app does not expose a port, cli will wait for runtime before starting app.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/1933

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -13,13 +13,19 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/dapr/cli/pkg/kubernetes"
 	"github.com/dapr/cli/pkg/metadata"
 	"github.com/dapr/cli/pkg/print"
 	"github.com/dapr/cli/pkg/standalone"
+	"github.com/dapr/cli/utils"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+)
+
+const (
+	runtimeWaitTimeoutInSeconds = 60
 )
 
 var appPort int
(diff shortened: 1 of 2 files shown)


cmd/run.go
```diff
@@ -142,6 +148,31 @@ Run sidecar only:
...
 				}
 
+				if appPort <= 0 {
+					// If app does not listen to port, we can check for Dapr's sidecar health before starting the app.
+					// Otherwise, it creates a deadlock.
+					sidecarUp := true
+					print.InfoStatusEvent(os.Stdout, "Checking if Dapr sidecar is listening on HTTP port %v", output.DaprHTTPPort)
+					err = utils.IsDaprListeningOnPort(output.DaprHTTPPort, time.Duration(runtimeWaitTimeoutInSeconds)*time.Second)
+					if err != nil {
+						sidecarUp = false
```
amanbha, line 158: sidecarUp will stay as false once it reaches here even if the second check to IsDaprListeningOnPort succeeds

cmd/run.go
```diff
@@ -142,6 +148,31 @@ Run sidecar only:
...
+						print.WarningStatusEvent(os.Stdout, "Dapr sidecar is not listening on HTTP port: %s", err.Error())
+					}
+
+					print.InfoStatusEvent(os.Stdout, "Checking if Dapr sidecar is listening on GRPC port %v", output.DaprGRPCPort)
```
amanbha, line 162: the same code block is running twice, move it to a loop and break from loop once side car is up. So that InfoStatusEvent is not printed twice and IsDaprListeningOnPort is not called again even if the first one succeeds

amanbha (changes requested):

cmd/run.go
artursouza: Correct. It means it is not listening to one of the ports.

cmd/run.go
artursouza: This is a different port. This is for GRPC.

amanbha: oh ok, thanks for clarifying.

amanbha (approved):

Issue #472, opened 2020-09-26
Dapr status command does not show the status of Dapr dashboard service in cluster


msfussell:
**C:\Users\mfussell\git\dapr\dapr>dapr status -k**

  NAME                   NAMESPACE    HEALTHY  STATUS   REPLICAS  VERSION      AGE  CREATED
  dapr-sidecar-injector  dapr-system  True     Running  1         0.11.0-rc.1  18s  2020-09-25 17:12.31
  dapr-placement         dapr-system  True     Running  1         0.11.0-rc.1  18s  2020-09-25 17:12.31
  dapr-operator          dapr-system  True     Running  1         0.11.0-rc.1  18s  2020-09-25 17:12.31
  dapr-sentry            dapr-system  True     Running  1         0.11.0-rc.1  18s  2020-09-25 17:12.31

**C:\Users\mfussell\git\dapr\dapr>k get pods -n dapr-system**

NAME                                    READY   STATUS    RESTARTS   AGE
dapr-dashboard-6d98bdb48f-gmmz2         1/1     Running   0          65s
dapr-operator-77ccd7b5d9-4pthh          1/1     Running   0          65s
dapr-placement-584d58c79f-5hmdl         1/1     Running   0          65s
dapr-sentry-57fdcbdb6d-tmb5z            1/1     Running   0          65s
dapr-sidecar-injector-548cc5855-lg7wx   1/1     Running   0          65s

Issue #471, opened 2020-09-26
dapr status -k returns error when there is no dapr installed on cluster


msfussell:
## Expected Behavior
Fails gracefully

## Actual Behavior
C:\Users\mfussell\git\dapr\quickstarts\hello-world>dapr status -k
panic: runtime error: index out of range [0] with length 0

goroutine 35 [running]:
github.com/dapr/cli/pkg/kubernetes.Status.func1(0xc000242160, 0xc0006f8bd8, 0xc00031e300, 0xc0006f8be0, 0x24b1059, 0xd)
        D:/a/cli/cli/pkg/kubernetes/status.go:52 +0x805
created by github.com/dapr/cli/pkg/kubernetes.Status
        D:/a/cli/cli/pkg/kubernetes/status.go:47 +0x145

## Steps to Reproduce the Problem
v0.11-rc2 build
take a cluster with no dapr installed and run *dapr status -k*

PR #473, opened 2020-09-26
Add dashboard to status command. Fail gracefully on error.

mukundansundar:
# Description

Add dashboard to status command. 

Fail gracefully on error. 

Made the status command testable and added unit tests for the same. 

For the other commands that use `ListPods` need a separate task to make them unit testable and add test. 

## Issue reference

Please reference the issue this PR will close: #472 #471 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pkg/kubernetes/pods.go
+++ b/pkg/kubernetes/pods.go
@@ -9,6 +9,14 @@ import (
 	k8s "k8s.io/client-go/kubernetes"
 )
 
+func ListPodsInterface(client k8s.Interface, labelSelector map[string]string) (*core_v1.PodList, error) {
+	opts := v1.ListOptions{}
+	if labelSelector != nil {
+		opts.LabelSelector = labels.FormatLabels(labelSelector)
+	}
+	return client.CoreV1().Pods(v1.NamespaceAll).List(opts)
+}
+
 func ListPods(client *k8s.Clientset, namespace string, labelSelector map[string]string) (*core_v1.PodList, error) {
 	opts := v1.ListOptions{}
 	if labelSelector != nil {
--- a/pkg/kubernetes/status.go
+++ b/pkg/kubernetes/status.go
@@ -7,16 +7,22 @@ package kubernetes
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"sync"
 
+	"errors"
+
 	"github.com/dapr/cli/pkg/age"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"github.com/dapr/cli/pkg/print"
+	k8s "k8s.io/client-go/kubernetes"
 )
 
-var (
-	controlPlaneLabels = []string{"dapr-operator", "dapr-sentry", "dapr-placement", "dapr-sidecar-injector"}
-)
+var controlPlaneLabels = []string{"dapr-operator", "dapr-sentry", "dapr-placement", "dapr-sidecar-injector", "dapr-dashboard"}
+
+type StatusClient struct {
+	client k8s.Interface
+}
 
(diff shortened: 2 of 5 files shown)


artursouza (approved):

pkg/kubernetes/pods.go
```diff
@@ -9,6 +9,14 @@ import (
 	k8s "k8s.io/client-go/kubernetes"
 )
 
+func ListPodsInterface(client k8s.Interface, labelSelector map[string]string) (*core_v1.PodList, error) {
```
tcnghia, line 12: Can you please check if *k8s.ClientSet implements k8s.Interface? I remember that it does. In that case you can just modify the other method to take k8s.Interface.

mukundansundar: Yes it does. I noted that point and I think it should be a separate task as it increases the scope of this PR a lot more than fixing status as ListPods is used in multiple places.

tcnghia: If *k8s.ClientSet implements k8s.Interface and you changed ListPod to take k8s.Interface then all existing calls are still valid.

pkg/kubernetes/status.go
```diff
@@ -45,10 +59,15 @@ func Status() ([]StatusOutput, error) {
 
 	for _, lbl := range controlPlaneLabels {
 		go func(label string) {
-			p, err := ListPods(client, v1.NamespaceAll, map[string]string{
+			defer wg.Done()
+			// Query all namespaces for Dapr pods.
+			p, err := ListPodsInterface(client, map[string]string{
 				"app": label,
 			})
 			if err == nil {
```
tcnghia, line 67: Since you changed `wg.Done()` to be defer, I think right here you can do
```
if err != nil {
  return
}
```
and unnest the main logic

mukundansundar:
On a cluster where Dapr is not initialized: 
```
❌  No status returned. Is Dapr initialized in your cluster?
```

Otherwise: 
```
 NAME                   NAMESPACE    HEALTHY  STATUS                       REPLICAS  VERSION  AGE  CREATED              
  dapr-dashboard         dapr-system  False    Waiting (ContainerCreating)  1         0.2.0    7s   2020-09-27 00:52.47  
  dapr-operator          dapr-system  False    Running                      1         0.10.0   7s   2020-09-27 00:52.47  
  dapr-placement         dapr-system  False    Waiting (ContainerCreating)  1         0.10.0   7s   2020-09-27 00:52.47  
  dapr-sentry            dapr-system  False    Running                      1         0.10.0   7s   2020-09-27 00:52.47  
  dapr-sidecar-injector  dapr-system  False    Waiting (ContainerCreating)  1         0.10.0   7s   2020-09-27 00:52.47  
```

pkg/kubernetes/status.go
```diff
@@ -45,60 +61,68 @@ func Status() ([]StatusOutput, error) {
...
-					if p.Status.ContainerStatuses[0].State.Running == nil {
-						running = false
-						break
-					}
-
-					if p.Status.ContainerStatuses[0].Ready {
-						healthy = "True"
-					}
+			if err != nil {
+				print.FailureStatusEvent(os.Stdout, "Error getting status, %s", err.Error())
```
yaron2, line 70: Since this will exit the goroutine and not cancel the entire status command execution, I recommend turning this into a warning and providing addition data about which control plane label this error is for.

```suggestion
print.WarningStatusEvent(os.Stdout, "Failed to get status for %s: %s", label, err.Error())
```

yaron2 (changes requested):

yaron2 (approved):
lgtm

PR #477, opened 2020-10-02
Add UTs for publish and invoke

mukundansundar:
# Description

Make the publish, invoke testable. 

Make List implement a receiver. 

Add unit tests for publish and invoke. 

Add testutils package, defining utilities for testing. 

Output: 
Using quickstarts pub-sub 

```
./dist/darwin_amd64/release/dapr publish --topic A --pubsub pubsub --data '{ "message": "This is a test edge ver 2" }'
✅  Event published successfully

Output: 
== APP == A:  {

== APP ==   id: 'e0792cec-2e93-46b7-999c-2f0ff3dd8130',

== APP ==   source: 'node-subscriber',

== APP ==   type: 'com.dapr.event.sent',

== APP ==   specversion: '1.0',

== APP ==   datacontenttype: 'application/json',

== APP ==   data: { message: 'This is a test edge ver 2' },

== APP ==   subject: '00-747e47361df80a76cab6bb5ae16c3a4b-b767ff08348189ec-01',

== APP ==   topic: 'A',

== APP ==   pubsubname: 'pubsub'

== APP == }

```

Using hello-world quickstart

```
$ ./dist/darwin_amd64/release/dapr invoke --app-id nodeapp --method neworder --payload "{\"data\": { \"orderId\": \"42\" } }"
✅  App invoked successfully
$ ./dist/darwin_amd64/release/dapr invokeGet --app-id nodeapp --method order
{"orderId":"42"}
✅  HTTP Get to method order invoked successfully

$ ./dist/darwin_amd64/release/dapr invokePost --app-id nodeapp --method neworder --payload "{\"data\": { \"orderId\": \"43\" } }"
✅  HTTP Post to method neworder invoked successfully
$ ./dist/darwin_amd64/release/dapr invokeGet --app-id nodeapp --method order
{"orderId":"43"}
✅  HTTP Get to method order invoked successfully
```

## Issue reference

a part of the implementation for #485 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pkg/standalone/client.go
+++ b/pkg/standalone/client.go
@@ -13,7 +13,6 @@ type daprProcess struct {
 }
 
 type Client interface {
-	DaprProcess
 	InvokeGet(appID, method string) (string, error)
 	InvokePost(appID, method, payload string) (string, error)
 	Publish(topic, payload, pubsubName string) error
@@ -25,8 +24,4 @@ type Standalone struct {
 
 func NewStandaloneClient() Client {
 	return &Standalone{process: &daprProcess{}}
-}
-
-func (s *Standalone) List() ([]ListOutput, error) {
-	return s.process.List()
 }
--- a/pkg/standalone/client.go
+++ b/pkg/standalone/client.go
@@ -12,9 +12,13 @@ type DaprProcess interface {
 type daprProcess struct {
 }
 
+// Client is the interface the wraps all the methods exposed by the Dapr CLI.
 type Client interface {
-	InvokeGet(appID, method string) (string, error)
-	InvokePost(appID, method, payload string) (string, error)
+	// Get is used to invoke a method on a Dapr application with GET verb.
+	Get(appID, method string) (string, error)
+	// Post is used to invoke a method on a Dapr application with POST verb.
+	Post(appID, method, payload string) (string, error)
+	// Publish is used to publish event to a topic in a pubsub.
 	Publish(topic, payload, pubsubName string) error
 }
 
@@ -22,6 +26,6 @@ type Standalone struct {
 	process DaprProcess
 }
 
-func NewStandaloneClient() Client {
+func NewClient() Client {
 	return &Standalone{process: &daprProcess{}}
 }
--- a/pkg/standalone/client.go
+++ b/pkg/standalone/client.go
@@ -1,0 +1,32 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package standalone
+
+type DaprProcess interface {
+	List() ([]ListOutput, error)
+}
+
+type daprProcess struct {
+}
+
+type Client interface {
+	DaprProcess
+	InvokeGet(appID, method string) (string, error)
+	InvokePost(appID, method, payload string) (string, error)
+	Publish(topic, payload, pubsubName string) error
+}
+
+type Standalone struct {
+	process DaprProcess
+}
+
+func NewStandaloneClient() Client {
+	return &Standalone{process: &daprProcess{}}
+}
+
+func (s *Standalone) List() ([]ListOutput, error) {
+	return s.process.List()
+}
--- a/pkg/standalone/client.go
+++ b/pkg/standalone/client.go
@@ -14,10 +14,10 @@ type daprProcess struct {
 
 // Client is the interface the wraps all the methods exposed by the Dapr CLI.
 type Client interface {
-	// Get is used to invoke a method on a Dapr application with GET verb.
-	Get(appID, method string) (string, error)
-	// Post is used to invoke a method on a Dapr application with POST verb.
-	Post(appID, method, payload string) (string, error)
+	// InvokeGet is used to invoke a method on a Dapr application with GET verb.
+	InvokeGet(appID, method string) (string, error)
+	// InvokePost is used to invoke a method on a Dapr application with POST verb.
+	InvokePost(appID, method, payload string) (string, error)
 	// Publish is used to publish event to a topic in a pubsub.
 	Publish(topic, payload, pubsubName string) error
 }
--- a/pkg/standalone/invoke_test.go
+++ b/pkg/standalone/invoke_test.go
@@ -1,0 +1,108 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package standalone
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInvoke(t *testing.T) {
+	testCases := []struct {
+		name          string
+		errorExpected bool
+		errString     string
+		appID         string
+		method        string
+		lo            ListOutput
+		listErr       error
+		expectedPath  string
+		postResponse  string
+		resp          string
+	}{
+		{
+			name:          "list apps error",
+			errorExpected: true,
+			errString:     assert.AnError.Error(),
+			listErr:       assert.AnError,
+		},
+		{
+			name:          "appID not found",
+			errorExpected: true,
+			appID:         "invalid",
+			errString:     "app ID invalid not found",
+			lo: ListOutput{
+				AppID: "testapp",
+			},
+		},
+		{
+			name:   "appID found successful invoke empty response",
+			appID:  "testapp",
+			method: "test",
+			lo: ListOutput{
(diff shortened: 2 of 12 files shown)


pkg/standalone/list.go
```diff
@@ -41,6 +41,11 @@ type runData struct {
 	appCmd     string
 }
 
+func (s *Client) List() ([]ListOutput, error) {
+	// TODO remove this method and have only one List method with a Client receiver
+	return List()
```
mukundansundar, line 46: For now not changing the List method. PR only changes List() method in places that have been tested completely.

youngbupark:
This is a great pr for adding tests to cli !

One comment is that instead of having mock for httpclient, go has the standard [httptest](https://golang.org/pkg/net/http/httptest/) package. Then we may not need to maintain additional interface/mocks and create fakeclient. 

Any opinion @tcnghia @mchmarny @yaron2 ?

tcnghia:
Using httptest sounds good to me. It is widely used and probably will be better maintained than our own mocks (just from the numbers of users and develppers)

yaron2:
> This is a great pr for adding tests to cli !
> 
> One comment is that instead of having mock for httpclient, go has the standard [httptest](https://golang.org/pkg/net/http/httptest/) package. Then we may not need to maintain additional interface/mocks and create fakeclient.
> 
> Any opinion @tcnghia @mchmarny @yaron2 ?

Should not use a mock where Go's stdlib has solutions in place, so `httptest` is the way to go.

pkg/standalone/client.go
```diff
@@ -0,0 +1,32 @@
...
+
+type DaprProcess interface {
+	List() ([]ListOutput, error)
+}
+
+type daprProcess struct {
+}
+
+type Client interface {
+	DaprProcess
```
youngbupark, line 16: Do we need separate DaprProcess from Client ? Can we just have `Client` including List() ?

mukundansundar: We can, but the tests that are using list, it needs to be mocked ...If the client also includes list, then the mock client for the test needs to implement all methods to confirm to the interface.

pkg/standalone/client.go
```diff
@@ -0,0 +1,32 @@
...
+	Publish(topic, payload, pubsubName string) error
+}
+
+type Standalone struct {
+	process DaprProcess
+}
+
+func NewStandaloneClient() Client {
+	return &Standalone{process: &daprProcess{}}
+}
```
tcnghia, line 28: Checkout "Avoid stutter" here https://blog.golang.org/package-names (tl;dr in many Go packages the naming convention is `http.Server`, not `http.HttpServer` similarly `httptest.Server`, not `httptest.HttpTestServer`).  So in this case, calling `standalone.NewClient()` would be enough to have no ambiguity.

```suggestion
func NewClient() Client {
	return &Standalone{process: &daprProcess{}}
}
```

pkg/standalone/publish.go
```diff
@@ -43,19 +42,22 @@ func SendPayloadToTopic(topic, payload, pubsubName string) error {
...
 	r, err := http.Post(url, "application/json", bytes.NewBuffer(b))
+	if err != nil {
+		return err
+	}
 
-	if r != nil {
-		defer r.Body.Close()
+	if r.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected status code %d on publishing to %s in %s", r.StatusCode, topic, pubsubName)
 	}
```
tcnghia, line 51: May be move this to inside the `r != nil` block to avoid dereferencing `nil` pointer, since you already have a `nil` check block down at line 53?

However... from the doc of `http` package I think `err == nil => r != nil` (see examples here https://golang.org/pkg/net/http ). So you probably can (1) remove the `nil` check altogether, (2) move the `defer r.Body.Close()` to before the status code check, so we close the body even not 200.

pkg/standalone/publish_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+package standalone
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPubLish(t *testing.T) {
```
tcnghia, line 14:
```suggestion
func TestPublish(t *testing.T) {
```

pkg/standalone/invoke_test.go
```diff
@@ -0,0 +1,108 @@
...
+				AppID: "testapp",
+			},
+			expectedPath: "/v1.0/invoke/testapp/method/test",
+			postResponse: "test payload",
+			resp:         "successful invoke",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
```
tcnghia, line 64: I don't see `t.Parallel()` or using go-routine here, so and this copy is not necessary for sequential runs (See https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables , and different samples with and without `t.Parallel()` here https://blog.golang.org/subtests )

pkg/standalone/publish_test.go
```diff
@@ -0,0 +1,97 @@
...
+			payload:      "test payload",
+			expectedPath: "/v1.0/publish/testPubsubName/testTopic",
+			postResponse: "test payload",
+			lo: ListOutput{
+				AppID: "notempty",
+			},
+		},
+	}
+	for _, tc := range testCases {
+		tc := tc
```
tcnghia, line 76: Same comment as the other instance.. this is not needed for sequential `t.Run(...)`

pkg/standalone/invoke_test.go
tcnghia: Also didn't see `&tc` in the `t.Run` so I think we are ok without copying.

pkg/standalone/client.go
youngbupark: I presume that you split to two parts because DaprProcess maintains Process of dapr runtime and Clients uses http client to call Dapr API endpoint. then it makes sense, but I do not think we need to embed DaprProcess to Client. 

The actual struct implementing Client interface can have DaprProcess instead of embedding it to Client interface. What do you think?

mukundansundar: Yes that works too. Will change.

pkg/standalone/publish.go
mukundansundar: done

pkg/standalone/client.go
```diff
@@ -0,0 +1,31 @@
...
+	List() ([]ListOutput, error)
+}
+
+type daprProcess struct {
+}
+
+// Client is the interface the wraps all the methods exposed by the Dapr CLI.
+type Client interface {
+	// Get is used to invoke a method on a Dapr application with GET verb.
+	Get(appID, method string) (string, error)
```
youngbupark, line 18: Get and Post seem vague to me. InvokeServicePost or InvokeServiceGet would be explicit. 

cc/ @tcnghia

mukundansundar: Made it InvokeGet and InvokePost... as we do not have any other get or post methods ... also it is not direct post or get but service invocation ... so Invoke makes sense ...

youngbupark (approved):

pkg/standalone/client.go
```diff
@@ -0,0 +1,32 @@
...
+package standalone
+
+type DaprProcess interface {
+	List() ([]ListOutput, error)
+}
+
+type daprProcess struct {
+}
+
+type Client interface {
```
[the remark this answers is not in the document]
tcnghia: Can you please add documentation for this interface? BTW, if you can add Done, and resolve it would make it easier for reviewer to see the new changes.

Otherwise if you object to a review feedback you can also object but don't resolve, that way we can see the on-going conversation.

pkg/standalone/client.go
tcnghia: Also here, the review comment was resolved without a change or an objection.

pkg/standalone/invoke_test.go
tcnghia: Same here, resolved without objection or change.

pkg/standalone/publish_test.go
tcnghia: Same.

tcnghia: Ah, I was reviewing outdated PR. Sorry about that :)!

tcnghia (approved):

pkg/standalone/client.go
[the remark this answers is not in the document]
tcnghia: Sorry, commented on an outdated diff.

PR #489, opened 2020-10-12
Remove deprecated commands

### LINKED ISSUE #422: Remove deprecated cli commands
## Describe the proposal

Remove all deprecated cli commands/flags for v1.0 release

yaron2:
Closes #422


mukundansundar:
This is a breaking change where quickstarts also need to be modified ... 
I am all for removing `invoke` command in total ...
@msfussell  Any reasons not to remove `invoke` command? 
We get more consistent way to invoke ... `invokePost` and `invokeGet`

README.md
```diff
@@ -306,10 +306,18 @@ $ dapr run --app-id nodeapp --app-port 3000 node app.js
...
 
-Invoke your app:
+Invoke your app with POST:
 
 ```
-$ dapr invoke --app-id nodeapp --method mymethod
+$ dapr invokePost --app-id nodeapp --method mymethod
+```
+
+**Note: To invoke a remote gRPC enabled app, use `invokePost`.
```
mukundansundar, line 315: is this necessary here?

yaron2: The note? IMO yeah, because users can use invoke for gRPC too but might not know since the command has an HTTP construct in its name.

yaron2:
> This is a breaking change where quickstarts also need to be modified ...
> I am all for removing `invoke` command in total ...
> @msfussell Any reasons not to remove `invoke` command?
> We get more consistent way to invoke ... `invokePost` and `invokeGet`

Actually, now that we're talking about it, today we allow users to invoke other services using ANY verb.

Keeping a special command for each HTTP verb doesn't allow this ability to come through.
As such, I recommend we get rid of `invokeX` commands and remain with `invoke` with an optional `--verb` flag, with the default being `POST`.

WDYT @mukundansundar and @msfussell

yaron2:
> > This is a breaking change where quickstarts also need to be modified ...
> > I am all for removing `invoke` command in total ...
> > @msfussell Any reasons not to remove `invoke` command?
> > We get more consistent way to invoke ... `invokePost` and `invokeGet`
> 
> Actually, now that we're talking about it, today we allow users to invoke other services using ANY verb.
> 
> Keeping a special command for each HTTP verb doesn't allow this ability to come through.
> As such, I recommend we get rid of `invokeX` commands and remain with `invoke` with an optional `--verb` flag, with the default being `POST`.
> 
> WDYT @mukundansundar and @msfussell

I will refactor this PR based on my proposal.

yaron2:
I will start a new PR based on this.

msfussell:
LGTM

PR #498, opened 2020-10-22
Add ssl command to dapr run

### LINKED ISSUE #494: App app-ssl flag to CLI

yaron2:
Closes #494


youngbupark (approved):
This is actually ssl *insecure* channel. I wonder if we need to rename the option to be explicit.

yaron2:
> This is actually ssl _insecure_ channel. I wonder if we need to rename the option to be explicit.

Not rename the command, but I'll add it to the docs.

youngbupark (approved):

Issue #560, opened 2020-12-15
Add Upgrade CLI command to upgrade the Dapr control plane services


msfussell:
Dapr has an init command to install Dapr into different environements such as local machine (self hosted) and Kubernetes.
For Kubernetes this installs the Dapr control plane services.  Currently when upgrading the K8s Dapr control plane services, this is done via Helm. 

It would be good to also be able to upgrade the control plane via the Dapr CLI, with a different options (note that with Helm there may be more options and we should look at the other choices available here). 

dapr upgrade -k --runtime-version=1.0.0 --enable-ha --enable-mtls --namespace

I suspect that upgrade will only apply to Kubernetes and other prod environments (i.e. Not locally since init is good enough here and upgrade add little value)

PR #566, opened 2020-12-17
Add dapr upgrade command

yaron2:
This PR adds a `dapr upgrade -k` command that greatly simplifies the Dapr upgrade experience.

This PR handles CRD updates and certificate exports for zero downtime upgrades.

Closes https://github.com/dapr/cli/issues/560

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,18 @@ To remove Dapr from your Kubernetes cluster, use the `uninstall` command with `-
 $ dapr uninstall -k
 ```
 
+### Upgrade Dapr on Kubernetes
+
+To perform a zero downtime upgrade of the Dapr control plane:
+
+```
+$ dapr upgrade -k --runtime-version=1.0.0-rc.2  
+```
+
+The example above shows how to upgrade from your current version to version `1.0.0-rc.2`.
+
+*Note: do not use the `dapr upgrade` command if you're upgrading from 0.x versions of Dapr*
+
 ### Launch Dapr and your app
 
 The Dapr CLI lets you debug easily by launching both Dapr and your app.
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ $ dapr uninstall --network dapr-network
 
 The init command will install Dapr to a Kubernetes cluster. For more advanced use cases, use our [Helm Chart](https://github.com/dapr/dapr/tree/master/charts/dapr).
 
-*Note: The default namespace is dapr-system*
+*Note: The default namespace is dapr-system. The installation will appear under the name `dapr` for Helm*
 
 ```
 $ dapr init -k
--- a/pkg/kubernetes/upgrade.go
+++ b/pkg/kubernetes/upgrade.go
@@ -41,7 +41,7 @@ func Upgrade(conf UpgradeConfig) error {
 		return errors.New("dapr is not installed in your cluster")
 	}
 
-	print.InfoStatusEvent(os.Stdout, "Dapr control plane version %s detected", status[0].Version)
+	print.InfoStatusEvent(os.Stdout, "Dapr control plane version %s detected in namespace %s", status[0].Version, status[0].Namespace)
 
 	helmConf, err := helmConfig(status[0].Namespace)
 	if err != nil {
@@ -67,6 +67,9 @@ func Upgrade(conf UpgradeConfig) error {
 	}
 
 	var vals map[string]interface{}
+	var ca []byte
+	var issuerCert []byte
+	var issuerKey []byte
 
 	if mtls {
 		secret, sErr := getTrustChainSecret()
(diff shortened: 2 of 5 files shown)


pkg/kubernetes/upgrade.go
```diff
@@ -0,0 +1,123 @@
...
+	"configuration",
+	"subscription",
+}
+
+type UpgradeConfig struct {
+	RuntimeVersion string
+}
+
+func Upgrade(conf UpgradeConfig) error {
+	helmConf, err := helmConfig("dapr-system")
```
mukundansundar, line 30: Add `--namespace` flag as it can be a different namespace.

pkg/kubernetes/upgrade.go
```diff
@@ -0,0 +1,123 @@
...
+
+	upgradeClient := helm.NewUpgrade(helmConf)
+	upgradeClient.ResetValues = true
+	upgradeClient.Namespace = status[0].Namespace
+	upgradeClient.CleanupOnFail = true
+	upgradeClient.Wait = true
+
+	print.InfoStatusEvent(os.Stdout, "Starting upgrade...")
+
+	mtls, err := IsMTLSEnabled()
```
mukundansundar, line 64: should using old certs during upgrade be a flag? `--mtls` Only export certs during upgrade if that flag is provided.

mukundansundar (dismissed):

mukundansundar:
Please add unit tests for this change.

pkg/kubernetes/upgrade.go
yaron2: Its our recommended best practice to carry the certs over. In this PR, if mTLS is used, they are carried over.

pkg/kubernetes/upgrade.go
yaron2: updated

yaron2:
@mukundansundar please review again.

pkg/kubernetes/upgrade.go
```diff
@@ -0,0 +1,137 @@
...
+	status, err := sc.Status()
+	if err != nil {
+		return err
+	}
+
+	if len(status) == 0 {
+		return errors.New("dapr is not installed in your cluster")
+	}
+
+	print.InfoStatusEvent(os.Stdout, "Dapr control plane version %s detected", status[0].Version)
```
mukundansundar, line 44: probably print the Namespace also here? 
Can there be more than one dapr installation in a cluster?

pkg/kubernetes/upgrade.go
```diff
@@ -0,0 +1,137 @@
...
+			return err
+		}
+	}
+
+	err = applyCRDs(fmt.Sprintf("v%s", conf.RuntimeVersion))
+	if err != nil {
+		return err
+	}
+
+	if _, err = upgradeClient.Run("dapr", daprChart, vals); err != nil {
```
mukundansundar, line 93: make a note somewhere that `dapr` is the release name that it looks to upgrade in helm. 
If it is installed under a different release name in helm and then upgrade is used it won't work. 
Can a `--release-name` flag be provided?

pkg/kubernetes/upgrade.go
```diff
@@ -0,0 +1,137 @@
...
+		secret, sErr := getTrustChainSecret()
+		if sErr != nil {
+			return sErr
+		}
+
+		ca := secret.Data["ca.crt"]
+		issuerCert := secret.Data["issuer.crt"]
+		issuerKey := secret.Data["issuer.key"]
+
+		ha := highAvailabilityEnabled(status)
```
mukundansundar, line 81: Can't ha be independent of mttls? So shouldn't ha values in the chart be set independently?

mukundansundar (commented):
@yaron2 added some comments

pkg/kubernetes/upgrade.go
yaron2: Changed

pkg/kubernetes/upgrade.go
yaron2: I don't want to put Helm specific constructs in the command, also I don't think there's a need here: Helm users will know to list the installation with `helm list`, so specifying the release seems redundant here. If a user installed Dapr through Helm with a non-default release name, they should upgrade using Helm.

pkg/kubernetes/upgrade.go
yaron2: There can't be more than one Dapr installation in a cluster, but I added the namespace for good measure.

pkg/kubernetes/upgrade.go
mukundansundar: Can you make a note regarding that in README? 
Also there needs to be addition of upgrade command in the rc3 docs.

README.md
```diff
@@ -211,6 +211,18 @@ To remove Dapr from your Kubernetes cluster, use the `uninstall` command with `-
 $ dapr uninstall -k
 ```
 
+### Upgrade Dapr on Kubernetes
+
+To perform a zero downtime upgrade of the Dapr control plane:
```
mukundansundar, line 216: Make a note of the release name being `dapr` and if installed using helm, should use helm to upgrade.

yaron2: Added a note for the installation, where it belongs.

pkg/kubernetes/upgrade.go
yaron2: This is now changed, in last commit I changed the behavior and now the upgrade command will work with any Helm installation of Dapr.

mukundansundar (approved):

Issue #412, opened 2020-07-20
[E2E Scenario Tests] Initial e2e tests for DAPR CLI. Init, Uninstall


vinayada1:
## Describe the proposal
Test basic scenarios such as:-

- dapr init
- dapr init -k
- dapr uninstall

mukundansundar:
removing epic label since this issue just mentions e2e tests for init and uninstall in standalone and kubernetes mode.

PR #629, opened 2021-03-15
Add --wait amd --timeout flags for dapr init -k which are passed to Helm

### LINKED ISSUE #597: Add --wait flag to dapr init and dapr upgrade
## Describe the proposal

When using infrastructure as code in a DevOps pipeline my K8s cluster is created and I then run dapr init and helm install (of my application) if you do that too quickly no side cars will be injected into my pods because the dapr containers are still being created.

This required me to write custom logic to repeatedly calling dapr status and checking if I found the word `false` or not. This works but makes my pipeline fragile. If the output of dapr status were to change my pipeline would fail.

If dapr init and dapr upgrade offered a --wait flag that would cause the call to block until all the containers are healthy would eliminate the need for customer implementations of waiting logic.

## Release Note

RELEASE NOTE:

pkedy:
# Description

This PR adds:

* The `--wait` and `--timeout` flags to `dapr init`
* The `--timeout` flag to `dapr upgrade` and `dapr uninstall` (These two commands already wait so the desired behavior described in the issue #597 is already implemented).

This will be tested in an upcoming PR that will add e2e tests per #412.

## Issue reference

Resolves #597

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,12 @@ $ dapr init -k --enable-ha=true
 $ dapr init -k --enable-mtls=false
 ```
 
+#### Waiting for the Helm install to complete
+
+```
+$ dapr init -k --wait
+```
+
 #### Uninstall Dapr on Kubernetes
 
 To remove Dapr from your Kubernetes cluster, use the `uninstall` command with `--kubernetes` flag or the `-k` shorthand.
--- a/README.md
+++ b/README.md
@@ -211,10 +211,10 @@ $ dapr init -k --enable-ha=true
 $ dapr init -k --enable-mtls=false
 ```
 
-#### Waiting for the Helm install to complete
+#### Waiting for the Helm install to complete (default timeout is 300s/5m)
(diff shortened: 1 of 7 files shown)


mukundansundar:
@pkedy should the timeout be configurable?

pkedy:
@mukundansundar You read my mind.  The default is 300s/5m which is probably fine but having the option can't hurt.

mukundansundar:
@pkedy Mention that in the doc and in the argument docstring?

pkedy:
@mukundansundar Yes on the docs. No need to put the default value in the docstring since Cobra prints it out for you on a `--help`.

mukundansundar:
@pkedy Can you look at the build? It seems to be failing ...

artursouza (dismissed):

artursouza (approved):

PR #632, opened 2021-03-17
Adding some initial CLI e2e tests

### LINKED ISSUE #412: [E2E Scenario Tests] Initial e2e tests for DAPR CLI. Init, Uninstall

pkedy:
Co-authored-by: Mukundan Sundararajan <musundar@microsoft.com>

# Description

Added e2e tests for:
* `dapr init` (self-hosted)
* `dapr uninstall` (self-hosted)
* `dapr init -k` (kubernetes)
* `dapr uninstall -k` (kubernetes)

## Issue reference

Resolves #412 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


.github/workflows/kind_e2e.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+    - name: Get KinD info
+      run: |
+        kubectl cluster-info --context kind-kind
+        NODE_IP=$(kubectl get nodes \
+                  -lkubernetes.io/hostname!=kind-control-plane \
+                  -ojsonpath='{.items[0].status.addresses[?(@.type=="InternalIP")].address}')
+        echo "MINIKUBE_NODE_IP=$NODE_IP" >> $GITHUB_ENV
+
+    - name: Setup local registry
```
mukundansundar, line 126: do we really need this?

.github/workflows/kind_e2e.yaml
```diff
@@ -0,0 +1,140 @@
+# ------------------------------------------------------------
+# Copyright (c) Microsoft Corporation and Dapr Contributors.
+# Licensed under the MIT License.
+# ------------------------------------------------------------
+name: E2E tests on KinD
```
mukundansundar, line 5: This should be Run E2E tests ?
Kind is just one platform that is being tested?

tests/e2e/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,307 @@
...
+func uninstall() (string, error) {
+	daprPath := getDaprPath()
+
+	return spawn.Command(daprPath,
+		"uninstall", "-k",
+		"-n", daprNamespace,
+		"--log-as-json")
+}
+
+func testUninstall(t *testing.T) {
```
mukundansundar, line 87: should a check be added to test CRD and role removal also?

.github/workflows/kind_e2e.yaml
pkedy: I'll try removing it

.github/workflows/kind_e2e.yaml
pkedy: I'll rename to include "Self-Hosted"

tests/e2e/kubernetes/kubernetes_test.go
pkedy: I will do some magic to test reverse polarity.

mukundansundar:
@pkedy I think one thing missing here is that Self Hosted mode is only tested in linux. 

My opinion would be splitting the e2e tests into two parts 
1. Self Hosted run: matrix Windows, darwin and linux OS with different archs
2. KinD k8s : test different versions of k8s support.(though is that needed since dapr is already tested with different k8s versions in e2e test) I think a simple workflow with one major version of k8s should be enough? @artursouza @yaron2 Please chime in on this.

.github/workflows/kind_e2e.yaml
pkedy: Worked without it

tests/e2e/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,326 @@
...
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/dapr/cli/pkg/kubernetes"
+	"github.com/dapr/cli/tests/e2e/spawn"
+
+	k8s "k8s.io/client-go/kubernetes"
+
+	//  azure auth
+	_ "k8s.io/client-go/plugin/pkg/client/auth/azure"
```
artursouza, line 32: Why do we need support for all these auth in the e2e test?

tests/e2e/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,326 @@
+// +build e2e
+
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation and Dapr Contributors.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package standalone_test
```
artursouza, line 8: k8s namespace?

tests/e2e/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+	return config, nil
+}
+
+func homeDir() string {
+	if h := os.Getenv("HOME"); h != "" {
+		return h
+	}
+	return os.Getenv("USERPROFILE") // windows
+}
```
artursouza, line 326: Good start. There are many more scenarios but start here is good.

.github/workflows/kind_e2e.yaml
```diff
@@ -0,0 +1,124 @@
...
+        - v1.18.8
+        - v1.19.1
+        mode:
+        - ha
+        - non-ha
+        # Map between K8s and KinD versions.
+        # This is attempting to make it a bit clearer what's being tested.
+        # See: https://github.com/kubernetes-sigs/kind/releases/tag/v0.9.0
+        include:
+        - k8s-version: v1.17.11
```
artursouza, line 52: I think it is OK to have E2E test for multiple versions. This is good in case a future change unintentionally breaks compatibility with an older version.

artursouza:
> @pkedy I think one thing missing here is that Self Hosted mode is only tested in linux.
> 
> My opinion would be splitting the e2e tests into two parts
> 
> 1. Self Hosted run: matrix Windows, darwin and linux OS with different archs
> 2. KinD k8s : test different versions of k8s support.(though is that needed since dapr is already tested with different k8s versions in e2e test) I think a simple workflow with one major version of k8s should be enough? @artursouza @yaron2 Please chime in on this.

I think that testing many different K8s version is actually better because it can identify an issue where a change unintentionally breaks compatibility with an older version - we want to catch that.

artursouza (approved):

Issue #631, opened 2021-03-16
CLI returns non-zero when dapr stop is used on Windows


wcs1only:
## Expected Behavior

When ```dapr stop --app-id foo``` the dapr process which launched ```foo``` should gracefully clean up and exit with a return code of 0. 

Also we expect the following to be printed:
```
ℹ️  
terminated signal received: shutting down
✅  Exited Dapr successfully
✅  Exited App successfully
```

## Actual Behavior

On Windows the dapr process returns 1. It does not print the expected success message.

## Steps to Reproduce the Problem

On cygwin/bash:

```bash
$ dapr run run --app-id nodeapp --app-port 3000 --dapr-http-port 3500 ; echo $?
Starting Dapr with id nodeapp. HTTP Port: 3500. gRPC Port: 61850
time="2021-03-16T22:33:36.7871694Z" level=info msg="starting Dapr Runtime -- version edge -- commit v1.0.0-rc.4-49-g7a47aa3" app_id=nodeapp instance=win scope=dapr.runtime type=log ver=edge
time="2021-03-16T22:33:36.7888331Z" level=info msg="log level set to: info" app_id=nodeapp instance=win scope=dapr.runtime type=log ver=edge
time="2021-03-16T22:33:36.7898677Z" level=info msg="metrics server started on :61851/" app_id=nodeapp instance=win scope=dapr.metrics type=log ver=edge
time="2021-03-16T22:33:36.7969606Z" level=info msg="standalone mode configured" app_id=nodeapp instance=win scope=dapr.runtime type=log ver=edge
time="2021-03-16T22:33:36.7971191Z" level=info msg="app id: nodeapp" app_id=nodeapp instance=win scope=dapr.runtime type=log ver=edge
time="2021-03-16T22:33:36.7977909Z" level=info msg="mTLS is disabled. Skipping certificate request and tls validation" app_id=nodeapp instance=win scope=dapr.runtime type=log ver=edge
time="2021-03-16T22:33:36.8187992Z" level=info msg="local service entry announced: nodeapp -> 172.16.0.6:61855" app_id=nodeapp instance=win scope=dapr.contrib type=log ver=edge
time="2021-03-16T22:33:36.820799Z" level=info msg="Initialized name resolution to standalone" app_id=nodeapp instance=win scope=dapr.runtime type=log ver=edge
[...]

PR #638, opened 2021-03-19
Use a named event for dapr stop on Windows (#631)

wcs1only:
# Description

Windows doesn't have an easy SIGTERM equivalent, so we use a named event to signal Windows to stop Dapr gracefully.

## Issue reference

Please reference the issue this PR will close: #631 

## Manual Testing
Without this change (run `dapr stop --app-id` in another terminal):

```bash
You're up and running! Both Dapr and your app logs will appear here.


C:\Users\Charlie\quickstarts\hello-world>echo %ERRORLEVEL%
1
```

With this change (run `dapr stop --app-id` in another terminal):

```bash
You're up and running! Both Dapr and your app logs will appear here.


terminated signal received: shutting down
Exited Dapr successfully
Exited App successfully

C:\Users\Charlie\quickstarts\hello-world>echo %ERRORLEVEL%
0
```
## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/shutdown_windows.go
+++ b/cmd/shutdown_windows.go
@@ -1,0 +1,34 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation and Dapr Contributors.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/dapr/cli/pkg/print"
+	"golang.org/x/sys/windows"
+)
+
+func setupShutdownNotify(sigCh chan os.Signal){
+	//This will catch Ctrl-C
+	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
+
+	// Unlike Linux/Mac, you can't just send a SIGTERM from another process
+	// In order for 'dapr stop' to be able to signal gracefully we must use a named event in Windows
+	go func() {
+		eventName, _ := syscall.UTF16FromString(fmt.Sprintf("dapr_cli_%v", os.Getpid()))
+		eventHandle, _ := windows.CreateEvent(nil, 0, 0, &eventName[0])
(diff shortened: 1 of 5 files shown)


cmd/shutdown_windows.go
```diff
@@ -0,0 +1,34 @@
+// ------------------------------------------------------------
```
artursouza, line 1: Shouldn't this also have the +build windows header?

cmd/shutdown_windows.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/dapr/cli/pkg/print"
+	"golang.org/x/sys/windows"
+)
+
+func setupShutdownNotify(sigCh chan os.Signal){
+	//This will catch Ctrl-C
+	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
+
+	// Unlike Linux/Mac, you can't just send a SIGTERM from another process
+	// In order for 'dapr stop' to be able to signal gracefully we must use a named event in Windows
```
artursouza, line 23: Ha! So, that is why we see process leaks on Windows.

pkg/standalone/stop_windows.go
```diff
@@ -0,0 +1,38 @@
+// ------------------------------------------------------------
```
artursouza, line 1: header here too?

artursouza (changes requested):

cmd/shutdown_windows.go
wcs1only: Adding the suffix `_windows.go` to the filename achieves the same thing: https://golang.org/cmd/go/#hdr-Build_constraints

cmd/shutdown_windows.go
wcs1only: See above.

pkg/standalone/stop_windows.go
wcs1only: See above

artursouza (approved):

Issue #645, opened 2021-03-26
dapr start app error need to stop the running dapr process


marviniter:
## Expected Behavior
dapr run app err need to stop the running daprd process

## Actual Behavior
dapr run  to start app with not found command cause "exec: "-": executable file not found in $PATH" error then exit but the daprd proccess is running and dapr cli can not find and stop it, need kill the daprd process by cmd.

## Steps to Reproduce the Problem
dapr run --app-id testpubsub  --dapr-http-port 3500 - --components-path ./components/
```
ℹ️  Dapr sidecar is up and running.
❌  exec: "-": executable file not found in $PATH
```
but the daprd is still running.

## Release Note

RELEASE NOTE:

mukundansundar:
closed by #646

PR #646, opened 2021-03-26
when start app error try to stop daprd process

marviniter:
# Description
Receive appRunning channel value if the value false then try to stop daprd and exit

## Issue reference
close:#645 
## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -150,13 +150,15 @@ var RunCmd = &cobra.Command{
 			stdErrPipe, pipeErr := output.AppCMD.StderrPipe()
 			if pipeErr != nil {
 				print.FailureStatusEvent(os.Stdout, fmt.Sprintf("Error creating stderr for App: %s", err.Error()))
-				os.Exit(1)
+				appRunning <- false
+				return
 			}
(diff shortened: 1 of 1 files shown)


cmd/run.go
```diff
@@ -176,13 +178,24 @@ var RunCmd = &cobra.Command{
...
 			}
 
 			appRunning <- true
 		}()
 
-		<-appRunning
+		appRunStatus := <-appRunning
+		if !appRunStatus {
+			// Start App failed, try to stop Dapr and exit.
+			err = output.DaprCMD.Process.Kill()
```
mukundansundar, line 191: @wcs1only Do you think this is sufficient in windows for the process to be killed?

wcs1only: Yeah, should be fine. In general, the CLI can kill a process it started. The issue we fixed a few weeks ago was related to running `dapr stop` from another terminal (killing a process that isn't your child is not something you can do gracefully on Windows).

wcs1only (approved):

PR #683, opened 2021-04-09
Add some negative self-hosted e2e. On error in stop exit with error

mukundansundar:
# Description

Add some negative self-hosted e2e. On error in stop exit with error

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/tests/e2e/standalone/standalone_test.go
+++ b/tests/e2e/standalone/standalone_test.go
@@ -58,6 +58,63 @@ func TestStandaloneInstall(t *testing.T) {
 	}
 }
 
+func TestNegativeScenarios(t *testing.T) {
+	// Ensure a clean environment
+	uninstall()
+	daprPath := getDaprPath()
+
+	homeDir, err := os.UserHomeDir()
+	require.NoError(t, err, "expected no error on querying for os home dir")
+
+	t.Run("run without install", func(t *testing.T) {
+		output, err := spawn.Command(daprPath, "run", "test")
+		require.NoError(t, err, "expected no error status on run without install")
+		path := filepath.Join(homeDir, ".dapr", "components")
+		require.Contains(t, output, path+": no such file or directory", "expected output to contain message")
+	})
+
+	t.Run("list without install", func(t *testing.T) {
+		output, err := spawn.Command(daprPath, "list")
+		require.NoError(t, err, "expected no error status on list without install")
+		require.Equal(t, "No Dapr instances found.\n", output)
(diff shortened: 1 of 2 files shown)


tests/e2e/standalone/standalone_test.go
```diff
@@ -58,6 +58,63 @@ func TestStandaloneInstall(t *testing.T) {
...
+	// Ensure a clean environment
+	uninstall()
+	daprPath := getDaprPath()
+
+	homeDir, err := os.UserHomeDir()
+	require.NoError(t, err, "expected no error on querying for os home dir")
+
+	t.Run("run without install", func(t *testing.T) {
+		output, err := spawn.Command(daprPath, "run", "test")
+		require.NoError(t, err, "expected no error status on run without install")
```
mukundansundar, line 71: I think a separate issue has to be created so that we properly exit with error on error from run. 
This needs to be separate task since the scenarios for run command are a bit more complex.

artursouza (approved):

tests/e2e/standalone/standalone_test.go
```diff
@@ -58,6 +58,63 @@ func TestStandaloneInstall(t *testing.T) {
...
+		require.Contains(t, output, "The id for your application, used for service discovery", "expected usage to be printed")
+	})
+
+	t.Run("uninstall without install", func(t *testing.T) {
+		output, err := spawn.Command(daprPath, "uninstall", "--all")
+		require.NoError(t, err, "expected no error on uninstall without install")
+		require.Contains(t, output, "Removing Dapr from your machine...", "expected output to contain message")
+		path := filepath.Join(homeDir, ".dapr", "bin")
+		require.Contains(t, output, "WARNING: "+path+" does not exist", "expected output to contain message")
+		require.Contains(t, output, "WARNING: dapr_placement container does not exist", "expected output to contain message")
```
artursouza, line 109: These tests will fail when log message changes. Ideally, those could be a shared constant.

tests/e2e/standalone/standalone_test.go
mukundansundar: Added https://github.com/dapr/cli/issues/684

tests/e2e/standalone/standalone_test.go
wcs1only: I mean... changing them later is not hard.

wcs1only (approved):

PR #685, opened 2021-04-09
Add some negative tests for invoke and publish

wcs1only:
# Description

Add a few negative e2e test cases for the --data-file option in invoke & publish.
## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


mukundansundar (approved):
i think asserting the output is also something that should be done

wcs1only:
> i think asserting the output is also something that should be done

I thought about that, but the file not found error is os specific, I think. I suppose we could just do a contains for the parts that we're echoing...

mukundansundar (approved):

Issue #688, opened 2021-04-09
Save test results as a build artifact of a test run


wcs1only:
## Describe the proposal

We've long struggled with visibility in flaky end-to-end tests. One of the big problems is that we really have no way to see test failures over time other than to click on individual test runs and manually scan the output. To really be able to correctly track test flakiness over time, we need to be saving test results in a programatic way. An easy first step might be to store the test results' json output as a build artifact of each run. That will at least give us a way to programmatically scrape the data and produce reports.

## Release Note

RELEASE NOTE: NOT APPLICABLE

PR #689, opened 2021-04-09
Upload test results as a github actions run artifact

wcs1only:
# Description

Save the test output for e2e and unit tests as a json file attached to the build job. If we want to scan this data and build a report it will be much much easier than clicking through the github UI.

You can find the JSON test output attached to each run:
https://github.com/dapr/cli/actions/runs/734751808

## Issue reference

Please reference the issue this PR will close: #688

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,8 @@ endif
 export GOOS ?= $(TARGET_OS_LOCAL)
 export BINARY_EXT ?= $(BINARY_EXT_LOCAL)
 
+TEST_OUTPUT_FILE ?= test_output.json
+
 # Use the variable H to add a header (equivalent to =>) to informational output
 H = $(shell printf "\033[34;1m=>\033[0m")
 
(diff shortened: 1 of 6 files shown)


wcs1only:
Going to do the same for dapr/dapr next

Makefile
```diff
@@ -120,19 +122,26 @@ endif
 .PHONY: release
 release: build archive
 
+.PHONY: test-deps
+test-deps:
+	# The desire here is to download this test dependency without polluting go.mod
+	# In golang >=1.16 there is a new way to do this with `go install gotest.tools/gotestsum@latest`
```
mukundansundar, line 128: The CLI is already changed to Go 1.16. Why do we have to do it the old way for now? The workflow runs with Go 1.16 so `go install` should work?

wcs1only: It will, I didn't want to force our contributors to upgrade, especially since we haven't even done a release with 1.16 yet.

I can change it, if you think it is necessary. I just want to merge this PR so we can start collecting data.

artursouza: If the old way works for both, I am fine with that.

artursouza (approved):

PR #696, opened 2021-04-23
`dapr run` shut down app and daprd gracefully

### LINKED ISSUE #695: `dapr run` CTRL-C kills both the app and daprd
## Expected Behavior

Pressing CTRL-C when running `dapr run` should send interrupt signals to both the application and daprd so that they shutdown gracefully. This would also allow daprd to close components down also per dapr/dapr#3088.

## Actual Behavior

Both the application and daprd are killed immedatiately.

Use `dapr run` and press CTRL-C once the application is initialized.

## Release Note

RELEASE NOTE: **FIX** CTRL-C when using `dapr run` sends interrupt signals to the application and daprd instead of killing them.

pkedy:
Send the running app and dapr an interrupt signal instead of killing them immediately. This allows the processes to clean up appropriately.

# Description

This PR makes a change to send interrupt signals to the application and daprd processes and waits for them to shutdown gracefully instead of killing them immediately. The user will see the 5 second delay for outstanding requests to finish.

## Issue reference

Resolves #695

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


cmd/interrupt_windows.go
```diff
@@ -0,0 +1,49 @@
...
+	if r == 0 {
+		return err
+	}
+
+	return nil
+}
+
+func suppressCtrlBreak(err error) error {
+	if exitErr, ok := err.(*exec.ExitError); ok {
+		if exitErr.ExitCode() == 3221225786 {
```
mukundansundar, line 43: @pkedy What is this number?

mukundansundar (commented):
@pkedy Is there a specific reason to load a DLL file to call an event in windows? Is there any other simpler way to do this? 
@artursouza @wcs1only Thoughts on this change?

cmd/interrupt_windows.go
pkedy: I'm adding a constant for this but that is the exit/status code for Ctrl-Break.

mukundansundar: Also please add any documentation links for references as comments.

tests/e2e/standalone/standalone_test.go
```diff
@@ -359,10 +359,9 @@ func testRun(t *testing.T) {
 		output, err := spawn.Command(daprPath, "run", "--dapr-http-port", "9999", "--", "bash", "-c", "curl -v http://localhost:9999/v1.0/shutdown; sleep 10; exit 1")
 		t.Log(output)
 		require.NoError(t, err, "run failed")
-		assert.Contains(t, output, "Exited App successfully", "App should be shutdown before it has a chance to return non-zero")
+		//assert.Contains(t, output, "Exited App successfully", "App should be shutdown before it has a chance to return non-zero")
```
artursouza, line 362: Why is this commented out?

artursouza (changes requested):
nit: there is a commented out line of code. If not needed, please, remove.

artursouza:
@mukundansundar Depending on a DLL is a problem. We should avoid that. I saw that the latest version does not have that so I am fine with this.

tests/e2e/standalone/standalone_test.go
pkedy: I was trying to keep that check, but I could not find a way too. I'll put a comment there instead.

mukundansundar: Is the output different from what was expected?

pkedy: Its been a while since I was working on this, but the issue is that if you send `curl` a CRTL-C/interrupt, it still comes back with a non-zero exit code depending on the OS. We need a process that exits with `0` even if sent an interrupt/ctrl-c.

wcs1only: Have you tried `trap` ?

pkedy: Thought about it but avoided that in case we can get Windows docker working for GitHub actions. (That would also assume Windows would have some form of curl installed)

pkedy: In the grand scheme of things, Im not :100: sure we really should test the exit code of the app, just that it ran and existed. In Daprd, it has explicit code to exit with `0` if it is interrupted. Most apps don't do that.

tests/e2e/standalone/standalone_test.go
```diff
@@ -359,10 +359,11 @@ func testRun(t *testing.T) {
 		output, err := spawn.Command(daprPath, "run", "--dapr-http-port", "9999", "--", "bash", "-c", "curl -v http://localhost:9999/v1.0/shutdown; sleep 10; exit 1")
 		t.Log(output)
 		require.NoError(t, err, "run failed")
-		assert.Contains(t, output, "Exited App successfully", "App should be shutdown before it has a chance to return non-zero")
+		// It would be ideal to check that the spawned app existed with a non-zero
+		// exit code, however when sending a SIGINT/CTRL-C to curl, it does not.
+		// In the future we could find a program that returns 0 in this scenario.
```
wcs1only, line 364: You might as well not have have this test at all without this check. The exit 1 is how we tell the difference between "shutdown received and respected" vs "shutdown ignored". 

```
willardstanley@Willards-MacBook-Pro cli % bash -c 'trap "exit 0" SIGINT; sleep 200; exit 12'; echo $? 
^C0
```

tests/e2e/standalone/standalone_test.go
wcs1only: > In the grand scheme of things, Im not 💯 sure we really should test the exit code of the app, just that it ran and existed. In Daprd, it has explicit code to exit with `0` if it is interrupted. Most apps don't do that.

That's fine if they don't do that. There is no harm in telling the user what the return code was.

tests/e2e/standalone/standalone_test.go
wcs1only: Litmus test: If you mangle the curl URL, so that the shutdown API doesn't get called, does the test fail like it should?

cmd/interrupt.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+// interruptProcess sends an Interrupt signal to the process.
+func interruptProcess(proc *os.Process) error {
+	return proc.Signal(os.Interrupt)
```
wcs1only, line 13: If you sent SIGTERM bash wouldn't catch it in the test. I think SIGTERM makes more sense if we consider that a dapr app is likely to be a daemon...

wcs1only:
Did you run the E2E tests on windows? When I run them on my Windows box for this change, they fail quite hard. 

I suspect that you need to set the CREATE_NEW_PROCESS_GROUP on the processes that the CLI spawns in order to safely send a CTL_BREAK to them. See https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags for more info.

cmd.SysProcAttr = &syscall.SysProcAttr{
		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP,
}

mukundansundar:
@pkedy Is this PR still valid?

pkedy:
@mukundansundar Yes, I think this is still valid. Last I looked at it the tests were hanging. Now that #835 is in, I'll take another look.

mukundansundar:
@pkedy Can this PR be reopened and will you be looking into this?

PR #701, opened 2021-04-27
Verify GitHub auth

pkedy:
Just testing


pkedy:
Fix confirmed. Closing

Issue #664, opened 2021-04-04
MTLS automatically enabled on running `dapr upgrade` on a cluster with MTLS disabled


mukundansundar:
## Expected Behavior

`dapr upgrade` command does not change the mtls setting from disabled to enabled. 

## Actual Behavior

`dapr upgrade` command turns on mtls even if it was disabled previously. 

## Steps to Reproduce the Problem

<details>
<summary> Run `dapr init` with version 1.0.0 with mtls disabled </summary>

```bash

$ dapr init -k --runtime-version 1.0.0 --enable-mtls=false
⌛  Making the jump to hyperspace...
ℹ️  Note: To install Dapr using Helm, see here: https://docs.dapr.io/getting-started/install-dapr-kubernetes/#install-with-helm-advanced

✅  Deploying the Dapr control plane to your cluster...
✅  Success! Dapr has been installed to namespace dapr-system. To verify, run `dapr status -k' in your terminal. To get started, go here: https://aka.ms/dapr-getting-started

$ dapr mtls -k
Mutual TLS is disabled in your Kubernetes cluster
$ k get configurations.dapr.io -n dapr-system daprsystem -o yaml
```

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  annotations:
    meta.helm.sh/release-name: dapr
    meta.helm.sh/release-namespace: dapr-system
  creationTimestamp: "2021-04-04T17:26:34Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: Helm
  managedFields:
  - apiVersion: dapr.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:meta.helm.sh/release-name: {}
          f:meta.helm.sh/release-namespace: {}
        f:labels:
          .: {}
          f:app.kubernetes.io/managed-by: {}
      f:spec:
        .: {}
        f:metric:
          .: {}
          f:enabled: {}
        f:mtls:
          .: {}
          f:allowedClockSkew: {}
          f:enabled: {}
          f:workloadCertTTL: {}
    manager: Go-http-client
    operation: Update
    time: "2021-04-04T17:26:34Z"
  name: daprsystem
  namespace: dapr-system
  resourceVersion: "178509"
  uid: c4f9664c-ae4e-442e-baad-5ec2198bb559
spec:
  metric:
    enabled: true
  mtls:
    allowedClockSkew: 15m
    enabled: false
[...]

PR #706, opened 2021-05-08
fix reset values on upgrade

yaron2:
Closes https://github.com/dapr/cli/issues/664.

Reference: https://medium.com/@kcatstack/understand-helm-upgrade-flags-reset-values-reuse-values-6e58ac8f127e

### DIFF
--- a/pkg/kubernetes/upgrade.go
+++ b/pkg/kubernetes/upgrade.go
@@ -84,7 +84,6 @@ func Upgrade(conf UpgradeConfig) error {
 	var issuerKey []byte
 
 	if mtls {
-		fmt.Println("aaaaaa")
 		secret, sErr := getTrustChainSecret()
 		if sErr != nil {
 			return sErr
--- a/pkg/kubernetes/upgrade.go
+++ b/pkg/kubernetes/upgrade.go
@@ -66,6 +66,7 @@ func Upgrade(conf UpgradeConfig) error {
 	}
 
 	upgradeClient := helm.NewUpgrade(helmConf)
+	upgradeClient.ReuseValues = true
 	upgradeClient.Namespace = status[0].Namespace
 	upgradeClient.CleanupOnFail = true
 	upgradeClient.Wait = true
--- a/pkg/kubernetes/upgrade.go
+++ b/pkg/kubernetes/upgrade.go
@@ -66,7 +66,7 @@ func Upgrade(conf UpgradeConfig) error {
 	}
 
 	upgradeClient := helm.NewUpgrade(helmConf)
-	upgradeClient.ReuseValues = true
+	upgradeClient.ResetValues = true
 	upgradeClient.Namespace = status[0].Namespace
 	upgradeClient.CleanupOnFail = true
 	upgradeClient.Wait = true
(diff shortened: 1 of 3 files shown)


mukundansundar:
@yaron2 Please fix the upgrade e2e test failures.

yaron2:
@mukundansundar can you advise on the failing test here?

mukundansundar:
> @mukundansundar can you advise on the failing test here?

@yaron2 The Failing test is not related to this PR. I am looking into it and seeing if it can be fixed. The issue is that for the Launch Daemon file location source in the command that is being run for Docker in Mac, the file is not available in the location. That file is needed for further Docker Desktop instal in Mac.

mukundansundar:
@yaron2 Based on the doc https://medium.com/@kcatstack/understand-helm-upgrade-flags-reset-values-reuse-values-6e58ac8f127e
If RestValues flag is not set, it will not apply the new values in the new chart during upgrade command run. But isn't that what we want? 
Additionally, if the command is run with a --set flag in dapr, which is passed onto Helm, it seems that reset-values is automatically true. Which would end up resetting MTLS value to the the default true in the chart. 

My question is should Dapr upgrade command emulate what is happening with Helm upgrade command, if so shouldn't reset-values be set to true ? 

@artursouza @pkedy Thoughts?

yaron2:
> If RestValues flag is not set, it will not apply the new values in the new chart during upgrade command run. But isn't that what we want?

In this PR, the ResetValues flag is set to `true`, meaning it will applies new values in the chart, which is what we want.

pkg/kubernetes/upgrade.go
```diff
@@ -66,7 +66,6 @@ func Upgrade(conf UpgradeConfig) error {
 	}
 
 	upgradeClient := helm.NewUpgrade(helmConf)
-	upgradeClient.ResetValues = true
```
mukundansundar, line 69: > > If RestValues flag is not set, it will not apply the new values in the new chart during upgrade command run. But isn't that what we want?
> 
> In this PR, the ResetValues flag is set to `true`, meaning it will applies new values in the chart, which is what we want.

I think I am missing something here. This PR removes the line which sets the Reset Values to "true" ?

yaron2: I meant ReuseValues, and apparently I didn't push it.. until now.

yaron2:
I pushed the missing `ReuseValues=true` commit.

I tested the following scenarios:

1. Install without mTLS, without --set
2. Install without mTLS, with --set

Then did upgrade for both, with and without --set. The original values were kept while changes with `--set` were updated.

mukundansundar:
> I pushed the missing `ReuseValues=true` commit.
> 
> I tested the following scenarios:
> 
> 1. Install without mTLS, without --set
> 2. Install without mTLS, with --set
> 3. Install with mTLS (default)
> 
> Then did upgrade for both, with and without --set. The original values were kept while changes with `--set` were updated.

@yaron2 with the reuse values change the upgrade e2e is failing on the upgrade step and follow up status check.

yaron2:
> > I pushed the missing `ReuseValues=true` commit.
> > I tested the following scenarios:
> > 
> > 1. Install without mTLS, without --set
> > 2. Install without mTLS, with --set
> > 3. Install with mTLS (default)
> > 
> > Then did upgrade for both, with and without --set. The original values were kept while changes with `--set` were updated.
> 
> @yaron2 with the reuse values change the upgrade e2e is failing on the upgrade step and follow up status check.

So I looked into the Helm code to see what it does: the pods will restart even though no values have changed, and indeed the old control plane version will be used and not the one in the target chart. That's misleading.. and in any case the result is not correct.

I returned the `ResetValues=true` behavior with a `--set` command to disable mTLS if it wasn't previously used.

artursouza (approved):

mukundansundar (approved):

PR #714, opened 2021-05-19
Apply CRDs on init -k

yaron2:
This PR applies the Dapr CRDs during the `init` process, the same as `upgrade`.


wcs1only (approved):
Approved with the caveat that we add a CLI E2E test for this.

Issue #719, opened 2021-05-19
`dapr init -k` broken in 1.2.0-rc.1


mukundansundar:
## Expected Behavior

dapr initializes the latest version of runtime in kubernetes

## Actual Behavior

```
error: unable to read URL "https://raw.githubusercontent.com/dapr/dapr/vlatest/charts/dapr/crds/components.yaml", server reported 404 Not Found, status code=404
```

Suspect some issue with the https://github.com/dapr/cli/pull/714 change.

## Steps to Reproduce the Problem

Run `dapr init -k` in 1.2.0-rc.1

## Release Note

RELEASE NOTE:

PR #721, opened 2021-05-19
Fix init -k without --runtime-version

yaron2:
Fixes https://github.com/dapr/cli/issues/719.


mukundansundar:
@yaron2 Can you base it off of release-1.2 branch ?

wcs1only (approved):
Approving so we can cut an rc, but please add E2E tests for this before release day.

yaron2:
> Approving so we can cut an rc, but please add E2E tests for this before release day.

@mukundansundar is on point for the E2E scenario.

Issue #647, opened 2021-03-28
Support to get api response and replace base download url


newbe36524:
## Describe the proposal

Support to get github release api json from specified path instead of api.github.com
Support to download binary files from specified asset host instead of github.com

for example, developer could run this script to download and install dapr in China to get better network expreince:

```powershell
./install.ps1 -DaprReleaseJson "https://gitee.com/dapr-cn/dapr-bin-mirror/raw/main/cli/releases.json" -CustomAssetPrefix "https://ghproxy.com/https://github.com/dapr/cli/releases/download/"
```

refer to #640

## Release Note

RELEASE NOTE: **ADD** Support to  get api response and replace base download url

PR #732, opened 2021-06-02
Add CI/CD Test for Windows Powershell Installer

berndverst:
# Description
Add CI/CD Test for Windows Powershell Installer

## Issue reference

To verify the work in #647

### DIFF
--- a/.github/workflows/installdaprwin.yml
+++ b/.github/workflows/installdaprwin.yml
@@ -35,6 +35,7 @@ jobs:
         uses: Amadevus/pwsh-script@v2.0.1
         with:
           script: |
+            $Env:Path += ";c:\dapr"
             $output=dapr --version
             $runtimeVerified = False
             $cliVerified = False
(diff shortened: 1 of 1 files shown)


berndverst:
@mukundansundar - per our discussion. Feel free to change

mukundansundar (commented):
@berndverst Can you run dapr --version to check for proper install? Check runtime version is n/a and CLI version is set

berndverst:
@mukundansundar done.

.github/workflows/installdaprwin.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      # Install Dapr
+      - name: Install DAPR CLI
+        run: powershell "install/install.ps1"
+      
+      - name: Verify installed CLI version
+        uses: Amadevus/pwsh-script@v2.0.1
+        with:
+          script: |
+            $Env:Path += ";c:\dapr"
```
mukundansundar, line 38: @berndverst Isn't the script already setting the env path?

berndverst: It doesn't work without this line because in Github Actions the Path isn't updated between steps unless you explicitly tell it to do so in my experience.

All I can say is that this was required in my testing and it certainly doesn't hurt. So why not approve unless there is something else?

mukundansundar (approved):

Issue #736, opened 2021-06-08
Dapr List Displays Dapr Dashboards As Empty Services


jasonviviano:
## Expected Behavior

It seems that 'dapr list' should not be picking up dapr dashboard services.

## Actual Behavior

It's picking up dapr dashboard services and listing them out as blank processes.

## Steps to Reproduce the Problem

1) Start a dapr dashboard

`dapr dashboard -p 5555`

2) Open a new terminal window and type

`dapr list`
![Screen Shot 2021-06-08 at 1 02 57 PM](https://user-images.githubusercontent.com/83607984/121229375-d9684300-c85b-11eb-95e5-7636e970dd54.png)

PR #759, opened 2021-07-06
filter instances that don't start sidecar from dapr list

tanvigour:
# Description
This change filters the Dapr instances that don't start sidecar in Dapr list: https://github.com/dapr/cli/issues/736

## Issue reference
Please reference the issue this PR will close: #736


tanvigour:
@wcs1only , @artursouza : Please let me know your thoughts on this

pkg/standalone/list_test.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestListRunWhenNoDaprrd(t *testing.T) {
+	t.Run("test Cmd", func(t *testing.T) {
+		startDashboard := exec.Command("dapr", "dashboard", "-p", "5555")
```
artursouza, line 17: Please, move this to E2E test because is is spawning a process.

pkg/standalone/list.go
```diff
@@ -180,7 +180,10 @@ func List() ([]ListOutput, error) {
 				listRow.Command = utils.TruncateString(run.appCmd, 20)
 			}
 
-			list = append(list, listRow)
+			// filter all dapr instances that do not start sidecar
+			if listRow.HTTPPort != 0 && listRow.GRPCPort != 0 {
```
artursouza, line 184: Is there a better way to detect only the dashboard process? I am asking this because we see instance where the sidecar does not start correctly and we end up with "zombie" CLI processes that we later on kill, so listing those "zombie" CLI processes is useful.

artursouza (changes requested):

tests/e2e/standalone/list_test.go
```diff
@@ -0,0 +1,23 @@
...
+package standalone
+
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestListRunWhenNoDaprrd(t *testing.T) {
```
wcs1only, line 15: It might make sense to group this test here: https://github.com/dapr/cli/blob/58263ce/tests/e2e/standalone/standalone_test.go#L394

wcs1only (approved):
Optional comment about where the test belongs

yaron2:
@tanvigour tests are failing.

tests/e2e/standalone/list_test.go
tanvigour: Thanks Charlie, latest change has the test moved to standalone_test.go and I have deleted the old test

tanvigour:
> @tanvigour tests are failing.

Hi @yaron2 , just fixed the test, feel free to merge when you get a chance. Thanks!

yaron2 (approved):

Issue #220, opened 2019-12-11
`dapr list` does not include independent `daprd` instances


philliphoff:
## Expected Behavior

With 0.3.0 of the CLI and runtime:

1. Start `daprd` independently (e.g. `daprd --app-port 3000 ...`)
1. Start your application
1. Run `dapr list`

The `daprd` instance should be in the list returned by `dapr list`.

## Actual Behavior

The `daprd` instance is not in the list returned by `dapr list` (i.e. only instances started via `dapr run` are listed).

## Steps to Reproduce the Problem

See the expected behavior.

sk593:
Just wanted to see if there was any progress on this issue. 

The Dapr VSCode extension (which helps users integrate Dapr into their applications by providing debugging capabilities, interaction with the Dapr runtime, etc) starts applications by invoking an instance of `daprd` directly and these running applications also do not show up in `dapr list`. The Tye extension also starts applications with a direct `daprd` call which yields similar results when running a `dapr list` command.

orizohar:
@wcs1only @mukundansundar @yaron2 - addressing this issue will really help with some tooling issues that have been prioritized for the VS Code extension. Can this be given higher priority?

AaronCrawfis:
Issue is when parsing the arguments starting at https://github.com/dapr/cli/blob/0a37376658ef83e8a1b49178cf4e6f1c93eb1dfc/pkg/standalone/list.go#L87

If an argument doesn't exist the entire process is dropped from the list instead of it displaying the process without the argument

AaronCrawfis:
Also, once all parameters are set correctly, it fails when resolving cliPIDString: https://github.com/dapr/cli/blob/0a37376658ef83e8a1b49178cf4e6f1c93eb1dfc/pkg/standalone/list.go#L117

![image](https://user-images.githubusercontent.com/54363786/124179667-0200e880-da68-11eb-8e27-e985b7bcd319.png)

AaronCrawfis:
And now realizing we pull the cli process as the list to return, not the daprd process https://github.com/dapr/cli/blob/0a37376658ef83e8a1b49178cf4e6f1c93eb1dfc/pkg/standalone/list.go#L144

We need to update the logic to include daprd instances without CLI processes.

PR #763, opened 2021-07-14
Updated `dapr list` to include independent `daprd` instances

sk593:
# Description

This PR updates `dapr list` to include independently run `daprd` sidecars (in addition to `dapr` instances) in its output. 

## Issue reference

Resolves issue: #220 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


wcs1only (commented):
Thanks! Can you please add some E2E tests to validate the new behavior?

Let me point you in the right direction:
https://github.com/dapr/cli/blob/master/tests/e2e/standalone/standalone_test.go#L401

tests/e2e/standalone/standalone_test.go
```diff
@@ -430,6 +430,33 @@ func testList(t *testing.T) {
...
+
+		cmd := exec.Command(daprdPath, "--app-id", "dapr_e2e_list", "--dapr-http-port", "3555", "--dapr-grpc-port", "4555", "--app-port", "0")
+		cmd.Start()
+
+		output, err := spawn.Command(getDaprPath(), "list")
+		t.Log(output)
+		require.NoError(t, err, "dapr list failed with daprd instance")
+		listOutputCheck(t, output)
+
+		output, err = spawn.Command(getDaprPath(), "stop", "--app-id", "dapr_e2e_list")
```
wcs1only, line 454: So the stop command today actually talks to the CLI process that called `dapr run` . Since you are calling daprd directly, the stop command isn't actually shutting down your daprd instance. I believe why the tests are failing. There is a way to shutdown daprd cleanly and directly. Send an HTTP POST to http://localhost:3555/v1.0/shutdown to gracefully exit daprd. Alternatively, you can cmd.Kill()

sk593: Doesn't the stop command code call the `dapr list` function (i.e. `List()`) in order to find the running `dapr` process and then kills the process with a matching `app id`? So, if the `List()` logic changes to include independent `daprd` processes, the stop command should also work on those processes, right? I had tested that locally before committing the tests and `dapr stop` also shuts down independent `daprd` processes. I'll add the direct shutdown to make sure that it works, though

wcs1only: Specifically `dapr stop` is meant to be used in conjunction with `dapr run` for two reasons:
1. The dapr CLI process is the only one that knows the pid of the app that was spawned in addition to the daprd process. If we just kill daprd, there's the potential to orphan the app you were running with `dapr run`
2. On windows, there is no way to kill a proc that you don't own gracefully. So we signal the cli, which kills the processes it spawned.

We can change this behavior, but we have to make sure those cases are covered. And that it works on all platforms.

tests/e2e/standalone/standalone_test.go
```diff
@@ -430,6 +430,34 @@ func testList(t *testing.T) {
...
+
+		path := filepath.Join(homeDir, ".dapr")
+		binPath := filepath.Join(path, "bin")
+		daprdPath := filepath.Join(binPath, "daprd")
+
+		if runtime.GOOS == "windows" {
+			daprdPath += ".exe"
+		}
+
+		cmd := exec.Command(daprdPath, "--app-id", "dapr_e2e_list", "--dapr-http-port", "3555", "--dapr-grpc-port", "4555", "--app-port", "0")
```
mukundansundar, line 446: just to make the app name unique, use `daprd_e2e_list`

tests/e2e/standalone/standalone_test.go
mukundansundar: @sk593 Have the scenarios mentioned above been tested in Windows?

mukundansundar (commented):
@sk593 Added some comments

tests/e2e/standalone/standalone_test.go
sk593: Yes, the independent daprd instances are picked up in Windows

pkg/standalone/list.go
```diff
@@ -166,24 +118,17 @@ func List() ([]ListOutput, error) {
...
-			}
+			listRow.AppID = appID
+			listRow.HTTPPort = httpPort
+			listRow.GRPCPort = grpcPort
+			listRow.AppPort = appPort
+			listRow.MetricsEnabled = enableMetrics
+			listRow.Command = utils.TruncateString(appCmd, 20)
 
-			// filter only dashboard instance
-			if listRow.AppID != "" {
```
mukundansundar, line 184: @sk593 How is it that removing this check but still the filter dashboard instance from list is still working? What has changed there? 
https://github.com/dapr/cli/pull/763/files#diff-40b8b5acdd5260df8de205a924b446eb2483901d4e4c30c6a23932d3223e99e9R117

mukundansundar, line 184: @sk593 How is it that removing this check but still the filter dashboard instance from list is still working? What has changed there? @pkedy @wcs1only 
https://github.com/dapr/cli/pull/763/files#diff-40b8b5acdd5260df8de205a924b446eb2483901d4e4c30c6a23932d3223e99e9R117

pkg/standalone/list.go
```diff
@@ -166,24 +118,17 @@ func List() ([]ListOutput, error) {
 			listRow := ListOutput{
 				Created: createTime.Format("2006-01-02 15:04.05"),
 				Age:     age.GetAge(createTime),
-				PID:     proc.Pid(),
+				PID:     cliPID,
```
mukundansundar, line 121: The PID here is only `proc.Pid()` if there is an error parsing the line 119 https://github.com/dapr/cli/pull/763/files#diff-40b8b5acdd5260df8de205a924b446eb2483901d4e4c30c6a23932d3223e99e9R117

But do we not want this to always be the proc.Pid()? 
@artursouza @wcs1only @yaron2 thoughts?

pkg/standalone/list.go
sk593: The dashboard instance runs as a `dapr` process. The changes look for `daprd` sidecar processes and then attach the parent `dapr` application (if it exists). Since the dashboard doesn't have a `daprd` sidecar, it's not picked up

pkg/standalone/list.go
sk593: I think the logic is that the PID should be the PID of the `dapr` application and not the sidecar if a parent `dapr` process exists. If there is no parent `dapr` process running (i.e. you've run an independent sidecar), the cliPID throws an error since there is no parent process. In this case, the PID is set to proc.Pid() since only the sidecar is running

wcs1only: Hmm, we probably want to disambiguate what PID we're talking about. Giving the CLI pid when the CLI is being used and the daprd PID when it is not has the potential to confuse our users quite a bit. Maybe the solution is to have two columns in the table, one for CLI PID and one for daprd PID. We can leave the CLI pid column blank. Or the other option would be to always give the daprd pid.

wcs1only (approved):

wcs1only (changes requested):
Previous approval was an accident, we need to address the PID inconsistency before we merge this.

pkg/standalone/list.go
sk593: I talked to @artursouza when first implementing this, and he had said that using the sidecar PID if there was no parent CLI process should be okay. But if that's too ambiguous, I think specifying both PIDs could help with that

wcs1only: Put yourself in the shoes of the user, when they see:
```
  APP ID     HTTP PORT  GRPC PORT  APP PORT  COMMAND         AGE  CREATED              PID    
  pythonapp  61845      61846      0         python3 app.py  7s   2021-08-04 11:51.00  37011  
  nodeapp    3500       61800      3000      node app.js     13s  2021-08-04 11:50.54  36990  
```

What are they to think that PID means? It could actually mean three different things, because there are three processes running (is it the CLI, daprd, or my app's PID?). This is somewhat ambiguous to begin with, but it is even more ambiguous if the answer to that question changes depending on how daprd was launched.  

I'm personally leaning towards just changing that column name to "DAPRD PID" and always giving the daprd pid.

sk593: I originally said to give both the CLI and daprd PID because other dapr commands (mainly `dapr stop`) use the CLI PID. Compatibility with the stop command was also part of the reasoning behind setting the PID to be the daprd sidecar's PID if there was an independent one running. I think the Windows stop code specifically needs the CLI PID not the daprd PID so just using the daprd PID might not be the best approach

But if killing the daprd sidecar also kills the CLI, then maybe just showing the daprd PID makes most sense.

artursouza: `dapr stop` should also stop the app if it was started as part of `dapr run`. I am OK to remove ambiguity here. Maybe we do need to have 3 columns: CLI PID, Sidecar PID and App PID and they are not always present. Then, `dapr stop` must change its logic to know how to stop on each combination of these 3 PIDs being present or not.

sk593: I added a column for the sidecar PID and updated the stop commands to use the sidecar PID when there is no parent CLI process running. Not sure if we also want to have the App PID in there as well, but I feel like that might be unnecessary information. @wcs1only @artursouza

pkg/standalone/stop_windows.go
```diff
@@ -21,7 +21,12 @@ func Stop(appID string) error {
 
 	for _, a := range apps {
 		if a.AppID == appID {
-			eventName, _ := syscall.UTF16FromString(fmt.Sprintf("dapr_cli_%v", a.PID))
+			pid := fmt.Sprintf("dapr_cli_%v", a.CliPID)
+			if pid == "0" {
+				pid = fmt.Sprintf("dapr_cli_%v", a.DaprdPID)
```
wcs1only, line 26: So I'm pretty sure this won't work unless we modify daprd to listen for this event. We can do that, of course, but daprd doesn't support this on windows today.

wcs1only (dismissed):

pkg/standalone/stop_windows.go
mukundansundar: @sk593 Can you create an issue in dapr/dapr to add this event listen to daprd windows?

mukundansundar (approved):

pkg/standalone/stop_windows.go
sk593: Sure, just made one [here](https://github.com/dapr/dapr/issues/3524)

pkg/standalone/stop_windows.go
```diff
@@ -21,7 +21,12 @@ func Stop(appID string) error {
 
 	for _, a := range apps {
 		if a.AppID == appID {
-			eventName, _ := syscall.UTF16FromString(fmt.Sprintf("dapr_cli_%v", a.PID))
+			pid := fmt.Sprintf("dapr_cli_%v", a.CliPID)
+			if pid == "0" {
```
artursouza, line 25: This will not work because pid is a formatted string and not simply the integer value converted to string.

artursouza (changes requested):

pkg/standalone/stop_windows.go
sk593: Good catch, just fixed that.

artursouza (approved):

frozenskys:
Any update on this? anything I can do to help it get merged?

yaron2:
> Any update on this? anything I can do to help it get merged?

Please resolve conflicts and we'll merge, as the PR is approved.

/cc @artursouza

tillig:
It'd be awesome to get this merged and released, it would also fix [the issues with the VS Code extension for Dapr](https://github.com/microsoft/vscode-dapr/issues/56) where the running applications don't show up in the dashboard or the app list.

mukundansundar:
closed by #1000

PR #770, opened 2021-07-22
Correct env variables passed to child applications

anhldbk:
# Description

- CLI only passes 03 env variables to applications. See [run.go#L185-L187](https://github.com/dapr/cli/blob/master/pkg/standalone/run.go#L185-L187)
- There are more.  See [Docs](https://docs.dapr.io/reference/environment/)
- Python SDK has a different view on such variables. See [global_settings.py](https://github.com/dapr/python-sdk/blob/master/dapr/conf/global_settings.py). This leads to `not-working` [tutorial](https://docs.dapr.io/developing-applications/building-blocks/service-invocation/howto-invoke-discover-services/#step-1-choose-an-id-for-your-service)

# Checklist

* [x] Add tests for CLI (run command) to validate env variables
* [x] Modify code to pass the missing variables
* [x] Refactor code (run.go can be cleaner)
* [x] Update documentation: https://github.com/dapr/docs/pull/1835


anhldbk:
@yaron2 PTAL

If you don't see [the refactor](https://github.com/dapr/cli/pull/770/commits/13b2966bf6e01475488582ae1beb3fafe7eddd87) is necessary, I can revert it

anhldbk:
@yaron2 I see the failed tests regarding to arguments `app-port`. 

Question: if the port is not set, do we need to assign it with a free port? See line https://github.com/dapr/cli/blob/master/pkg/standalone/run.go#L103

pkg/standalone/run.go
```diff
@@ -25,127 +26,250 @@ const sentryDefaultAddress = "localhost:50001"
...
-	PlacementHostAddr  string
-	ComponentsPath     string
-	AppSSL             bool
-	MetricsPort        int
-	MaxRequestBodySize int
+	EnableProfiling    bool   `arg:"enable-profiling"`
+	ProfilePort        int    `env:"DAPR_PROFILE_PORT" arg:"profile-port"`
+	LogLevel           string `arg:"log-level"`
+	MaxConcurrency     int    `arg:"app-max-concurrency"`
+	PlacementHostAddr  string `env:"HOST_ADDRESS" arg:"placement-host-address"`
```
mukundansundar, line 40: @anhldbk See lines https://github.com/dapr/cli/blob/master/cmd/run.go#L62 and https://github.com/dapr/cli/blob/master/cmd/run.go#L81.

When the Env var is given as DAPR_PLACEMENT_HOST_ADDRESS, it is picked up as placement-host-address.


@yaron2 Do you think this change is needed?

anhldbk: @mukundansundar actually there's no such env DAPR_PLACEMENT_HOST_ADDRESS right? I'm gonna remove it

yaron2:
Tests are failing

anhldbk:
> Tests are failing

@yaron2 Pls answer my question above https://github.com/dapr/cli/pull/770#issuecomment-885343000

pkg/standalone/run.go
mukundansundar: The prefix is `DAPR` for the viper bindings ... the PLACEMENT_HOST_ADDRESS is something that is being used right now so should not be removed ... It is translated to `placement-host-address` by `viper`

anhldbk: @mukundansundar Correct me if I'm wrong: all available env variables are defined here: https://github.com/dapr/dapr/blob/master/pkg/config/env/env_variables.go

mukundansundar: @anhldbk There are certain env vars for Dapr CLI alone that will not be defined in that file ... Some are available here https://docs.dapr.io/reference/environment/ ... Some more arguments and annotations supported by daprd, dapr CLI or k8s are here https://docs.dapr.io/reference/arguments-annotations-overview/

anhldbk: @mukundansundar Thanks for your pointer. Now I see we have the inconsistency between our docs & code. I'll add a todo in my checklist to fix it soon.

To clarify it a little bit for field `PlacementHostAddr`, we'll have:

- env variable: `DAPR_PLACEMENT_HOST`
- argument: `--placement-address` (not as the currently used `placement-host-address`). See [your link](https://docs.dapr.io/reference/arguments-annotations-overview/)

Correct?

mukundansundar: @anhldbk Sorry for the late reply, if you see `dapr run --help` as well as the `daprd` arguments, the field name has been changed to `--placement-host-address` the change needs to be done in docs. 
The Environment variable as a consequence will be `DAPR_PLACEMENT_HOST_ADDRESS`

mukundansundar:
> @yaron2 I see the failed tests regarding to arguments `app-port`.
> 
> Question: if the port is not set, do we need to assign it with a free port? See line https://github.com/dapr/cli/blob/master/pkg/standalone/run.go#L103

@anhldbk For this scenario no we should not be setting it using Freeport since then daprd will be waiting for the port to be active to communicate with the application. This port is used for daprd to app communication and if that is not needed and only the app communicates with dapr, the port will not be specified. 

See the docs link for the `--app-port` field 
This parameter tells Dapr which port your application is listening on.

artursouza:
What scenario is this change enabling? Basically trying to understand why we need to set those via env variables.

anhldbk:
@artursouza I noted on the description above.

> Python SDK has a different view on such variables. See global_settings.py. This leads to not-working tutorial

The root cause is: dapr cli doesn't pass necessary environment variables to child processes.

So is it good to make a PR?

mukundansundar:
> @artursouza I noted on the description above.
> 
> > Python SDK has a different view on such variables. See global_settings.py. This leads to not-working tutorial
> 
> The root cause is: dapr cli doesn't pass necessary environment variables to child processes.
> 
> So is it good to make a PR?

@anhldbk The specific settings that are in [global_settings.py](https://github.com/dapr/python-sdk/blob/master/dapr/conf/global_settings.py) are specific to python SDK. They will be processed by the SDK and not by the CLI directly. 
@wcs1only Thoughts?

For the [tutorial](https://docs.dapr.io/developing-applications/building-blocks/service-invocation/howto-invoke-discover-services/#step-1-choose-an-id-for-your-service) linked, I am able to run it using dapr CLI and it is working, there is a slight change in the docs that is needed, it is currently having the curl commands as 
```bash
curl http://localhost:3500/v1.0/invoke/cart/method/add -X POST
```
While that is true in general cases, the dapr run command is as follows 
```bash
dapr run --app-id cart --app-port 5000 python app.py
```

Here the argument for run setting the HTTP PORT for Dapr is not given, hence a random port is selected example:
```bash
$ dapr run --app-id cart --app-port 5000 python app.py
ℹ️  Starting Dapr with id cart. HTTP Port: 55738. gRPC Port: 55739
```

The tutorial then says run 
```bash
curl http://localhost:3500/v1.0/invoke/cart/method/add -X POST
curl: (7) Failed to connect to localhost port 3500: Connection refused
```
which fails with the above failure.

But if the CLI is used to invoke the method as shown in the CLI tab in the tutorial, 
```
$ dapr invoke --app-id cart --method add
Added!
```
it works as expected.


if it is run with 
```bash
dapr run --app-id cart --app-port 5000 -H 3500 python app.py
ℹ️  Starting Dapr with id cart. HTTP Port: 3500. gRPC Port: 55798
```
Then the HTTP port is also specified and the curl command also works: 
```bash
$ curl http://localhost:3500/v1.0/invoke/cart/method/add -X POST
Added!
```

I think this is a change that is needed in the docs, specifying the `-H` or `--dapr-http-port` argument for `dapr run` command to fix the HTTP port as 3500 for the following curl command in the tutorial . 
@artursouza @wcs1only thoughts on this?

anhldbk:
@mukundansundar I don't think it's specific only to Python SDK. 

The problem here is: 

- There's an inconsistency in code & docs in Env variables
- The inconsistency leads to wrongly SDK implementations 

@yaron2 What do you think? Please comment so I know how to proceed.

yaron2:
> @mukundansundar I don't think it's specific only to Python SDK.
> 
> The problem here is:
> 
> * There's an inconsistency in code & docs in Env variables
> * The inconsistency leads to wrongly SDK implementations
> 
> @yaron2 What do you think? Please comment so I know how to proceed.

@anhldbk you raise a real issue which is that there are inconsistencies. I think as step 1 we need to document the inconsistencies and then fix them in documentation. If you're willing to help out with this since you seem to have a good knowledge of this, then that'd be great and highly appreciated.

After we have a list of inconsistencies we can decide how to tackle it. In general, there are SDK specific env vars that shouldn't be passed to the child app. So I think mapping these out are a must before we continue looking at this PR.

anhldbk:
@yaron2 

> If you're willing to help out with this since you seem to have a good knowledge of this, then that'd be great and highly appreciated.

Yes! Gonna work on `dapr/docs`. Please keep this PR ;)

msfussell:
@anhldbk - Can you explain in the PR, exactly what has now changed in the PR for the Env Vars? This is both in terms of the list of the env variables that you propose to add/change, coupled with a "why" scenario for each of them? I think everyone is seeing different perspectives here and it would be good to bring some clarity. Not all of them need to be provided from the CLI to the SDK, and as Mukundan points out, for the Python SDK example, it could be clearer on how this is run with the correct CLI parameters that current exist.

anhldbk:
@msfussell Hi Mark

> Can you explain in the PR, exactly what has now changed in the PR for the Env Vars? 

I updated the description a little bit. To summary:

- I made a PR to this documentation about environment: https://docs.dapr.io/reference/environment/
- This PR adds missing env variables including APP_ID, APP_PORT. 

> Not all of them need to be provided from the CLI to the SDK, and as Mukundan points out, for the Python SDK example, it could be clearer on how this is run with the correct CLI parameters that current exist.

Sorry. I just put a checklist here for myself. Gonna remove it. 
If we correctly pass such env variables, I'm gonna open another PR on our Python SDK.

anhldbk:
@msfussell Hi Mark

> Can you explain in the PR, exactly what has now changed in the PR for the Env Vars? 

I updated the description a little bit. To summary:
- I made a PR to this docs: https://docs.dapr.io/reference/environment/
- This PR adds missing env variables including APP_ID, APP_PORT. 


This is both in terms of the list of the env variables that you propose to add/change, coupled with a "why" scenario for each of them? I think everyone is seeing different perspectives here and it would be good to bring some clarity. Not all of them need to be provided from the CLI to the SDK, and as Mukundan points out, for the Python SDK example, it could be clearer on how this is run with the correct CLI parameters that current exist.

yaron2:
@anhldbk please resolve conflicts.

@msfussell please finish the review

msfussell (dismissed):
LGTM

anhldbk:
@yaron2 Conflicts are resolved.

@msfussell You made my day. Thank you.

anhldbk:
@yaron2 WDYT? The failed tests relate to https://github.com/dapr/cli/issues/794

yaron2:
@anhldbk I re-ran the tests, it's probably flaky.

However, there's a minor code linting issue:

```
Error: File is not `gofumpt`-ed (gofumpt)
Error: File is not `gofumpt`-ed (gofumpt)
Error: File is not `gofumpt`-ed (gofumpt)
```

If you go to the Files Changed tab you should be able to see which files need to be run with `gofumpt -w <file-name>.go`.

PR #771, opened 2021-07-22
Revert pinning of Go 1.16.6

berndverst:
# Description

Revert pinning of Go 1.16.6 because that is the default as of today.


pkedy (approved):
LGTM

mukundansundar (approved):

berndverst:
Given that the release is out now I will merge this back in at this time. This will be handy in case we need to do a hotfix release in the future.

PR #788, opened 2021-08-23
runtime: support unix domain socket

daixiang0:
# Description

Ref https://github.com/dapr/dapr/issues/2864

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


cmd/invoke.go
```diff
@@ -75,6 +86,7 @@ func init() {
 	InvokeCmd.Flags().StringVarP(&invokeVerb, "verb", "v", defaultHTTPVerb, "The HTTP verb to use")
 	InvokeCmd.Flags().StringVarP(&invokeDataFile, "data-file", "f", "", "A file containing the JSON serialized data (optional)")
 	InvokeCmd.Flags().BoolP("help", "h", false, "Print this help message")
+	InvokeCmd.Flags().BoolVar(&invokeUseSocket, "enable-domain-socket", false, "Invoke thought Unix domain socket, not supported on Windows")
```
artursouza, line 89: Set a short version too for this param. Maybe "u" and rename the long version to "unix-domain-socket"

artursouza: Also, the path might not always be "/tmp", since @yaron2 has plans to make it work in K8s - which requires a shared volume. These commands should assume "/tmp" is configurable here too.

daixiang0: For now, since `dapr/dapr` only support `/tmp`, may update it first I think,

In my proposal, standalone mode support is first step, and the next is to support in K8s.

@yaron2 could share more about this plan here?

daixiang0: @artursouza changed, please review again.

artursouza: The concern still remains here regarding this being a string and not a boolean param. The problem is that the CLI params behave like a public API, so we need to keep supporting it or deprecate later. To avoid rework here, we should make the path to be configurable in daprd and make this param a string instead of bool.

/cc @yaron2 @wcs1only

daixiang0: To sync with runtime, I can make it as string and use tmp path as default value. I would do it after you agree.

yaron2: Please do that @daixiang0

daixiang0: Please review @artursouza @yaron2

yaron2: Are the test failures related to the change?

daixiang0: Since go-sdk still does not support it, only unit tests added.

yaron2:
@daixiang0 please fix the failing tests.

@mukundansundar can you please review?

daixiang0:
doc pr: https://github.com/dapr/docs/pull/1814

mukundansundar:
@daixiang0 Can you take a look at the Self Hosted E2E that is failing?

daixiang0:
@mukundansundar @yaron2 all tests pass, please review.

cmd/invoke.go
```diff
@@ -54,7 +59,14 @@ dapr invoke --app-id target --method sample --verb GET
 			bytePayload = []byte(invokeData)
 		}
 		client := standalone.NewClient()
-		response, err := client.Invoke(invokeAppID, invokeAppMethod, bytePayload, invokeVerb)
+
+		// TODO(@daixiang0): add Windows support
+		if runtime.GOOS == "windows" && invokeSocket != "" {
+			invokeSocket = ""
+			print.WarningStatusEvent(os.Stdout, "unix-domain-socket option still does not support Windows!")
```
artursouza, line 66: Just fail instead. No need to fallback.

cmd/publish.go
```diff
@@ -49,7 +54,12 @@ dapr publish --publish-app-id myapp --pubsub target --topic sample --data '{"key
 		}
 
 		client := standalone.NewClient()
-		err = client.Publish(publishAppID, pubsubName, publishTopic, bytePayload)
+		// TODO(@daixiang0): add Windows support
+		if runtime.GOOS == "windows" && publishSocket != "" {
+			publishSocket = ""
+			print.WarningStatusEvent(os.Stdout, "unix-domain-socket option still does not support Windows!")
```
artursouza, line 60: Just fail.

cmd/run.go
```diff
@@ -66,6 +68,18 @@ var RunCmd = &cobra.Command{
 			fmt.Println(print.WhiteBold("WARNING: no application command found."))
 		}
 
+		if unixDomainSocket != "" {
+			// TODO(@daixiang0): add Windows support
+			if runtime.GOOS == "windows" {
```
artursouza, line 73: Just fail.

cmd/invoke.go
daixiang0: All changed.

yaron2:
@daixiang0 please update the README with an example for how to use the new parameter.

PR #806, opened 2021-10-07
Allow specifying of runtime-version via Env Var for `dapr init`

berndverst:
# Description

Allows optionally specifying of the runtime version with an environment variable.

## Justification:

When using `dapr init` / `dapr init -k` in workflows and tests it is handy to specify the runtime version via an environment variables.

This is particularly true where Dapr's Mechanical Markdown is used and the runtime version itself should not appear in the code snippets.
E.g. Quickstart validation, sample validation for Python and Go SDKs

Full example

```bash
export DAPR_RUNTIME_VERSION=1.4.3-rc.1
dapr init
```

Output
```
ℹ️  Dapr runtime version: 1.4.3-rc.1
ℹ️  Installing runtime version 1.4.3-rc.1
↓  Downloading binaries and setting up components...
Dapr runtime installed to /Users/bverst/.dapr/bin, you may run the following to add it to your path if you want to run daprd directly:
    export PATH=$PATH:/Users/bverst/.dapr/bin
✅  Downloading binaries and setting up components...
✅  Downloaded binaries and completed components set up.
ℹ️  daprd binary has been installed to /Users/bverst/.dapr/bin.
ℹ️  dapr_placement container is running.
ℹ️  dapr_redis container is running.
ℹ️  dapr_zipkin container is running.
ℹ️  Use `docker ps` to check running containers.
✅  Success! Dapr is up and running. To get started, go here: https://aka.ms/dapr-getting-started
```

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -59,6 +59,14 @@ dapr init -s
 	Run: func(cmd *cobra.Command, args []string) {
 		print.PendingStatusEvent(os.Stdout, "Making the jump to hyperspace...")
 
+		if runtimeVersion == "latest" {
+			runtimeVersionEnv := viper.GetString("runtime_version")
+			if runtimeVersionEnv != "" {
+				runtimeVersion = runtimeVersionEnv
+			}
+		}
+		print.InfoStatusEvent(os.Stdout, "Dapr runtime version: %s", runtimeVersion)
+
 		if kubernetesMode {
 			print.InfoStatusEvent(os.Stdout, "Note: To install Dapr using Helm, see here: https://docs.dapr.io/getting-started/install-dapr-kubernetes/#install-with-helm-advanced\n")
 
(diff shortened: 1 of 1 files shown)


cmd/init.go
```diff
@@ -59,6 +59,14 @@ dapr init -s
 	Run: func(cmd *cobra.Command, args []string) {
 		print.PendingStatusEvent(os.Stdout, "Making the jump to hyperspace...")
 
+		if runtimeVersion == "latest" {
```
artursouza, line 62: There can be a different behavior where "latest" is the actual value of the `runtime-version` param and then the logic will override it. Instead, can you change the default to be the environment variable or "latest" if not set, and then, use as the default. So, I would move this change to `init()` in the following way:

```
defaultRuntimeVersion := "latest"
runtimeVersionEnv := viper.GetString("runtime_version")
if runtimeVersionEnv != "" {
}
```

Then, use it in the param.
`InitCmd.Flags().StringVarP(&runtimeVersion, "runtime-version", "", defaultRuntimeVersion, "The version of the Dapr runtime to install, for example: 1.0.0")`

berndverst: That's a very subtle difference but I see what you are saying.

berndverst: Addressed the feedback. Thanks!

artursouza (approved):

PR #826, opened 2021-11-10
Adds native binary support for Darwin Arm64 (e.g. M1 Mac)

berndverst:
# Description

Adds native binary support for Darwin Arm64 (e.g. M1 Mac)

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -51,12 +51,15 @@ verifySupported() {
     done
 
     if [ "$current_osarch" == "darwin-arm64" ]; then
-        echo "The darwin_arm64 arch has no native binary, however you can use the amd64 version so long as you have rosetta installed"
-        echo "Use 'softwareupdate --install-rosetta' to install rosetta if you don't already have it"
-        ARCH="amd64"
-        return
+        if isReleaseAvailable; then
+            return
+        else
+            echo "The darwin_arm64 arch has no native binary for this version of Dapr, however you can use the amd64 version so long as you have rosetta installed"
+            echo "Use 'softwareupdate --install-rosetta' to install rosetta if you don't already have it"
+            ARCH="amd64"
+            return
+        fi
     fi
-
 
     echo "No prebuilt binary for ${current_osarch}"
     exit 1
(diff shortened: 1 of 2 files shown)


berndverst:
This install script change works:

```
./install/install.sh
Your system is darwin_arm64
Installing Dapr CLI...

Getting the latest Dapr CLI...
Installing v1.4.0 Dapr CLI...
Downloading https://github.com/dapr/cli/releases/download/v1.4.0/dapr_darwin_arm64.tar.gz
```

(That of course fails because this artifact doesn't exist, but it will start working with a later 1.5.0 release candidate and the 1.5.0 release)

pkedy (dismissed):

install/install.sh
```diff
@@ -50,14 +50,6 @@ verifySupported() {
         fi
     done
 
-    if [ "$current_osarch" == "darwin-arm64" ]; then
```
wcs1only, line 53: So we need a bit more robust fallback than this. The install.sh script is unversioned and live, which means this change will break M1 macs until such time as the release is published. I think what we need to do is "if darwin-arm64 && darwin/arm64 exists and is public then download it else amd64"

berndverst: Makes sense -- keep in mind this script is local to the 1.5 branch right now, but I agree we should have this.

wcs1only: > Makes sense -- keep in mind this script is local to the 1.5 branch right now, but I agree we should have this.

True, but this also means it won't work until we merge release branch back to master, and thus we'll have no way to test it works before we go live.

berndverst:
After this change behavior is as before where Darwin Arm64 binary does not exist:
```
./install.sh
The darwin_arm64 arch has no native binary for this version of Dapr, however you can use the amd64 version so long as you have rosetta installed
Use 'softwareupdate --install-rosetta' to install rosetta if you don't already have it
Installing Dapr CLI...

Getting the latest Dapr CLI...
Installing v1.4.0 Dapr CLI...
Downloading https://github.com/dapr/cli/releases/download/v1.4.0/dapr_darwin_amd64.tar.gz ...
```

wcs1only (approved):
LGTM

PR #827, opened 2021-11-10
Install script update for Darwin ARM 64 (M1 Mac etc) support

berndverst:
- This cherry-picks commit 78db31a6b4946262854b4f8399b77f7f0ed19260 from Release-1.5 into master.
- Fixes an issue in the install script.

The install scripts are only run from the Master branch and therefore this change needs to be cherry picked ahead of the release.

* Add Darwin ARM64 binaries

* Update install script for Darwin Arm64


wcs1only (approved):
LGTM

Issue #832, opened 2021-11-12
 Have E2E tests automatically pull the latest RC/stable release


mukundansundar:
## Describe the proposal
 Have E2E tests automatically pull the latest RC/stable release

In the upgrade tests, add one test that automatically uses the previous stable version and the latest stable or RC version.

## Release Note

RELEASE NOTE:

mukundansundar:
related to #773

mukundansundar:
related to #773

mukundansundar:
Take a look at this https://github.com/dapr/cli/pull/806 and have a similar env var for dashboard version.

https://github.com/dapr/quickstarts/blob/master/.github/workflows/validate.yaml#L90
Have something similar added to the E2E test workflows .

PR #847, opened 2021-11-30
Added code to pull dashboard version automatically for E2E tests

amulyavarote:
# Description

Added code to get the latest Dapr dashboard version automatically for E2E tests

## Issue reference

Please reference the issue this PR will close: #832 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/kind_e2e.yaml
+++ b/.github/workflows/kind_e2e.yaml
@@ -46,9 +46,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GOVER: 1.16
-      #Added default values to the versions
-      DAPR_RUNTIME_VERSION: 10.5.1
-      DAPR_DASHBOARD_VERSION: 10.9.0
+      DAPR_RUNTIME_VERSION: 1.5.1
+      DAPR_DASHBOARD_VERSION: 0.9.0
     strategy:
       fail-fast: false # Keep running if one leg fails.
       matrix:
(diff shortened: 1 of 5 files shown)


mukundansundar:
@amulyavarote Please modify the E2E tests also the use a specified dashboard version.

mukundansundar:
@amulyavarote Take a look at https://github.com/dapr/quickstarts/blob/master/.github/workflows/validate.yaml#L90 and add similar functionality in the Yaml files for the E2E in CLI to pull the latest RC/Stable release of Runtime and dashboard, and make use of the passed in versions from Yaml files in the Tests ....

mukundansundar:
@amulyavarote Still the version 1.5.0 is what is being used please remove the overrides that are there for versions in the standalone_test.go and Kubernetes_test.go in tests/e2e

.github/workflows/kind_e2e.yaml
```diff
@@ -121,7 +121,18 @@ jobs:
     - name: Free up some diskspace
       run: |
         docker image prune -a -f
-
+    - name: Determine latest Dapr Runtime version including Pre-releases
```
artursouza, line 124: @mukundansundar Would this make builds not reproducible for release branches going forward? I remember pinning the runtime version in our Java SDK IT tests to make sure old releases continue to build the same way.

cmd/init.go
```diff
@@ -100,12 +100,18 @@ func init() {
 	if runtimeVersionEnv != "" {
 		defaultRuntimeVersion = runtimeVersionEnv
 	}
+	defaultDashboardVersion := "latest"
```
artursouza, line 103: LGTM

.github/workflows/kind_e2e.yaml
artursouza: I would keep the other changes but pin the runtime and dashboard versions here instead.

artursouza: Another option is to combine both where if in master branch use the latest but if from a release branch, use a pinned version.

mukundansundar: I agree @artursouza. Pinning the versions in the release branch and using the latest versions in the master branch and associated PRs makes perfect sense. 
@amulyavarote can you make changes such that PRs targeting release branch and the release branch itself have pinned versions whereas the master branch and PRs targeting master branch have the latest versions?

PR #852, opened 2021-12-07
fix get latest version

mukundansundar:
# Description

## Issue reference

Please reference the issue this PR will close: #853

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pkg/version/version_test.go
+++ b/pkg/version/version_test.go
@@ -62,6 +62,74 @@ func TestGetVersionsGithub(t *testing.T) {
 			"1.2.2",
 		},
 		{
+			"Only latest version is got",
+			"/latest",
+			`[
+  {
+    "url": "https://api.github.com/repos/dapr/dapr/releases/44766923",
+    "html_url": "https://github.com/dapr/dapr/releases/tag/v1.4.4",
+    "id": 44766926,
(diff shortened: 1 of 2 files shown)


pkg/version/version_test.go
```diff
@@ -48,6 +48,74 @@ func TestGetVersionsGithub(t *testing.T) {
 			"",
 			"1.2.2",
 		},
+		{
```
wcs1only, line 51: Can we add this same test for the helm chart version as well? We added a fallback to reading the helm chart when the github API is unavailable, and we should ensure it has the same behavior.

mukundansundar: Based on the commits 
https://github.com/dapr/helm-charts/commit/7f6459e5376235424a7ff5af7e1e6c85162f4d44
and 
https://github.com/dapr/helm-charts/commit/c7a5ac8221745f03d213025d1d6a2a5646fef50d

The index entries are written in the correct order and always the latest version is the first at the top. Any previous hot fixes are inserted into the entries list. 

cc @yaron2 @artursouza

yaron2 (approved):

shubham1172 (approved):

PR #861, opened 2021-12-22
Add dapr run `--app-grace-exit` flag to make dapr stop app gracefully.

li1234yun:
# Description

When I use FastAPI with dapr like dapr run -a app -H 3500 -G 50001 -p 8000 -P http --log-level info -- uvicorn --host 0.0.0.0 --port 8000 app.main:app, I can not get the FastAPI shutdown singal when dapr exited. But When I use kill pid to stop FastAPI app will get the shutdown singal, use kill -9 pid to stop FastAPI app also will not catch the single. So I think the dapr use kill -9 to stop app, maybe can change to kill -15 to stop app gracefully.

## Issue reference

close: dapr/dapr#4045

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


tanvigour (approved):
LGTM

PR #872, opened 2022-01-14
Upgrade dapr package to 1.6.0-rc.1

yaron2:


artursouza (changes requested):
Search for 1.5.1. There are other files to change.

yaron2:
> Search for 1.5.1. There are other files to change.

These modify the runtime version that's run in tests, but changing it to 1.6.0 will result in testing a version that's not released yet.

mukundansundar:
@yaron2 This can be closed I think since once release is done, release branch will be merged back into master, it already contains this change.

PR #885, opened 2022-01-26
Installer: Remove old Dapr CLI to avoid file corruption

berndverst:
Signed-off-by: Bernd Verst <4535280+berndverst@users.noreply.github.com>

# Description

Currently the installer corrupts binaries when the Dapr version is updated. This addresses the problem.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -176,6 +176,9 @@ installFile() {
         exit 1
     fi
 
+    if [ -f "$DAPR_CLI_FILE" ]; then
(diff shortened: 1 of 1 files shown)


berndverst:
I validated this works on M1 Mac (Arm64) and AMD64 Linux. In both cases I first installed CLI 1.5.1 manually and then used this script to install the latest 1.6.0 CLI.

yaron2 (approved):

PR #890, opened 2022-02-08
Exposed config path in standalone mode.

l0ll098:
# Description

Changes included in this PR expose a new propery called `ConfigPath` that, in case of standalone mode, indicates which configuration file has been used. Value of this property would be an absolute path to that config file, that can be then read by the dashboard server later on.
Similarly to the `MetricsEnabled` property, this would be only consumed by the Dashboard, so it won't appear in the `dapr list` command output.
With this new piece of information, dashboard could accordingly populate the "configurations" section.

For example, in the following screenshot you can see a local installation of the Dapr Dashboard (and Dapr CLI) with some changes to handle this case. In particular, I launched a single NodeJS app with the following command: `dapr --config dapr.config.yaml --app-id api --app-port 3000 run node dist/server.js`, where `dapr.config.yaml` is a Dapr Configuration file stored in my project directory.

dashboard

As you can see in the screenshot above, we have two configs:
 - `daprConfigDalleApi` is the value specified in the `metadata.name` property in the aforementioned `dapr.config.yaml` file
 - `daprConfig` is the usual Dapr Config file created as a result of the `dapr init` command

Note that with the current live Dapr version, you would only see the second one (that is `daprConfig`).

With these changes, from the point of view of an end user, configurations are handled in the same exact way, both in Standalone mode and in a Kubernetes cluster.
If this gets merged, I will proceed to open a new PR in the Dashboard project to add the required dashboard-side logic to handle this case, since I've already a patch ready to go (that depends on this PR).

## Issue reference

This PR exposes required information to solve this issue: dapr/dashboard#167

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


mukundansundar:
@yaron2 Please review this PR ...

l0ll098:
@yaron2 Any news on this one?

l0ll098:
Monthly ping :)

l0ll098:
Monthly ping :)

l0ll098:
Hey @yaron2, now that I've merged master and resolved some conflicts, can you take a look at this PR please?

mukundansundar (commented):
@l0ll098 Can you add E2E tests to verify the output in the different formats i.e. no output in `dapr list` and config available in JSON and YAML formats.

mukundansundar:
@l0ll098 Please fix the conflicts in the PR and add E2E tests and doc changes both in README and also https://docs.dapr.io/reference/cli/dapr-list/ in the dapr/docs repo

l0ll098:
Hi @mukundansundar, I've added some e2e tests as requested and resolved those merge conflicts. Now it should be ready to be reviewed.
As per the documentation changes: it appears that currently there is no documentation about each property returned by the `dapr list -o json` (or yaml for that matter). Since this pr does not change the general behaviour of the list command itself, but rather exposes a new piece of information if and only if the output target is either json or yaml, I'd say there is no need to change those documentation pages. Those changes would make sense once (and if) the documentation gets expanded to explain each field of the output of the `dapr list` command

l0ll098:
Montly ping

yaron2:
> Montly ping

Apologies, this has gone under my radar. Please resolve the conflicts while I review.

Issue #893, opened 2022-02-11
Dapr CLI should warn about upcoming TLS certificate expiration


berndverst:
Add warning to CLI for upcoming TLS certificate expiration

This is similar to https://github.com/dapr/dapr/issues/3940 which is for the sidecar / runtime only.

I propose that every `dapr` command targeting Kubernetes, e.g. `dapr upgrade -k`, or `dapr status -k` internally perform the `dapr mtls expiry` command and print a warning message if the TLS expiration is coming up in a set period of days, perhaps 7 days.

pravinpushkar:
Certificate expiry check can be added for the following commands - 
```
dapr components -k
dapr configurations -k
dapr dashboard -k
dapr list -k
dapr logs -k
dapr mtls -k
dapr mtls export
dapr status -k 
dapr upgrade -k
```

mukundansundar:
LGTM

msfussell:
This issue needs a  mention in the documentation. Also 7 days is too short, this should be 30 days or less IMO. Usually this takes time to plan for production workloads. If this is 7 days this needs changing.

pravinpushkar:
> @pravinpushkar @mukundansundar This issue needs a mention in the documentation. Also 7 days is too short, this should be 30 days or less IMO. Usually this takes time to plan for production workloads. If this is 7 days this needs changing.

@msfussell  It is 30 days in current code implementation. So, I think no change needed on this front.
I will mention this in documentation also.

PR #899, opened 2022-02-18
Include certificate expiry check in all kubernetes commands

pravinpushkar:

# Description

Show warning for certificate expiry on all kubernetes related commands

## Issue reference

Please reference the issue this PR will close: #893 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/kubernetes/mtls.go
+++ b/pkg/kubernetes/mtls.go
@@ -112,7 +112,7 @@ func ExportTrustChain(outputDir string) error {
 // Check and warn if cert expiry is less than `warningDaysForCertExpiry` days.
 func CheckForCertExpiry() {
 	expiry, err := Expiry()
-	// The intent is to warn for certificate expiry only when it can be fetched.
+	// The intent is to warn for certificate expiry, only when it can be fetched.
 	// Do not show any kind of errors with normal command flow.
 	if err != nil {
 		return
@@ -123,15 +123,15 @@ func CheckForCertExpiry() {
 		warningMessage := ""
 		switch {
 		case daysRemaining == 0:
-			warningMessage = "Root certificate of your kubernetes cluster expires today"
+			warningMessage = "Dapr root certificate of your Kubernetes cluster expires today."
 		case daysRemaining < 0:
-			warningMessage = "Root certificate your kubernetes cluster already expired"
+			warningMessage = "Dapr root certificate of your Kubernetes cluster has expired."
 		default:
-			warningMessage = fmt.Sprintf("Root certificate your kubernetes cluster expires in %v days", daysRemaining)
+			warningMessage = fmt.Sprintf("Dapr root certificate of your Kubernetes cluster expires in %v days.", daysRemaining)
 		}
-		helpMessage := "Kindly renew to avoid any service interuptions."
+		helpMessage := "Please see docs.dapr.io for certificate renewal instructions to avoid service interruptions."
 		print.WarningStatusEvent(os.Stdout,
-			fmt.Sprintf("%s. Expiry date: %s. \n %s", warningMessage, expiry.Format(time.RFC1123), helpMessage))
+			fmt.Sprintf("%s Expiry date: %s. \n %s", warningMessage, expiry.Format(time.RFC1123), helpMessage))
 	}
 }
 
--- a/pkg/kubernetes/mtls.go
+++ b/pkg/kubernetes/mtls.go
@@ -113,25 +113,25 @@ func ExportTrustChain(outputDir string) error {
 // Check and warn if cert expiry is less than `warningDaysForCertExpiry` days.
 func CheckForCertExpiry() {
 	expiry, err := Expiry()
-	if err != nil {
-		print.FailureStatusEvent(os.Stderr, fmt.Sprintf("error getting root cert expiry: %s", err))
-		return
-	}
-	daysRemaining := int(expiry.Sub(time.Now().UTC()).Hours() / 24)
-	if daysRemaining < warningDaysForCertExpiry {
-		warningMessage := ""
-		if daysRemaining == 0 {
-			warningMessage = "Root certificate of your kubernetes cluster expires today"
-		} else if daysRemaining < 0 {
-			warningMessage = "Root certificate your kubernetes cluster already expired"
-		} else {
-			warningMessage = fmt.Sprintf("Root certificate your kubernetes cluster expires in %v days", daysRemaining)
+	// The intent is to warn for certificate expiry only when it can be fetched.
+	// Do not show any kind of errors with normal command flow.
+	if err == nil {
+		daysRemaining := int(expiry.Sub(time.Now().UTC()).Hours() / 24)
+		if daysRemaining < warningDaysForCertExpiry {
+			warningMessage := ""
+			if daysRemaining == 0 {
+				warningMessage = "Root certificate of your kubernetes cluster expires today"
+			} else if daysRemaining < 0 {
+				warningMessage = "Root certificate your kubernetes cluster already expired"
+			} else {
+				warningMessage = fmt.Sprintf("Root certificate your kubernetes cluster expires in %v days", daysRemaining)
+			}
+			color.Set(color.FgHiYellow)
+			helpMessage := "Kindly renew to avoid any service interuptions."
+			message := fmt.Sprintf("%s. Expiry date: %s. \n %s", warningMessage, expiry.Format(time.RFC1123), helpMessage)
+			print.WarningStatusEvent(os.Stdout, message)
+			color.Unset()
 		}
-		color.Set(color.FgHiYellow)
-		helpMessage := "Kindly renew to avoid any service interuptions."
-		message := fmt.Sprintf("%s. Expiry date: %s. \n %s", warningMessage, expiry.Format(time.RFC1123), helpMessage)
-		print.WarningStatusEvent(os.Stdout, message)
-		color.Unset()
 	}
 }
 
--- a/pkg/kubernetes/mtls.go
+++ b/pkg/kubernetes/mtls.go
@@ -34,8 +34,9 @@ import (
 )
 
 const (
-	systemConfigName      = "daprsystem"
-	trustBundleSecretName = "dapr-trust-bundle" // nolint:gosec
+	systemConfigName         = "daprsystem"
+	trustBundleSecretName    = "dapr-trust-bundle" // nolint:gosec
+	warningDaysForCertExpiry = 30                  // in days
 )
 
 func IsMTLSEnabled() (bool, error) {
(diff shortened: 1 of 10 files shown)


pkg/kubernetes/mtls.go
```diff
@@ -107,6 +109,23 @@ func ExportTrustChain(outputDir string) error {
 	return nil
 }
 
+// Check and warn if cert expiry is less than 90 days.
+func WarnForCertExpiry() {
```
mukundansundar, line 113:
```suggestion
func CheckForCertExpiry() {
```

mukundansundar (commented):
@pravinpushkar Can you take a look at the k8s E2E test. All of them seem to be failing.

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +109,23 @@ func ExportTrustChain(outputDir string) error {
...
+	expiry, err := Expiry()
+	if err != nil {
+		print.FailureStatusEvent(os.Stderr, fmt.Sprintf("error getting root cert expiry: %s", err))
+		return
+	}
+	daysRemaining := int(expiry.Sub(time.Now().UTC()).Hours() / 24)
+	if daysRemaining < 30 {
```
shubham1172, line 120: nit: Should we keep 30 as a constant above?

mukundansundar: no magic numbers please ... Put it as constants above in code ...

pravinpushkar:
> @pravinpushkar Can you take a look at the k8s E2E test. All of them seem to be failing.

Running now.

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +110,31 @@ func ExportTrustChain(outputDir string) error {
 	return nil
 }
 
+// Check and warn if cert expiry is less than `warningDaysForCertExpiry` days.
+func CheckForCertExpiry() {
+	expiry, err := Expiry()
+	// The intent is to warn for certificate expiry only when it can be fetched.
+	// Do not show any kind of errors with normal command flow.
+	if err == nil {
```
shubham1172, line 118: nit: Usually it's easier to read go code that goes like this:
```go
if err != nil {
    return
}
// other logic
```

shubham1172: Reference https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +110,31 @@ func ExportTrustChain(outputDir string) error {
...
+		daysRemaining := int(expiry.Sub(time.Now().UTC()).Hours() / 24)
+		if daysRemaining < warningDaysForCertExpiry {
+			warningMessage := ""
+			if daysRemaining == 0 {
```
shubham1172, line 122: Same nit for line 120 as above.
For if/else checks from line 122, we can try using a switch case here for readability.

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +110,31 @@ func ExportTrustChain(outputDir string) error {
...
+				warningMessage = "Root certificate of your kubernetes cluster expires today"
+			} else if daysRemaining < 0 {
+				warningMessage = "Root certificate your kubernetes cluster already expired"
+			} else {
+				warningMessage = fmt.Sprintf("Root certificate your kubernetes cluster expires in %v days", daysRemaining)
+			}
+			color.Set(color.FgHiYellow)
+			helpMessage := "Kindly renew to avoid any service interuptions."
+			message := fmt.Sprintf("%s. Expiry date: %s. \n %s", warningMessage, expiry.Format(time.RFC1123), helpMessage)
+			print.WarningStatusEvent(os.Stdout, message)
```
shubham1172, line 132: There are colors defined in the `print` pkg which look unused. As a caller of `WarningStatusEvent`, I would rely on the function to set the colors instead. It can use different colors based on the OS, environment, accessibility guidelines, etc., but the caller should remain agnostic. What are your thoughts, @pravinpushkar, @mukundansundar?

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +110,31 @@ func ExportTrustChain(outputDir string) error {
...
+			if daysRemaining == 0 {
+				warningMessage = "Root certificate of your kubernetes cluster expires today"
+			} else if daysRemaining < 0 {
+				warningMessage = "Root certificate your kubernetes cluster already expired"
+			} else {
+				warningMessage = fmt.Sprintf("Root certificate your kubernetes cluster expires in %v days", daysRemaining)
+			}
+			color.Set(color.FgHiYellow)
+			helpMessage := "Kindly renew to avoid any service interuptions."
+			message := fmt.Sprintf("%s. Expiry date: %s. \n %s", warningMessage, expiry.Format(time.RFC1123), helpMessage)
```
shubham1172, line 131: nit: `WarningStatusEvent` takes variadic params and uses `fmt.Sprintf` internally. We can also do something like
```go
print.WarningStatusEvent(os.Stdout, "%s. Expiry date: %s.\n Kindly renew to avoid any service interruptions.", warningMessage, expiry.Format(time.RFC1123))
```

pkg/kubernetes/mtls.go
mukundansundar: the caller should remain agnostic of the colour ... Ideally only the behaviour that is wanted ... The colour will be set internally as defined by the function and needs to be consistent across all such calls.

shubham1172: Thanks for sharing your thoughts. For this PR maybe we can omit the color part then, and track the behavior of `print` pkg separately? I see color being used in only one more place that might need refactoring.

https://github.com/dapr/cli/blob/f89be62068c39ffd3492ab24025e3c0624a56c61/pkg/standalone/standalone.go#L788

pravinpushkar: I am also ok to remove the color. Just a thought,  normally people use blackish/whitish background terminal and in both these theme yellow(it is not actually yellow, it is kind of orange) would fit nicely at the same time we do want to highlight this because of the seriousness.

pkg/kubernetes/mtls.go
pravinpushkar: I have mixed feelings here. Normally switch cases are good if we compare with constants like.  I might be wrong though. Thoughts ?
```
n=5
switch n {
case 1:
case 2:
default:
}
```

pkg/kubernetes/mtls.go
pravinpushkar: It will be better to have diff PR. I will remove the color part.

pkg/kubernetes/mtls.go
shubham1172: The only benefit IMO is that this looks more idiomatic. I can see longer lines in our code-base, so might just want to keep it that way.  Or, we can introduce line breaks in the same statement (and where to put a line break can be subjective, for example, https://github.com/mvdan/gofumpt/issues/2). I would suggest this (the two function params are `io.Writer` and `...args`):
```go
print.WarningStatusEvent(os.Stdout, 
		"%s. Expiry date: %s.\n Kindly renew to avoid any service interruptions.", warningMessage, expiry.Format(time.RFC1123))
```

pkg/kubernetes/mtls.go
shubham1172: Sure @pravinpushkar, it's only a nit. Go does support conditional switch case though:
```go
switch {
case x < 0:
        // do something
case x > 0:
        // do something
case x == 0:
        // do something
}
```

mukundansundar (commented):
LGTM. The only concern that I have is, is it possible to have a temporary cert created following steps [here](https://docs.dapr.io/operations/security/mtls/#bringing-your-own-certificates-1) and have that verified as part of E2E ?
@berndverst Can you post your comments on this?

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +109,32 @@ func ExportTrustChain(outputDir string) error {
...
+	if err != nil {
+		return
+	}
+	daysRemaining := int(expiry.Sub(time.Now().UTC()).Hours() / 24)
+
+	if daysRemaining < warningDaysForCertExpiry {
+		warningMessage := ""
+		switch {
+		case daysRemaining == 0:
+			warningMessage = "Root certificate of your kubernetes cluster expires today"
```
berndverst, line 126:
```suggestion
			warningMessage = "Dapr root certificate of your Kubernetes cluster expires today."
```

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +109,32 @@ func ExportTrustChain(outputDir string) error {
...
+		case daysRemaining < 0:
+			warningMessage = "Root certificate your kubernetes cluster already expired"
```
berndverst, line 128:
```suggestion
			warningMessage = "Dapr root certificate of your Kubernetes cluster has expired."
```

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +109,32 @@ func ExportTrustChain(outputDir string) error {
...
+		default:
+			warningMessage = fmt.Sprintf("Root certificate your kubernetes cluster expires in %v days", daysRemaining)
```
berndverst, line 130:
```suggestion
			warningMessage = fmt.Sprintf("Dapr root certificate of your Kubernetes cluster expires in %v days", daysRemaining)
```

pkg/kubernetes/mtls.go
```diff
@@ -107,6 +109,32 @@ func ExportTrustChain(outputDir string) error {
...
+		}
+		helpMessage := "Kindly renew to avoid any service interuptions."
```
berndverst, line 132:
```suggestion
		helpMessage := "Please see docs.dapr.io for certificate renewal instructions to avoid service interruptions."
```

berndverst (approved):
Looks good to me assuming the various strings are updated.

I'm not an approve or maintainer of the CLI so @artursouza or @mukundansundar will need to review and merge after you make the update. You can just accept the changes I suggested (but that may require you to manually fix the DCO check).

berndverst:
> LGTM. The only concern that I have is, is it possible to have a temporary cert created following steps [here](https://docs.dapr.io/operations/security/mtls/#bringing-your-own-certificates-1) and have that verified as part of E2E ?
> @berndverst Can you post your comments on this?

@mukundansundar I don't have concerns here. The existing `dapr mtls expiry` command works with certificates generated by Dapr and certificates manually generated and provided per our docs. This PR here uses the same logic for getting the expiry.

We could create an E2E test for manually generating certificates, deploying them to Kubernetes and verifying with this command -- in fact that's what I did during the last community call (but manually so). I don't feel strongly that we need this. The `mtls expiry` command is well proven at this point.
I'll leave it up to you as the CLI maintainer.

This PR looks good to me assuming we change the various returned strings - they are a bit clumsy right now.

mukundansundar (approved):

Issue #909, opened 2022-03-04
Invoke invalid method does not fail


shubham1172:
## Expected Behavior

When I invoke a non-existent method, it should throw an error (Similar to when I invoke a non-existent app).

## Actual Behavior

It does not throw an error. Instead, app invocation is successful despite the 404. Below shows a screenshot with case (1) app ID and method name are correct, case (2) app ID is incorrect, case (3) method name is incorrect.

![image](https://user-images.githubusercontent.com/22132944/156709805-3d96b3b4-1560-4f23-9c40-313342a61e05.png)

## Steps to Reproduce the Problem

`dapr invoke` with invalid method name.

## Release Note

RELEASE NOTE: **FIX** CLI should throw an error when invalid method is invoked.

shubham1172:
As a short-term fix, we will update the dapr CLI to simply exit with error code whilst displaying the http response status code, and stop displaying the entire HTML body. Once the referred issue in runtime is fixed, we will add graceful handling here.

PR #905, opened 2022-03-02
Make CLI exit with error code if dapr run or invoke fails

shubham1172:
Signed-off-by: Shubham Sharma <shubhash@microsoft.com>

# Description

This PR looks out for errors from Dapr command and app command, and bubbles them to the Dapr CLI. 
It also adds error code for `dapr invoke` failures.

## Scenario 1 

Taken from the original issue, run Dapr without installing it.

### Before
```bash
$ dapr run test
❌  stat /home/shubham1172/.dapr/components: no such file or directory
$ echo $>
0
```
### Now
```bash
$ dapr run test
❌  stat /home/shubham1172/.dapr/components: no such file or directory
$ echo $>
1
```

## Scenario 2 

Run an app that exits with a error code.

Say an example script:
```sh
#!/usr/bin/env bash
exit 1
```

### Before
<details><summary>CLI logs</summary>

```
dapr run --log-level debug ./test.sh
ℹ️  Starting Dapr with id Browstream-Fancier. HTTP Port: 35199. gRPC Port: 35617
ℹ️  Checking if Dapr sidecar is listening on HTTP port 35199
INFO[0000] starting Dapr Runtime -- version 1.6.0 -- commit 4bb25fab444c4f1a1bf0ffd74293dbd4fdcea580  app_id=Browstream-Fancier instance=LAPTOP-G7VJUBHJ scope=dapr.runtime type=log ver=1.6.0
INFO[0000] log level set to: info                        app_id=Browstream-Fancier instance=LAPTOP-G7VJUBHJ scope=dapr.runtime type=log ver=1.6.0
INFO[0000] metrics server started on :41729/             app_id=Browstream-Fancier instance=LAPTOP-G7VJUBHJ scope=dapr.metrics type=log ver=1.6.0
INFO[0000] standalone mode configured                    app_id=Browstream-Fancier instance=LAPTOP-G7VJUBHJ scope=dapr.runtime type=log ver=1.6.0
INFO[0000] app id: Browstream-Fancier                    app_id=Browstream-Fancier instance=LAPTOP-G7VJUBHJ scope=dapr.runtime type=log ver=1.6.0
INFO[0000] mTLS is disabled. Skipping certificate request and tls validation  app_id=Browstream-Fancier instance=LAPTOP-G7VJUBHJ scope=dapr.runtime type=log ver=1.6.0
[...]

PR #910, opened 2022-03-04
Handle HTTP status checks in invoke

shubham1172:

# Description

This PR adds HTTP status code handling in `dapr invoke`.

## Issue reference

Please reference the issue this PR will close: #909

## Checklist

* [X] Code compiles correctly
* [x] Created/updated tests
* (NA) Extended the documentation

### DIFF
--- a/pkg/standalone/invoke.go
+++ b/pkg/standalone/invoke.go
@@ -68,6 +68,10 @@ func makeEndpoint(lo ListOutput, method string) string {
 }
 
 func handleResponse(response *http.Response) (string, error) {
+	if response.StatusCode < 200 || response.StatusCode >= 400 {
+		return "", fmt.Errorf("%s", response.Status)
+	}
+
 	rb, err := ioutil.ReadAll(response.Body)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


shubham1172:
As per the current approach, I am checking for any other status code besides the range [200, 400), and throwing an error with the status message.
image

Do you think it will be helpful to handle each status code separately and have a unique message for them? For example, translate "404 Not Found" to "method name is incorrect"? But that always might be not true, say there is a bug in CLI and it's calling the wrong URL, this will be misleading to the users.

/cc @mukundansundar @pravinpushkar

pravinpushkar:
> As per the current approach, I am checking for any other status code besides the range [200, 400), and throwing an error with the status message. image
> 
> Do you think it will be helpful to handle each status code separately and have a unique message for them? For example, translate "404 Not Found" to "method name is incorrect"? But that always might be not true, say there is a bug in CLI and it's calling the wrong URL, this will be misleading to the users.
> 
> /cc @mukundansundar @pravinpushkar

@shubham1172 It looks fine to me. but If we have a meaningful `response.Body` here, may be we can use that along with current status.

shubham1172:
As discussed, we have created another issue in dapr/dapr for long-term solution for method invocation error (issue 4344). This PR will return error from `Invoke` whenever dapr/dapr does not return a valid status code. Tests will pass once #905 merges.

mukundansundar (approved):
lgtm

@yaron2 thoughts?

yaron2:
> lgtm
> 
> @yaron2 thoughts?

As the CLI is an aid/tool, this change seems fine to me. I've left a comment regarding https://github.com/dapr/dapr/issues/4344.

Issue #892, opened 2022-02-11
Renew or Replace auto-generated TLS certificates using the CLI


berndverst:
This is a request for a CLI command to easily renew / replace auto-generated Dapr TLS certificates.

`dapr upgrade --renew-certificates`
`dapr upgrade -k --renew-certificates`

This would generate new certificates and:
1. Update the Dapr Trust Bundle Secret
2. Restart the Dapr Sentry Service
3. Restart the rest of the Dapr Control Plane

Additionally, there needs to be an option to specify a private key / root password to be used for certificate generation.

`dapr upgrade --renew-certificates --certificate-password-file myprivatekey.pass`

The contents of this private key file would need to be used to sign the new auto-generated TLS certificates.

pravinpushkar:
For kubrenetes, we will have below 3 options - 

```
1. dapr upgrade -k --renew-certificates
2. dapr upgrade -k --renew-certificates --certificate-password-file myprivatekey.pass
3. dapr upgrade -k --certificate-authority-root-certificate <ca.crt> --issuer-private-key<issuer.key> --issuer-public-certificate<issuer.crt> 
```
@berndverst @mukundansundar

pravinpushkar:
I think it would be better to use subcommand for renew certificate. It will be easier to organize code this way.

```
~ ❯ ./dist/linux_amd64/release/dapr upgrade renew-certificate -h
Rotates Dapr root certificate of your kubernetes cluster

Usage:
  dapr upgrade renew-certificate [flags]

Examples:

# Generates new root and issuer certificates for kubernetest cluster
dapr upgrade renew-certificate -k

# Uses existing private root.key to generate new root and issuer certificates for kubernetest cluster
dapr upgrade renew-certificate -k --certificate-password-file myprivatekey.key

# Rotates certificate of your kubernetes cluster with provided ca.cert, issuer.crt and issuer.key file path
dapr upgrade renew-certificate -k --ca-root-certificate <ca.crt> --issuer-private-key<issuer.key> --issuer-public-certificate<issuer.crt>

# See more at: https://docs.dapr.io/getting-started/


Flags:
  -h, --help         help for renew-certificate
  -k, --kubernetes   Upgrades/Renews root certificate of Dapr in a Kubernetes cluster

Global Flags:
      --log-as-json   Log output in JSON format
```

mukundansundar:
@pravinpushkar I agree, a subcommand is better to a flag in this scenario. 
[...]

PR #912, opened 2022-03-04
Changes for cli command to renew certificate

pravinpushkar:

# Description

Adding new commands to renew expiring certificate.
```
# Generates new root and issuer certificates for kubernetest cluster
dapr mtls renew-certificate -k --valid-until <no of days> --restart

# Uses existing private root.key to generate new root and issuer certificates for kubernetes cluster
dapr mtls renew-certificate -k --certificate-password-file myprivatekey.key --valid-until <no of days>

# Rotates certificate of your kubernetes cluster with provided ca.cert, issuer.crt and issuer.key file path
dapr mtls renew-certificate -k --ca-root-certificate <ca.crt> --issuer-private-key <issuer.key> --issuer-public-certificate <issuer.crt> --restart
```

## Issue reference

Please reference the issue this PR will close: #892 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/cmd/mtls.go
+++ b/cmd/mtls.go
@@ -101,6 +101,6 @@ func init() {
 	MTLSCmd.MarkFlagRequired("kubernetes")
 	MTLSCmd.AddCommand(ExportCMD)
 	MTLSCmd.AddCommand(ExpiryCMD)
-	MTLSCmd.AddCommand(RenewCertificateCmd)
+	MTLSCmd.AddCommand(RenewCertificateCmd())
 	RootCmd.AddCommand(MTLSCmd)
 }
--- a/cmd/mtls.go
+++ b/cmd/mtls.go
@@ -101,5 +101,6 @@ func init() {
 	MTLSCmd.MarkFlagRequired("kubernetes")
 	MTLSCmd.AddCommand(ExportCMD)
 	MTLSCmd.AddCommand(ExpiryCMD)
+	MTLSCmd.AddCommand(RenewCertificateCmd)
 	RootCmd.AddCommand(MTLSCmd)
 }
--- a/cmd/renew_certificate.go
+++ b/cmd/renew_certificate.go
@@ -25,7 +25,6 @@ var (
 	caRootCertificateFile       string
 	issuerPrivateKeyFile        string
 	issuerPublicCertificateFile string
-	newCertDir                  string
 	validUntil                  int
 )
 
@@ -47,17 +46,29 @@ dapr upgrade renew-cert -k --ca-root-certificate <ca.crt> --issuer-private-key <
 	Run: func(cmd *cobra.Command, args []string) {
 		if kubernetesMode {
 			if caRootCertificateFile != "" && issuerPrivateKeyFile != "" && issuerPublicCertificateFile != "" {
-				kubernetes.RenewCertificate(caRootCertificateFile, issuerPrivateKeyFile, issuerPublicCertificateFile)
+				err := kubernetes.RenewCertificate(kubernetes.RenewCertificateParams{
+					RootCertificateFilePath:   caRootCertificateFile,
+					IssuerCertificateFilePath: issuerPrivateKeyFile,
+					IssuerPrivateKeyFilePath:  issuerPublicCertificateFile,
+				})
+				if err != nil {
+					fmt.Errorf("certificate creation failed")
+				}
 			} else if certificatePasswordFile != "" {
 				fmt.Println("Reuse root password to generatre th root.pem file")
-				_, err := kubernetes.GenerateNewCertificates(validUntil, newCertDir, certificatePasswordFile)
+				err := kubernetes.RenewCertificate(kubernetes.RenewCertificateParams{
+					RootPrivateKeyFilePath: certificatePasswordFile,
+					ValidUntil:             validUntil,
+				})
 				if err != nil {
 					fmt.Errorf("certificate creation failed")
 				}
 
 			} else {
 				fmt.Println("Generate fresh certificate")
-				_, err := kubernetes.GenerateNewCertificates(validUntil, newCertDir, "")
+				err := kubernetes.RenewCertificate(kubernetes.RenewCertificateParams{
+					ValidUntil: validUntil,
+				})
 				if err != nil {
 					fmt.Errorf("certificate creation failed")
 				}
@@ -65,7 +76,6 @@ dapr upgrade renew-cert -k --ca-root-certificate <ca.crt> --issuer-private-key <
 		} else {
 			fmt.Println("standalone mode")
 		}
-		fmt.Println("in subcommand renew-certificate")
 	},
 }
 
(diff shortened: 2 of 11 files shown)


cmd/mtls.go
```diff
@@ -101,5 +101,6 @@ func init() {
 	MTLSCmd.MarkFlagRequired("kubernetes")
 	MTLSCmd.AddCommand(ExportCMD)
 	MTLSCmd.AddCommand(ExpiryCMD)
+	MTLSCmd.AddCommand(RenewCertificateCmd)
```
shubham1172, line 104: nit
```suggestion
	MTLSCmd.AddCommand(RenewCertificateCMD)
```
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

Since we are already using mixed conventions, we can stick to the recommended one.

cmd/renew_certificate.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/dapr/cli/pkg/kubernetes"
+	"github.com/dapr/cli/pkg/print"
+	"github.com/dapr/cli/utils"
+)
+
+var (
+	certificatePasswordFile     string
+	caRootCertificateFile       string
```
shubham1172, line 30: nit:
```suggestion
	CARootCertificateFile       string
```

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

cmd/mtls.go
shubham1172, line 104: Should the `mtls` command reside in `mtls.go`, just like `export` and `expiry` commands?

cmd/renew_certificate.go
```diff
@@ -0,0 +1,134 @@
...
+	issuerPrivateKeyFile        string
+	issuerPublicCertificateFile string
+	validUntil                  int
+	restartDaprServices         bool
+)
+
+var RenewCertificateCmd = &cobra.Command{
+	Use:   "renew-cert",
```
shubham1172, line 38: For consistency
```suggestion
	Use:   "renew-certificate",
```

cmd/renew_certificate.go
```diff
@@ -0,0 +1,134 @@
...
+	Short: "Rotates Dapr root certificate of your kubernetes cluster",
+	Example: `
+# Generates new root and issuer certificates for kubernetest cluster
+dapr mtls renew-cert -k --valid-until <no of days> --restart true
```
shubham1172, line 42: Other mtls commands don't seem to support the `-k` flag. Is that expected? 

```bash
dapr mtls expiry -k
Error: unknown shorthand flag: 'k' in -k
```

cmd/mtls.go
pravinpushkar: I created new file on purpose. IMO it would have made mtls file bulky and hard to read. At the same time line 104 give us idea about the new subcommand.

shubham1172: I would vote for creating a new directory for mtls then. 

Example: https://github.com/argoproj/argo-workflows/blob/master/cmd/argo/commands/cron/root.go

cmd/renew_certificate.go
pravinpushkar: this is expected in case we want to enhance it for standalone.
`export` and `expiry` does not make much sense for standalone mode imo.
In general, I think it should be with the specific subcommands whether -k flag is needed or not.

shubham1172: If `mtls` only supports k8s, then the subcommands should also enforce it ideally. Instead of silently failing, we should complain that it's a required flag and only k8s mode is supported. This can be tracked separately, but I think we should do it. Thoughts? 

/cc @mukundansundar

pravinpushkar: I will be adding `RenewCertificateCmd.MarkFlagRequired("kubernetes")` in new commit.
We can enforce this flag through all subcommands but it will change the current behavior from  `dapr mtls expiry` to of `dapr mts expiry -k` 
@mukundansundar

yaron2 (dismissed):

artursouza (changes requested):

tests/e2e/kubernetes/kubernetes_test.go
```diff
@@ -152,9 +152,8 @@ func TestKubernetesHAModeMTLSEnabled(t *testing.T) {
 	})...)
 
 	tests = append(tests, common.GetTestsOnUninstall(currentVersionDetails, common.TestOptions{
-		UninstallAll: true,
```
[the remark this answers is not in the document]
pravinpushkar: Basically, `ComponentsTestOnInstallUpgrade` is failing if we have more tests after `UninstallAll`. This was not the case earlier.
Has there been any change in uninstall --all related to removal of resources or components?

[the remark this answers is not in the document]
pravinpushkar: I have also modified some certificate renewal tests, there also I was noticing time out because of a flawed logic in my tests and I was having 3 time taking tests in one test block.

cmd/renew_certificate.go
```diff
@@ -0,0 +1,154 @@
...
+	command := &cobra.Command{
+		Use:   "renew-certificate",
+		Short: "Rotates Dapr root certificate of your kubernetes cluster",
+
+		Example: `
+# Generates new root and issuer certificates for kubernetes cluster
+dapr mtls renew-certificate -k --valid-until <no of days> --restart
+
+# Uses existing private root.key to generate new root and issuer certificates for kubernetes cluster
+dapr mtls renew-certificate -k --certificate-password-file myprivatekey.key --valid-until <no of days>
```
mukundansundar, line 47: `--private-key`

cmd/renew_certificate.go
```diff
@@ -0,0 +1,154 @@
...
+						RootCertificateFilePath:   caRootCertificateFile,
+						IssuerCertificateFilePath: issuerPublicCertificateFile,
+						IssuerPrivateKeyFilePath:  issuerPrivateKeyFile,
+						Timeout:                   timeout,
+					})
+					if err != nil {
+						logErrorAndExit(err)
+					}
+				} else if privateKey != "" {
+					print.InfoStatusEvent(os.Stdout, "Using password file to generate root certificate")
```
mukundansundar, line 70:
```suggestion
					print.InfoStatusEvent(os.Stdout, "Using existing private key file to generate root certificate")
```

tests/e2e/kubernetes/kubernetes_test.go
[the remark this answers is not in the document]
pravinpushkar: corrected all tests

tests/e2e/common/common.go
```diff
@@ -386,6 +398,78 @@ func CRDTest(details VersionDetails, opts TestOptions) func(t *testing.T) {
...
+			"--issuer-public-certificate", "./certs/issuer.crt",
+			"--restart",
+		}
+		output, err := spawn.Command(daprPath, args...)
+		t.Log(output)
+		require.NoError(t, err, "expected no error on certificate renewal")
+		assert.Contains(t, output, "Certificate rotation is successful!")
+
+		// remove cert directory created earlier.
+		os.RemoveAll("./certs")
```
mukundansundar, line 443: not like to have a folder created by one test removed by another .... can this be in the Kubernetes_test.go file instead of common ?
A single flow ... 
1. create a folder `./certs`
2. Run GenerateNewCerts -- export cert to the previously created folder 
3. Run the UserProvidedNewCertsAndRenew test ....
4. Remove the ./certs folder that was created in step 1

1 and 4 in the Test function ... and the rest can be what it is now ...

mukundansundar (commented):
few more changes please ...

mukundansundar:
@yaron2 @artursouza Don't we also need to restart sidecar-injector and dashboard as part of certificate rotation ?

yaron2:
> @yaron2 @artursouza Don't we also need to restart sidecar-injector and dashboard as part of certificate rotation ?

Sentry watches certs from the filesystem.

artursouza (dismissed):

cmd/renew_certificate.go
yaron2, line 47: Change `--certificate-password-file` to `--private-key`.

yaron2 (dismissed):

mukundansundar:
During release validations during endgame ... test this against a cluster running apps and validate working.

mukundansundar:
@pravinpushkar Wait for all containers to be up needs to be added in this PR.

cmd/renew_certificate.go
mukundansundar: This change needs to be done?

mukundansundar (approved):

Issue #726, opened 2021-05-25
Error - Docker Pull rate limiting.


mukundansundar:
CLI GitHub Workflow / CI runs encounter Docker Pull rate limits.

## Expected Behavior

No issues on pulling new Docker container

## Actual Behavior

{"time":"2021-05-25T15:15:33.488371Z","status":"failure","msg":"Unable to find image 'daprio/dapr:1.1.1' locally\ndocker: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.\nSee 'docker run --help'.\n"}

https://github.com/dapr/cli/runs/2666586343?check_suite_focus=true#step:5:180

## Steps to Reproduce the Problem

Known Docker Pull limit

## Release Note

RELEASE NOTE: Default `dapr init` to use GitHub container registry and have Docker Hub as a fallback

daixiang0:
I think there are some solutions:
1. add docker user token
2. download image into cache, while need to update when image upgrade

@dapr/maintainers-cli WDYT?

artursouza:
@daixiang0 LGTM. We can officially document that `docker login` is the way to workaround Docker rate limiting.

mukundansundar:
Based on the previous comments this is more of a documentation issue now.

paulyuk:
Let's fix this for real.  The images we need are now in GHCR per https://github.com/dapr/dapr/pull/4365

I'll PR the fix.

paulyuk:
I spoke to @mukundansundar and @pravinpushkar and recognize now there is a lot more too it and they are already working on it.  The work likely spans to dapr v1.8.  I will contribute to them versus PR myself.

paulyuk:
I do suggest P1 because the rate limiting problem is becoming a real problem for our tests and github actions.

PR #907, opened 2022-03-02
dapr init: enable redis query to work in self-hosted environment

dmitsh:
# Description

Enable Redis Query to work in self-hosted developer deployments with Dapr init

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #906

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

PR #930, opened 2022-03-24
 Switching default to Docker Hub and Minor code refactoring

pravinpushkar:

# Description

Some code refactoring for image pull logic on dapr init

## Issue reference

Please reference the issue this PR will close: NA

related to #726 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -123,6 +123,7 @@ type daprImageInfo struct {
 	ghcrImageName      string
 	dockerHubImageName string
 	imageRegistryURL   string
+	imageRegistryName  string
 }
 
 // Check if the previous version is already installed.
@@ -299,6 +300,7 @@ func runZipkin(wg *sync.WaitGroup, errorChan chan<- error, info initInfo) {
 			ghcrImageName:      zipkinGhcrImageName,
 			dockerHubImageName: zipkinDockerImageName,
 			imageRegistryURL:   info.imageRegistryURL,
+			imageRegistryName:  defaultImageRegistryName,
 		})
 		if err != nil {
 			errorChan <- err
@@ -364,6 +366,7 @@ func runRedis(wg *sync.WaitGroup, errorChan chan<- error, info initInfo) {
 			ghcrImageName:      redisGhcrImageName,
 			dockerHubImageName: redisDockerImageName,
 			imageRegistryURL:   info.imageRegistryURL,
+			imageRegistryName:  defaultImageRegistryName,
 		})
 		if err != nil {
 			errorChan <- err
@@ -416,6 +419,7 @@ func runPlacementService(wg *sync.WaitGroup, errorChan chan<- error, info initIn
 		ghcrImageName:      daprGhcrImageName,
 		dockerHubImageName: daprDockerImageName,
 		imageRegistryURL:   info.imageRegistryURL,
+		imageRegistryName:  defaultImageRegistryName,
 	})
 	if err != nil {
 		errorChan <- err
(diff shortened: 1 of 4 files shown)


artursouza (approved):

mukundansundar:
@pravinpushkar Please add Unit test for the refactored function ...

mukundansundar:
@yaron2 Please review this PR ...

yaron2:
@pravinpushkar please make Dockerhub the default - this was discussed and agreed upon with @artursouza and @mukundansundar.

mukundansundar:
@yaron2 Please review ...

pkg/standalone/standalone.go
```diff
@@ -1062,3 +1066,21 @@ func getPlacementImageWithTag(name, version string) string {
...
 	return fmt.Sprintf("%s:%s", name, version)
 }
+
+func resolveImageURI(imageInfo daprImageInfo) (string, error) {
+	if imageInfo.imageRegistryURL != "" {
+		if imageInfo.imageRegistryURL == ghcrURI || imageInfo.imageRegistryURL == "docker.io" {
+			err := fmt.Errorf("flag %s not set correctly", "--image-registry")
+			return "", err
+		}
+		return fmt.Sprintf(privateRegTemplateString, imageInfo.imageRegistryURL, imageInfo.ghcrImageName), nil
```
yaron2, line 1076: Should this be `imageInfo.ghcrImageName` is `imageRegistryURL` is set to Docker?

pravinpushkar: @yaron2 here `imageRegistryURL` will be provided by user. It is for user who have images in their private registry. 
Also, in case of private registries we decided to keep images under similar directory structure as it for ghcr.

example - 
private registry can be - `docker.io/pravin` or any other value like `ghcr.io/<pravin|some_username>`. User can not just have `docker.io` or `ghcr.io` or `ghcr.io/dapr`. Last one is to avoid conflict with existing dapr `Github container registry`.
and expectation is, this user will have images under his registry like below - 
dapr/dapr:<version>
dapr/3rdparty/zipkin
dapr/3rdparty/rejon

pkg/standalone/standalone.go
```diff
@@ -55,7 +55,7 @@ const (
 	ghcrURI                     = "ghcr.io/dapr"
 	daprGhcrImageName           = "dapr"
 	dockerContainerRegistryName = "dockerhub"
-	redisGhcrImageName          = "3rdparty/redis"
+	redisGhcrImageName          = "3rdparty/rejson"
```
artursouza, line 58: We cannot default to rejson because it is amd64 arch only. The redis image we have now support all architectures. We should default to redis and (optionally) allow override to rejson if needed.

artursouza (changes requested):

pkg/standalone/standalone.go
mukundansundar: @artursouza The only thing that we can do here is let the user know that they have to use `rejson` for any Redis query based tasks and it will not be part of `dapr init`. Having a flag/env var override for a specific component goes against the portability that we want with dapr. Agreed that we by default init with Redis and Zipkin but having overrides as flags/env vars is not something we want to support. 

cc @yaron2 Any thoughts on this?

mukundansundar: Alos as a follow up, docker can still run amd64 images using emulation... but that in itself has issues wit hth rejson image ... where in the rejson container crashes on QEMU emulation on M1 Mac ...  https://github.com/redis/redis/issues/10162

pravinpushkar: @artursouza we might have to revert this PR  -https://github.com/dapr/cli/pull/907/files as well then. Thoughts?

utils/utils.go
```diff
@@ -236,12 +236,12 @@ func GetDefaultRegistry(githubContainerRegistryName, dockerContainerRegistryName
 	val := strings.ToLower(os.Getenv("DAPR_DEFAULT_IMAGE_REGISTRY"))
 	switch val {
 	case "":
+		print.InfoStatusEvent(os.Stdout, "Container images will be pulled from Docker Hub")
+		return dockerContainerRegistryName, nil
+	case githubContainerRegistryName:
 		print.InfoStatusEvent(os.Stdout, "Container images will be pulled from GitHub container registry")
```
mukundansundar, line 242: minor nit:
```suggestion
		print.InfoStatusEvent(os.Stdout, "Container images will be pulled from Dapr GitHub container registry")
```

yaron2 (approved):

mukundansundar (approved):

pkg/standalone/standalone.go
artursouza: LGTM

artursouza (approved):

Issue #939, opened 2022-03-30
dapr init pulling from dockerhub even if private registry is supplied


shivamkm07:
## Expected Behavior

dapr init --image-registry <private-registry> pulls from private-registry always, irrespective of environment variables.

## Actual Behavior

If default image registry is set to ghcr, and the corresponding image is not there in  ghcr, it would pull from dockerhub even if --image-registry flag is supplied.

## Steps to Reproduce the Problem

export DAPR_DEFAULT_IMAGE_REGISTRY="ghcr"
dapr init --image-registry <private_registry>

It would give logs like this:
- - -
⌛ Making the jump to hyperspace...
⚠ Flag --image-registry is a preview feature and is subject to change. It is only available from CLI version 1.7 onwards.
ℹ️ Container images will be pulled from Dapr GitHub container registry
ℹ️ Installing runtime version 1.6.1
→ Downloading binaries and setting up components... ℹ️ Placement image not found in Github container registry, pulling it from Docker Hub
↙ Downloading binaries and setting up components...
Dapr runtime installed to /home/pravinpushkar/.dapr/bin, you may run the following to add it to your path if you want to run daprd directly:
export PATH=$PATH:/home/pravinpushkar/.dapr/bin
✅ Downloading binaries and setting up components...
✅ Downloaded binaries and completed components set up.
ℹ️ daprd binary has been installed to /home/pravinpushkar/.dapr/bin.
ℹ️ dapr_placement container is running.
ℹ️ dapr_redis container is running.
ℹ️ dapr_zipkin container is running.
ℹ️ Use `docker ps` to check running containers.
✅ Success! Dapr is up and running. To get started, go here: https://aka.ms/dapr-getting-started
- - -

## Release Note

RELEASE NOTE: **FIX** Bug in CLI when pulling from private registry

mukundansundar:
closed by #940

PR #940, opened 2022-03-30
Fix for image pull from private registry

pravinpushkar:

# Description

Added a check to not retry pull and not print unnecessary log when image-registry flag is provided

## Issue reference

Please reference the issue this PR will close: #939 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -146,9 +146,11 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		if !dockerInstalled {
 			return errors.New("could not connect to Docker. Docker may not be installed or running")
 		}
-		defaultImageRegistryName, err = utils.GetDefaultRegistry(githubContainerRegistryName, dockerContainerRegistryName)
-		if err != nil {
-			return err
+		if imageRegistryURL == "" {
+			defaultImageRegistryName, err = utils.GetDefaultRegistry(githubContainerRegistryName, dockerContainerRegistryName)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
@@ -427,7 +429,7 @@ func runPlacementService(wg *sync.WaitGroup, errorChan chan<- error, info initIn
 	}
 	image = getPlacementImageWithTag(image, info.runtimeVersion)
 
-	if defaultImageRegistryName == githubContainerRegistryName {
+	if info.imageRegistryURL == "" && defaultImageRegistryName == githubContainerRegistryName {
 		if !TryPullImage(image) {
 			print.InfoStatusEvent(os.Stdout, "Placement image not found in Github container registry, pulling it from Docker Hub")
 			image = getPlacementImageWithTag(daprDockerImageName, info.runtimeVersion)
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -146,7 +146,9 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		if !dockerInstalled {
 			return errors.New("could not connect to Docker. Docker may not be installed or running")
 		}
-		if useDefaultReg(imageRegistryURL, fromDir) {
+
+		// Initialise default registry only if it is not airgap mode or init for private registries.
+		if len(strings.TrimSpace(imageRegistryURL)) == 0 && len(strings.TrimSpace(fromDir)) == 0 {
 			defaultImageRegistryName, err = utils.GetDefaultRegistry(githubContainerRegistryName, dockerContainerRegistryName)
 			if err != nil {
 				return err
@@ -430,7 +432,7 @@ func runPlacementService(wg *sync.WaitGroup, errorChan chan<- error, info initIn
 	}
 	image = getPlacementImageWithTag(image, info.runtimeVersion)
 
-	if checkFallbackImg(imgInfo, info.fromDir) {
+	if checkFallbackImgForPlacement(imgInfo, info.fromDir) {
 		if !TryPullImage(image) {
 			print.InfoStatusEvent(os.Stdout, "Placement image not found in Github container registry, pulling it from Docker Hub")
 			image = getPlacementImageWithTag(daprDockerImageName, info.runtimeVersion)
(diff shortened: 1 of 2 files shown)


pkg/standalone/standalone.go
```diff
@@ -1067,6 +1070,20 @@ func getPlacementImageWithTag(name, version string) string {
...
 
+func checkFallbackImg(imageInfo daprImageInfo, fromDir string) bool {
+	if imageInfo.imageRegistryURL != "" || fromDir != "" {
+		return false
+	}
+	return imageInfo.imageRegistryName == githubContainerRegistryName
+}
+
+func useDefaultReg(privateReg, fromDir string) bool {
+	if privateReg != "" || fromDir != "" {
```
shubham1172, line 1081:
```suggestion
	return privateReg == "" && fromDir == ""
```

shubham1172: Can we also add a comment on what's the meaning of this function?

pkg/standalone/standalone.go
```diff
@@ -1067,6 +1070,20 @@ func getPlacementImageWithTag(name, version string) string {
...
 }
 
+func checkFallbackImg(imageInfo daprImageInfo, fromDir string) bool {
+	if imageInfo.imageRegistryURL != "" || fromDir != "" {
+		return false
+	}
+	return imageInfo.imageRegistryName == githubContainerRegistryName
+}
+
+func useDefaultReg(privateReg, fromDir string) bool {
```
shubham1172, line 1080: Is `privateReg` supposed to be a URL? We can make the variable name reflect that.

pkg/standalone/standalone.go
```diff
@@ -146,9 +146,11 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		if !dockerInstalled {
 			return errors.New("could not connect to Docker. Docker may not be installed or running")
 		}
-		defaultImageRegistryName, err = utils.GetDefaultRegistry(githubContainerRegistryName, dockerContainerRegistryName)
-		if err != nil {
-			return err
+		if useDefaultReg(imageRegistryURL, fromDir) {
```
shubham1172, line 149: If this is only used here, we can also simply add the condition here instead of creating a one-liner function.

pkg/standalone/standalone.go
surenderssm, line 1081: What happens if `privateReg` or `fromDir` is a whitespace?
Suggestion : 

`if len(strings.TrimSpace(privateReg)) != 0   ||  len(strings.TrimSpace(fromDir) != 0`

pkg/standalone/standalone.go
surenderssm: URL => URI

pkg/standalone/standalone_test.go
```diff
@@ -232,3 +232,99 @@ func TestResolveImageErr(t *testing.T) {
 		})
 	}
 }
+
+func TestCheckFallbackImg(t *testing.T) {
+	daprImgWithPrivateRegAndDefAsDocker := daprImageInfo{
+		ghcrImageName:      daprGhcrImageName,
+		dockerHubImageName: daprDockerImageName,
+		imageRegistryURL:   "example.io/user",
+		imageRegistryName:  "dockerhub",
```
surenderssm, line 241: nit: moving this as constant might help

pkg/standalone/standalone.go
```diff
@@ -415,19 +417,20 @@ func runPlacementService(wg *sync.WaitGroup, errorChan chan<- error, info initIn
...
 		errorChan <- err
 		return
 	}
 	image = getPlacementImageWithTag(image, info.runtimeVersion)
 
-	if defaultImageRegistryName == githubContainerRegistryName {
+	if checkFallbackImg(imgInfo, info.fromDir) {
 		if !TryPullImage(image) {
 			print.InfoStatusEvent(os.Stdout, "Placement image not found in Github container registry, pulling it from Docker Hub")
 			image = getPlacementImageWithTag(daprDockerImageName, info.runtimeVersion)
```
DeepanshuA, line 436: Seems only placement image has a fallback. If that is the case, could you please update method comments to reflect the intent, so that a future developer is quickly able to get context.

pravinpushkar: I have updated the method name to more idea about the use

pkg/standalone/standalone.go
```diff
@@ -146,9 +146,13 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		if !dockerInstalled {
 			return errors.New("could not connect to Docker. Docker may not be installed or running")
 		}
-		defaultImageRegistryName, err = utils.GetDefaultRegistry(githubContainerRegistryName, dockerContainerRegistryName)
-		if err != nil {
-			return err
+
+		// Initialise default registry only if it is not airgap mode or init for private registries.
```
shubham1172, line 150:
```suggestion
		// Initialize default registry only if --image-registry and --from-dir are not provided.
```

pkg/standalone/standalone.go
```diff
@@ -1067,6 +1072,13 @@ func getPlacementImageWithTag(name, version string) string {
 	return fmt.Sprintf("%s:%s", name, version)
 }
 
+func checkFallbackImgForPlacement(imageInfo daprImageInfo, fromDir string) bool {
```
shubham1172, line 1075: Can we add comment to this fn?

pkg/standalone/standalone.go
shubham1172: Can we rename it to convey what it's doing? Something like `if useGitHubRegistry(imgInfo, info.fromDir)`.

shubham1172 (changes requested):
LGTM overall, have some nit requests.

pkg/standalone/standalone.go
pravinpushkar: I have updated this to `checkFallbackImgForPlacement`.Basically this function is just checking whethere this `init` flow qualifies for `tryPull` or not

shubham1172: With the `check` prefix in `checkFallbackImgForPlacement`, it might not be very apparent on (1) what the function does and (2) what's the consumer's intent. If this returns true, we "check" fallback image for placement service - is slightly confusing. Instead, if it says "pullFromGithubRegistry", or "useGitHubRegistry", it might both justify what the method is doing, and how the consumer is using it. Thoughts?

shubham1172 (approved):

mukundansundar (approved):

Issue #975, opened 2022-05-05
CLI does not install Resiliency CRD


yaron2:
When upgrading from Dapr 1.X to 1.7.X or when installing a new 1.7.x Dapr installation on Kubernetes, the resiliency CRD isn't applied to the cluster.

This causes the resiliency feature to be unusable as users can't deploy the CRs to the cluster.

yaron2:
Closed via #977.

PR #976, opened 2022-05-05
Fix CRD handling for upgrades and removals

yaron2:
Fixes https://github.com/dapr/cli/issues/975.

This PR does two things:

1. Fixes issue #975 by adding the resiliency CRD to the list of CRDs to be applied to the cluster and removed when uninstalled
2. Fixes an issue where a user tries to install a version of Dapr (for example, 1.1.0) with a CLI that tries to apply CRDs that are not available in the desired version (for example, trying to install Dapr 1.1.0 with CLI 1.5.0 will fail because the CLI can't find the subscription CRD for 1.1.0). This is done by checking if the CRD exists for the required version before attempting to apply it to the cluster.
3. Fixes an issue where users trying to uninstall an older version of Dapr with a newer CLI is failing because the older version doesn't have a CRD that the CLI is trying to uninstall. (for example, trying to remove Dapr 1.6.0 with the 1.7.0 CLI will fail because it can't find a resiliency policy). The PR treats the removal of a CRD as a warning instead of an error.


yaron2:
Wrong branch..

Issue #954, opened 2022-04-06
Dapr CLI Cannot Upgrade Dapr Cluster on Edge Version


halspang:
## Expected Behavior
When a k8s cluster is running an edge version of Dapr, it should be able to be upgraded to a released version.

## Actual Behavior
Instead of upgrading, it panics on a nil pointer when trying to determine if it's an upgrade or downgrade.

## Steps to Reproduce the Problem
1. Deploy Dapr with a locally built (edge) version.
2. Use the Dapr CLI to upgrade to a different version `dapr upgrade -k --runtime-version 1.7.0-rc.3`

```
➜  ~/Workspace/cli git:(master) 
$ dapr status -k
  NAME                   NAMESPACE    HEALTHY  STATUS   REPLICAS  VERSION          AGE  CREATED              
  dapr-operator          dapr-system  True     Running  1         dev-linux-amd64  4d   2022-04-01 16:23.57  
  dapr-placement-server  dapr-system  True     Running  1         dev-linux-amd64  4d   2022-04-01 16:23.57  
  dapr-dashboard         dapr-system  True     Running  1         0.9.0            4d   2022-04-01 16:23.56  
  dapr-sidecar-injector  dapr-system  True     Running  1         dev-linux-amd64  4d   2022-04-01 16:23.58  
  dapr-sentry            dapr-system  True     Running  1         dev-linux-amd64  4d   2022-04-01 16:23.56

➜  ~/Workspace/cli git:(master) 
$ dapr upgrade -k --runtime-version 1.7.0-rc.3
ℹ️  Dapr control plane version dev-linux-amd64 detected in namespace dapr-system
ℹ️  Starting upgrade...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x2505ad1]

goroutine 1 [running]:
github.com/hashicorp/go-version.(*Version).String(0x0, 0xc000cb3366, 0xa)
        /Users/runner/go/pkg/mod/github.com/hashicorp/go-version@v1.3.0/version.go:371 +0x51
github.com/hashicorp/go-version.(*Version).Compare(0xc000bcc370, 0x0, 0xc0003da000)
        /Users/runner/go/pkg/mod/github.com/hashicorp/go-version@v1.3.0/version.go:118 +0x51
github.com/hashicorp/go-version.(*Version).LessThan(...)
[...]

PR #983, opened 2022-05-20
Fix for panic during upgrade of edge version dapr

pravinpushkar:

# Description

Handle panic caused while trying to upgrade an edge version of Dapr.

## Issue reference

Please reference the issue this PR will close: #954 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/kubernetes/upgrade.go
+++ b/pkg/kubernetes/upgrade.go
@@ -173,7 +173,7 @@ func isDowngrade(targetVersion, existingVersion string) bool {
 	if err != nil {
 		print.FailureStatusEvent(
 			os.Stderr,
-			fmt.Sprintf("Upgrade failed, %s. Upgrading an edge version of Dapr is not supported!", err.Error()))
+			fmt.Sprintf("Upgrade failed, %s. The current installed version does not have sematic versioning", err.Error()))
 		os.Exit(1)
 	}
 	return target.LessThan(existing)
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/upgrade.go
```diff
@@ -169,7 +169,12 @@ func upgradeChartValues(ca, issuerCert, issuerKey string, haMode, mtls bool, arg
 
 func isDowngrade(targetVersion, existingVersion string) bool {
 	target, _ := version.NewVersion(targetVersion)
-	existing, _ := version.NewVersion(existingVersion)
-
+	existing, err := version.NewVersion(existingVersion)
+	if err != nil {
```
mukundansundar, line 173: check if existing is `edge` version ... or `nightly` version ... Then fail ...

mukundansundar (changes requested):

pkg/kubernetes/upgrade.go
pravinpushkar: @mukundansundar  What if the existing installation is locally built say with version `dev-linux-amd64`(as mentioned in the issue) or any other version . Here also the upgrade will fail because the `version.NewVersion()` won't be able to parse.

mukundansundar: @pravinpushkar Then the output log to the console should be different ....

pravinpushkar: @mukundansundar  Can we go with any one of these - 
1. `fmt.Sprintf("Upgrade failed, %s. The current installed version does not have sematic versioning", err.Error()))`
2. `fmt.Sprintf("Upgrade failed, %s", err.Error()))`

Issue #962, opened 2022-04-12
Add timeout to HTTP call in `downloadFile` method


ItalyPaleAle:
May fix an issue an user reported.

In this line in the `downloadFile` method (part of the `dapr init` command):

https://github.com/dapr/cli/blob/4765d02e5ab4675985aa59cbc6d5ecd91ff14b68/pkg/standalone/standalone.go#L1050

The binary is downloaded by using the default HTTP client with no timeout configured. In case of a network issue, that may cause the application to hang forever.

See how we implemented that in VS Code installer for the Windows Store (which ensures there's a minimum flow of data rather than a set timeout, to account for people on slow connections): https://github.com/microsoft/vscode-winsta11er/blob/main/common/common.go#L94

PR #994, opened 2022-06-03
Add timeout to HTTP call in downloadFile method

hueifeng:
# Description

Add timeout to HTTP call in downloadFile method

## Issue reference

#962

Please reference the issue this PR will close: #962

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


pkg/standalone/standalone.go
```diff
@@ -1047,10 +1049,27 @@ func downloadFile(dir string, url string) (string, error) {
 		return "", nil
 	}
 
-	resp, err := http.Get(url)
+	client := http.Client{
+		Timeout: 0,
```
pravinpushkar, line 1053: @hueifeng Why the `Timeout` here is set to 0. If this is required then should not we have some +ve value?

pravinpushkar, line 1053: @hueifeng Why the `Timeout` here is set to 0. If this is required then should not we have some +ve value?

hueifeng: Because file can take a while to download on slow connections, so set `Timeout` to 0.

mukundansundar: @hueifeng This can cause a process to hang forever .... Based on the comments in #962 can a method be added to track incremental progress?

pravinpushkar: There is one suggestion mentioned in the attached issue to check if some data got transferred in last 5 sec or timeout if not.

hueifeng:
hi @mukundansundar @pravinpushkar, I added the progress bar, please help me review it.

pkg/standalone/standalone.go
hueifeng: @ItalyPaleAle I add a link to mark, is it okay?

ItalyPaleAle:
This adds only a bar however which does not stop if the request is hanging. It does provide feedback to users so they can stop the process themselves.

pkg/standalone/standalone.go
```diff
@@ -1065,14 +1086,87 @@ func downloadFile(dir string, url string) (string, error) {
...
-	_, err = io.Copy(out, resp.Body)
+	_, err = copyWithTimeout(context.Background(), out, resp.Body)
 	if err != nil {
 		return "", err
 	}
 
 	return filepath, nil
 }
 
+//see https://github.com/microsoft/vscode-winsta11er/blob/main/common/common.go#L169
```
ItalyPaleAle, line 1097: We need to add the copyright notice here to be compliant with the MIT license. Please change this to:

```go
/*!
See: https://github.com/microsoft/vscode-winsta11er/blob/4b42060da64aea6f47adebe1dd654980ed87a046/common/common.go
Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License.
*/
```

ItalyPaleAle (changes requested):

ItalyPaleAle (approved):
LGTM! Thanks for this contribution :)

pravinpushkar (dismissed):

hueifeng:
@mukundansundar I repaired `lint`, But I didn't see two mistakes:

```
FallbackDelay, KeepAlive, Resolver, Cancel, Control are missing in Dialer (exhaustruct)
```

```
CheckRedirect, Jar are missing in Client (exhaustruct)
```

Please give me some help.

ItalyPaleAle:
@hueifeng you can use `//nolint: exhaustruct` to suppress those (also note there's no space after `//` when comments are meant to be machine-readable)

pravinpushkar (approved):
LGTM

mukundansundar (approved):
lgtm

Issue #953, opened 2022-04-06
Rename the run `--components-path` flag to a name for suitable for loading many resource types


msfussell:
## Describe the proposal
When the CLI came out there were only components, now we have 2 more, subscriptions and reliability spec so we need to change the naming accordingly to reflect that this path is used to load multiple types of resources. We can still keep --components-path and slowly deprecate this. 

I suggest `--resources-path` or `--spec-resources-path` given that we refer to these as specs in the docs.
Also see https://github.com/dapr/cli/issues/950 as a related issue.

## Release Note

RELEASE NOTE:

mukundansundar:
I agree that once resiliency files are added as part of CLI init, this would make sense to rename the flag. 
Also following the deprecation policy in dapr, the `--components-path` will need to be deprecated first aliased as some other appropriate flag name and later removed 2 releases later.

msfussell:
I do not think the name should contain resiliency, since there are also subscription resource types and I expect there will be more resource types in the future. Part of this is that we need to agree on a good name for these resources (staying away from CRDs which is a K8s term).

hueifeng:
>I suggest --resources-path or --spec-resources-path given that we refer to these as specs in the docs.

We use ` --spec-resources-path` is more appropriate? I will assign this issue.

mukundansundar:
@artursouza @yaron2 @msfussell Should this be changed in `daprd` binary flags as well? As `dapr` CLI just passes on the value to the `daprd` executable. 
From `daprd -help`
```
-components-path string
    	Path for components directory. If empty, components will not be loaded. Self-hosted mode only
```
In this case I also prefer the `--resources-path` instead of `spec-resources-path`.

yaron2:
> @artursouza @yaron2 @msfussell Should this be changed in `daprd` binary flags as well? As `dapr` CLI just passes on the value to the `daprd` executable. From `daprd -help`
> 
> ```
> -components-path string
>     	Path for components directory. If empty, components will not be loaded. Self-hosted mode only
> ```
> 
> In this case I also prefer the `--resources-path` instead of `spec-resources-path`.

Yes this should be a `daprd` flag as well.

msfussell:
Agree this should be a `daprd` flag

mukundansundar:
Moving to 1.10 as there are still some unresolved questions that need decision on the associated pr

pravinpushkar:
[...]

PR #1012, opened 2022-06-27
support resources-path

hueifeng:
# Description

Signed-off-by: HueiFeng <695979933@qq.com>

## Issue reference

#953 

Please reference the issue this PR will close: #953 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pkg/standalone/run.go
```diff
@@ -49,6 +49,7 @@ type RunConfig struct {
 	MaxConcurrency     int    `arg:"app-max-concurrency"`
 	PlacementHostAddr  string `arg:"placement-host-address"`
 	ComponentsPath     string `arg:"components-path"`
+	SpecResourcesPath  string `arg:"spec-resources-path"`
```
mukundansundar, line 52: getArgs() is used to get the `arg` def here and that is passed on to `daprd` executable. So if the arg is `spec-resources-path` that I believe that would break `daprd`. @hueifeng Can you verify that functionality?

mukundansundar (commented):
@hueifeng Based on the decision in the issue #953 , docs needs to be added and changes be made.
Also for the existing code please add E2E tests here.

pkg/standalone/run.go
hueifeng: Let me try it.

mukundansundar: @hueifeng Were you able to check this out?

hueifeng: @mukundansundar It looks like I forgot to add it to the `cmd` folder.

hueifeng: @mukundansundar Add `resources-path` to the `daprd` binary https://github.com/dapr/dapr/pull/4998. After the dapr repository merge, I think it's ready for commit e2e testing.

yaron2: @mukundansundar this is now merged: https://github.com/dapr/dapr/pull/4998.

msfussell:
There needs to be some clarity on next steps before this PR is merged
1) Does this PR create a local resources directory as there is for components? I believe this is needed, since this is where you can put your subscriptions and resiliency policies. See below. 

| `--components-path`, `-d`      |                      | `Linux & Mac: $HOME/.dapr/components`, `Windows: %USERPROFILE%\.dapr\components`   | The path for components directory                                                                    |
| `--resources-path`     |                      |   | The path for resources directory                                                                    |

2) Are we going to deprecate the --components-path? If so then this needs a tracking issue as part of this PR
3) The docs need more updates than just the CLI. This is an important change and needs to be covered in other docs. I have left comments into the docs issue.

addjuarez:
@hueifeng

hueifeng:
Currently we are using substitution in `CLI`, by default using `--components-path`, by way of judgement in `daprd` [binary](https://github.com/dapr/dapr/pull/4998/files). I want to reduce unnecessary code and replace `--components-path` directly with `resources-path` after the next two releases. 

2.  Can I also put Components into this directory and they are loaded by the runtime?

     Now we need to specify it with the `--resources-path` parameter.

4. Are we going to deprecate the `--components-path`? If so then this needs a tracking issue as part of this PRs next steps

    According to our initial plan, and the current `PR` is a direct remove.

Maybe I still need to do a lot of work and maybe we can discuss what's next and whether to make big changes to the `PR`. What do you guys have in mind?

cc @msfussell @yaron2 @mukundansundar @pravinpushkar

pkg/standalone/run.go
```diff
@@ -132,6 +133,10 @@ func (config *RunConfig) validate() error {
 		return err
 	}
 
+	if config.ResourcesPath != "" {
+		config.ComponentsPath = ""
```
pravinpushkar, line 137: @hueifeng is this intentional?
or do we intend to set `components-path` when `resource-path` is set ?

hueifeng: > @hueifeng is this intentional? or do we intend to set `components-path` when `resource-path` is set ?

Yes, this is intentional, what I think is to replace it slowly to `resource-path`, If we copy the code of `component-path` again, it will be a bit repetitive, after all, we only need PR to change the name.

mukundansundar (approved):
lgtm

Issue #1015, opened 2022-06-30
Commands print non-json warning when output set to json


MichaelHindley:
When using the list command(and others it seems) always print the warning.

This makes it very hard to programatically do things based on for instance json output, since the json output is not valid json due to the warning that is always printed, introducing parsing logic for  the warning before working with the json output.

```bash
➜  k8s git:(main) ✗ dapr list -k --namespace default -o json
⚠  In future releases, this command will only query the "default" namespace by default. Please use the --namespace flag for a specific namespace, or the --all-namespaces (-A) flag for all namespaces.
{
  "namespace": "default",
  "appId": "bakery",
  "appPort": "8080",
  "age": "16h",
  "created": "2022-06-29 18:01.58"
}
```

`SyntaxError: Unexpected token ⚠ in JSON at position 0`

Suggestion: If namespace is provided to the command, or json output requested, disable the warning from printing.
Or add a flag to disable the warning (--no-warning) or similar.

https://github.com/dapr/cli/blob/ce0af7c01c6b9b59d237f1a6c960bcc13783125e/cmd/list.go#L81

pravinpushkar:
I think, `--no-warning` can be a good addition here. @mukundansundar @yaron2

mukundansundar:
@MichaelHindley  I think warnings should be printed in STDERR ... Thereby if only STDOUT is read, only the output will be there .... 
Also for flag, this is too specific to this warning ... I am thinking it should probably be some form of a well known flag name like `--quiet` or `--silent` and that it should apply to all commands globally in dapr CLI ... 
@yaron2 Thoughts?

FlorentATo:
Gentle bump :)

Mukuwul:
Picked this up in #1673. The warning that corrupts `dapr list -k -o json` now goes to stderr, along with a dozen other error messages that had drifted to stdout since the #748 cleanup. Happy to adjust if there's a preference for suppressing the warning entirely under `-o json` instead.

PR #1017, opened 2022-07-04
feat: add flag to disable namespace warning

MichaelHindley:
# Description
Adds flag to disable printing warning messages for default namespaces.

print.WarningStatusEvents is also used to report on errors
that occur during command execution.

First it was considered to check for the flag within
print.WarningStatusEvent, but this does not seem in tune with
the spirit of this flag, which is to prevent breaking
output formats due to a warning that is always printed.

For the context of this flag, its thus explicitly named so
its clear only the namespace warning is covered,
and not all warnings that are generated throughout the CLI.

Names likes noWarning, disableWarning were also considered.

Single char shorthand flag was also considered.

## Issue reference

Please reference the issue this PR will close: #1015

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests N/A ?
* [ ] Extended the documentation N/A?


MichaelHindley:
@hueifeng I will when I start implementing the changes as discussed in #1015.
@mukundansundar can I assume the changes from your comment in regard to `--quiet/silent` are green to contribute?

mukundansundar:
@yaron2 Any issues with having a `--silent` or `--quite` flag to not print any warnings ?

pravinpushkar:
As suggested by @mukundansundar 
Alternate approach can be having the flag at RootCmd level like, So that it is available for every command. 
```
RootCmd.PersistentFlags().BoolVarP(&silent, "silent", "s", false, "Silence all output except errors")
```
and handle the console output in `print.go` at success, info and warning level, something like - 
```
func InfoStatusEvent(w io.Writer, fmtstr string, a ...interface{}) {
	if silent {
		return
	}
...... 
```
Thoughts @mukundansundar @MichaelHindley ?

mukundansundar:
> As suggested by @mukundansundar Alternate approach can be having the flag at RootCmd level like, So that it is available for every command.
> 
> ```
> RootCmd.PersistentFlags().BoolVarP(&silent, "silent", "s", false, "Silence all output except errors")
> ```
> 
> and handle the console output in `print.go` at success, info and warning level, something like -
> 
> ```
> func InfoStatusEvent(w io.Writer, fmtstr string, a ...interface{}) {
> 	if silent {
> 		return
> 	}
> ...... 
> ```
> 
> Thoughts @mukundansundar @MichaelHindley ?

If we want to go for this flag, it needs to be applied retroactively to all the print statements that we do today ...

pravinpushkar:
@mukundansundar yes, thats correct. In print.go file we will have to do handle this in 3 places-  success, info and warning print functions. Is there any other simpler way ?

mukundansundar:
> @mukundansundar yes, thats correct. In print.go file we will have to do handle this in 3 places-  success, info and warning print functions. Is there any other simpler way ?


Can we clearly define as to what output is something that should be there regardless of the silent flag or not 

For dapr init... dapr installed successfully alone and not any of the other output except errors if any...

Similarly for other commands that use the print.go functions for output all will be affected by the silent flag... So some clear definition of what should be affected vs what should not be will need to be thought through and appropriate changes be made.... For those conversations I suggest we do it in the parent issue and not in this PR

mukundansundar:
@MichaelHindley 
Please fix conflicts and changed implementation to what was discussed in #1015 
is there any ETA on this PR?

yaron2:
@MichaelHindley please resolve the conflicts so we can merge.

yaron2:
ping @MichaelHindley

cicoyle:
ping @MichaelHindley - I think this would be a great value add for next release if you could please fix the merge conflicts?

Issue #932, opened 2022-03-28
Automate e2e CLI validation for Windows


shubham1172:
## Describe the proposal

Currently, the e2e pipeline is not run for Windows. It needs manual validation before each release.

https://github.com/dapr/cli/blob/a1348603081a2ace4d6355ebf19471e60eb8c319/.github/workflows/self_hosted_e2e.yaml#L42-L51

With this issue, add a Windows entry as well.

RELEASE NOTE: Automate CLI validation in Windows via `--slim` installation

mukundansundar:
Keep in mind that we have to come up with a solution where in windows runners can use linux containers. See https://github.com/dapr/docs/issues/162

shubham1172:
For reference, GH actions do not support container ops on Windows runner yet https://github.com/actions/runner/issues/904

PR #1036, opened 2022-08-02
Refactor standalone e2e and add support for Windows runner

shubham1172:

# Description

This PR achieves two things majorly
1. Refactor the e2e test suite for standalone mode - it decouples the tests and allows tests to be run independently while locally running them. It also covers scenarios for slim mode and Windows operating system.
1. Introduces Windows tests - Uses --slim mode to initialize Dapr without having to use containers (LXCs are unsupported on Windows GitHub agent). This will prevent manual Windows testing with each release.

I have also added some minor checks to the tests while refactoring them.

## Issue reference

Please reference the issue this PR will close: #932 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


tests/e2e/standalone/commands.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// cmdUninstall uninstalls Dapr with --all flag and returns the command output and error.
+func cmdUninstall() (string, error) {
+	return spawn.Command(common.GetDaprPath(), "uninstall", "--log-as-json", "--all")
+}
+
+// cmdPublish publishes a message to the specified pubsub and topic, and returns the command output and error.
+func cmdPublish(appId, pubsub, topic, unixDomainSocket string, args ...string) (string, error) {
+	args = append([]string{"publish", "--log-as-json", "--publish-app-id", appId, "--pubsub", pubsub, "--topic", topic}, args...)
```
mukundansundar, line 45: application `args...` should be appended after `dapr` CLI args.

tests/e2e/standalone/init_test.go
```diff
@@ -0,0 +1,220 @@
...
+				delete(daprContainers, name)
+			}
+		}
+
+		assert.Empty(t, daprContainers, "Missing containers: %v", daprContainers)
+	})
+}
+
+// verifyBinaries ensures that the correct binaries are present in the correct path.
+func verifyBinaries(t *testing.T, daprPath, runtimeVersion, dashboardVersion string) {
```
mukundansundar, line 106: on slim mode ... verify placement binary install ...

mukundansundar (commented):
@shubham1172 how are you guaranteeing the order of execution ie the publish test runs after init etc ...
Also a lot more tests are there in standalone init that needs to be added ... 

One approach is this way, ie only test the slim installation. 
I was wondering if we should bring up a windows VM on the fly with Docker installed, enable linux containers and then run E2E against that instance, ie instead of using the GH windows runners?
@yaron2 @pravinpushkar @artursouza WDYT?

That would obviously need a separate Workflow which runs only against master since that would include secrets.

tests/e2e/standalone/init_test.go
shubham1172: > @shubham1172 how are you guaranteeing the order of execution ie the publish test runs after init etc ... 

I am not. The tests should not be tightly coupled. Today if I make changed to Publish tests, I need to run the entire Standalone test suite. This change decouples the tests. There is a function call in Publish test to ensure that Dapr is installed. This ensures that locally anyone can selectively run tests.


> Also a lot more tests are there in standalone init that needs to be added ...

Yes, this is a WIP PR, will add them shortly.

shubham1172:
> @shubham1172 how are you guaranteeing the order of execution ie the publish test runs after init etc ... 

@mukundansundar I am not. The tests should not be tightly coupled. Today if I make changed to Publish tests, I need to run the entire Standalone test suite. This change decouples the tests. There is a function call in Publish test to ensure that Dapr is installed. This ensures that locally anyone can selectively run tests.


> Also a lot more tests are there in standalone init that needs to be added ...

Yes, this is a WIP PR, will add them shortly.

tests/e2e/standalone/init_test.go
shubham1172: Only containers/components are skipped and the configuration is slightly altered (no zipkin) for slim mode.

tests/e2e/standalone/commands.go
shubham1172: Thanks, fixed.

tests/e2e/standalone/init_test.go
```diff
@@ -0,0 +1,248 @@
...
+	binPath := filepath.Join(daprPath, "bin")
+	require.DirExists(t, binPath, "Directory %s does not exist", binPath)
+
+	binaries := map[string]string{
+		"daprd":     runtimeVersion,
```
mukundansundar, line 139: Verify placement binary also installed on slim init

shubham1172: Done, thanks!

tests/e2e/standalone/init_test.go
shubham1172: Sorry I missed this earlier, I was not checking "placement". Updated, thanks!

tests/e2e/standalone/list_test.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		cmd.Process.Kill()
+	})
+
+	t.Run("dashboard instance should not be listed", func(t *testing.T) {
+		if isSlimMode() {
+			t.Skip("skipping test in slim mode")
+		}
+		go cmdDashboard("5555")
```
pravinpushkar, line 104: Do we need to ensure that dashboard is launched before we execute the cmdList?

tests/e2e/standalone/init_negative_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestStandaloneInitNegatives(t *testing.T) {
```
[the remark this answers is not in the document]
shubham1172: The problem is that it will increase the e2e test time. The negative scenarios in this file requires Dapr to be uninstalled. So if say list tests contain both tests that require Dapr to be installed and uninstalled, each time we will have to bear that overhead.

tests/e2e/standalone/list_test.go
shubham1172: Thinking more about it - `dapr dashboard` is a blocking call, so the spawn function will be hanged. We will need a special handling to ensure that this is working.

tests/e2e/standalone/init_test.go
```diff
@@ -0,0 +1,254 @@
...
+		"dashboard": dashboardVersion,
+	}
+
+	if isSlimMode() {
+		binaries["placement"] = ""
```
mukundansundar, line 144: same as runtimeVersion ...

tests/e2e/standalone/list_test.go
mukundansundar: Why is this being skipped in slim init?

tests/e2e/standalone/uninstall_test.go
```diff
@@ -0,0 +1,72 @@
...
+	assert.Contains(t, output, "Dapr has been removed successfully")
+
+	// verify that .dapr directory does not exist.
+	homeDir, err := os.UserHomeDir()
+	require.NoError(t, err, "failed to get user home directory")
+
+	daprPath := filepath.Join(homeDir, ".dapr")
+	require.NoDirExists(t, daprPath, "Directory %s does not exist", daprPath)
+
+	verifyNoContainers(t)
```
mukundansundar, line 47: how does this run successfully in windows?

tests/e2e/standalone/init_test.go
shubham1172: `placement --version` does not work, right?

tests/e2e/standalone/list_test.go
shubham1172: Removed it, thanks for the catch! Have also updated the handling mechanism, but it ends up creating a background `process` since there is no elegant way to kill the dashboard process. Let me know your thoughts.

tests/e2e/standalone/uninstall_test.go
shubham1172: Interesting, have put this test behind slim mode check.

tests/e2e/standalone/uninstall_test.go
```diff
@@ -0,0 +1,72 @@
...
+		"dapr_redis":     true,
+	}
+
+	for _, container := range containers {
+		t.Logf("Found container %v %s %s\n", container.Names, container.Image, container.State)
+		name := strings.TrimPrefix(container.Names[0], "/")
+		delete(daprContainers, name)
+	}
+
+	assert.Equal(t, 3, len(daprContainers), "Found Dapr containers still running")
```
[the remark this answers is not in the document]
shubham1172: @mukundansundar it should be exactly equal to 3, right? If any one container is running , the `delete` will remove that from the `daprContainers` map and the count will change.

tests/e2e/standalone/list_test.go
shubham1172: And this is also currently causing the Uninstall test to fail, since the dashboard binary can't be deleted.

shubham1172: FYI looks like Dapr was uninstalled in the previous versions of the test,
so dashboard command would always fail and this test would pass.

https://github.com/dapr/cli/blob/c14bd7b7c05a778ce4c16ff2e261c7270e4f4bdd/tests/e2e/standalone/standalone_test.go#L152-L157

mukundansundar:
> Are there any functionalities that do not get tested in the slim mode? VMs will be much heavier (might significantly increase e2e test time and cost).

The main reason I wanted a windows VM was to be able to run docker in the VM and test `dapr init` with docker in Windows.

tests/e2e/standalone/uninstall_test.go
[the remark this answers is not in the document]
mukundansundar: @shubham1172 Interesting inverse logic. Can you add a comment above the assert to make sure that we except no deletes to happen as we expect no containers to be running ?

mukundansundar (commented):
@shubham1172 Added few more comments, Can you also check as to why Windows e2e failed this time in Uninstal?

tests/e2e/standalone/stop_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	executeAgainstRunningDapr(t, func() {
+		t.Run("stop", func(t *testing.T) {
+			output, err := cmdStop("dapr_e2e_stop")
+			t.Log(output)
+			require.NoError(t, err, "dapr stop failed")
+			assert.Contains(t, output, "app stopped successfully: dapr_e2e_stop")
+		})
+
+		t.Run("stop with unknown flag", func(t *testing.T) {
```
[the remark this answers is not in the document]
shubham1172: Well, that is true but the test should fail nonetheless since the flag is invalid. I have moved it outside to avoid any confusion and it's currently not wrapped inside `executeAgainstRunningDapr`.

tests/e2e/standalone/list_test.go
```diff
@@ -0,0 +1,155 @@
...
+		ctx, cancelFunc := context.WithCancel(context.Background())
+		err := cmdDashboard(ctx, "5555")
+		require.NoError(t, err, "dapr dashboard failed")
+
+		output, err := cmdList("")
+		t.Log(output)
+		require.NoError(t, err, "expected no error status on list")
+		require.Equal(t, "No Dapr instances found.\n", output)
+
+		cancelFunc()
```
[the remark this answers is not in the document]
shubham1172: Thanks @mukundansundar, updated. Do you have any thoughts here? https://github.com/dapr/cli/pull/1036#discussion_r942300261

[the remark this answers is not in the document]
mukundansundar: > Thanks @mukundansundar, updated. Do you have any thoughts here? [#1036 (comment)](https://github.com/dapr/cli/pull/1036#discussion_r942300261)

the spawn.Command should propagate errors correctly ... If you can please check that. 


Also I am seeing the windows test failed again ... with port not available (50001), should a different one be used to make it less flaky?

tests/e2e/standalone/init_negative_test.go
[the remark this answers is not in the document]
shubham1172: ping @pravinpushkar

tests/e2e/standalone/list_test.go
shubham1172: As discussed offline, removing this test from Windows now due to its inability to delete .bin when dashboard is running.

tests/e2e/standalone/list_test.go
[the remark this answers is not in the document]
shubham1172: The uninstall test was failing due to dashboard. Please refer to the attached comment for more.

mukundansundar (approved):
Lgtm

PR #1039, opened 2022-08-08
Added `--daprd-flags` option to `dapr run`

### LINKED ISSUE #1038: `dapr run` should have a generic `--daprd-flags` option
## Describe the proposal

Right now, the Dapr CLI has a number of flags to pass options to `daprd`. However,  only the flags that are allowed in the CLI can be passed.

For end users, this means that only flags that have been explicitly allow-listed by the Dapr CLI can be passed (and there are lots of flags missing today).

For developers working on Dapr, it makes it harder to test new CLI flags we're adding.

My proposal is to add a new generic flag that is a string that will be passed to `daprd` as-is, for example `--daprd-flags`

Example:

```
dapr run \
--app-port 4000 \
--daprd-flags "--dapr-public-port 4020 --enable-foo"

## Release Note

RELEASE NOTE: **ADD** `--daprd-flags` CLI option to pass custom flags to daprd

ItalyPaleAle:
# Description

Adds a `--daprd-flags` option to `dapr run` to allow passing arbitrary flags to daprd.

## Issue reference

Fixes #1038

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
   * This is an advanced option and can be learnt about from `dapr run --help`. I don't know if this needs to be put in the docs in addition to that


pravinpushkar:
If it is only for dev/test purpose(not recommended to be used by users) then can we have something like `--daprd-dev-flags`  or something like that . I think it will provide clarity whether user should prefer direct flags or this one. 
@mukundansundar @yaron2

mukundansundar (changes requested):
I do not think we should be providing a catch all flag like `--daprd-flags` in the dapr cli if its main purpose is only dev.
cc @yaron2 WDYT?

ItalyPaleAle:
Bumping this up. @mukundansundar @yaron2 WDYT about this PR?

One thing that I learnt more recently than this PR is that we could mark the flag as "hidden", so it will still available but it won't appear in the list when people do `dapr --help`. It's a sort of "IYKYK" 😃

ItalyPaleAle:
Still pretty much relevant to me

mukundansundar:
closing this PR for now. Can be handled as part of running multiple dapr apps feature if needed.

Issue #257, opened 2020-02-08
Support for podman (or also, removing hard docker dependency)


bureado:
## Describe the proposal

In standalone mode, it'd be nice to have `podman` support in `dapr` (and perhaps `daprd`) at least for `dapr init` and basic `dapr run` use cases. I don't have `docker` in my system.

Even if I `alias docker=podman` in my system, the `github.com/docker/docker/client` library is used in `dapr/cli` which makes `dapr init` fail.

If I remove the `isDockerInstalled()` functionality from `standalone.go` and nix the dependency, and then also change `RunCmdAndWait()` instances of `docker` for `podman`, I'm able to `dapr init` with some hacks (see #255):

```
$ podman ps | grep dapr
231a26a31000  docker.io/library/redis:latest   redis-server          4 minutes ago  Up 4 minutes ago  0.0.0.0:6379->6379/tcp    dapr_redis
f49eb9753492  docker.io/daprio/dapr:latest                           4 minutes ago  Up 4 minutes ago  0.0.0.0:50005->50005/tcp  dapr_placement
```

Those `conmon` processes are running unprivileged, `dapr run --app-id foo -- cat` starts up successfully and I'm able to run `samples/1.hello-world`. Things I haven't tested include binding to a particular network. I also haven't looked into any `dapr/dapr` hard dependencies on `docker`.

yaron2:
`daprd` (`dapr/dapr`) doesn't have any dependencies on `docker`.

The only reason why there's a Docker dependency for the CLI is to install `Redis` as a default development statestore/pubsub for the developer, so they only need to focus on writing code.

There's also an optional placement service for actors, that we install as a Docker container but that could just as well run as a binary.

In effect, Docker just makes setting up a consistent dev environment very easy, and takes care of lifecycle/restarts/versioning etc that would otherwise involve a lot of boilerplate logic to do.

bureado:
Got it. [Here are the rough changes](https://github.com/dapr/cli/compare/master...bureado:jmp-podman-blunt?expand=1) I made to use `podman` instead of `docker`. If the `docker` usage in `dapr/cli` is expected to be limited to the scenarios you describe, perhaps the easiest solution here would be to try for `podman` and if not available, fallback to `docker`, perhaps with a flag to select. If it's valuable, let me think of something and I'll send a PR.

yaron2:
Can `podman` run on Windows?

yaron2:
Hmm.. only with `WSL2` it seems.

[...]

PR #948, opened 2022-04-04
feature/podman: Updating dapr cli to allow different container runtime

martin-bucinskas:
# Description

Adding a new argument `--container-runtime` to be passed during `dapr init` and `dapr uninstall` commands.
This allows a user to provide a container runtime environment such as `podman` where in some instances `docker` is unavailable.

This would be used to stand up a repeatable Dapr environment with the placement, zipkin and redis containers in the desired container runtime environment.

The default container runtime is docker, therefore the behavior is unaltered unless we provide podman as the runtime.

Example:

```bash
# Install Dapr using podman as container runtime
$ dapr init --container-runtime podman

# Uninstall Dapr using podman as container runtime
$ dapr uninstall --cotnainer-runtime podman
```

## Issue reference

closes #257

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Extended the documentation

PR #1041, opened 2022-08-09
feature/podman: Updating dapr cli to allow different container runtime

pravinpushkar:
# Description

This is to continue work on - https://github.com/dapr/cli/pull/948

Taken from original PR - 

Adding a new argument --container-runtime to be passed during dapr init and dapr uninstall commands.
This allows a user to provide a container runtime environment such as podman where in some instances docker is unavailable.

This would be used to stand up a repeatable Dapr environment with the placement, zipkin and redis containers in the desired container runtime environment.

The default container runtime is docker, therefore the behavior is unaltered unless we provide podman as the runtime.

Example:

```
# Install Dapr using podman as container runtime
$ dapr init --container-runtime podman

# Uninstall Dapr using podman as container runtime
$ dapr uninstall --cotnainer-runtime podman
```

## Issue reference

Please reference the issue this PR will close: #257 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


mukundansundar:
@pravinpushkar Please fix failures .. Also enable e2e to run on macOS-latest with `brew install podman` .

pravinpushkar:
@mukundansundar 
With podman also, sometime `podman machine start` gets stuck. But with re-run the tests couple of times, it works.
RN in the current PR I have configured the macos tests to use podman. As you can see all tests except uninstall is passing.
I am facing a peculiar issue with uninstall whereas it is passing locally but failing in github action.

mukundansundar:
> @mukundansundar With podman also, sometime `podman machine start` gets stuck. But with re-run the tests couple of times, it works. RN in the current PR I have configured the macos tests to use podman. As you can see all tests except uninstall is passing. I am facing a peculiar issue with uninstall whereas it is passing locally but failing in github action.

Could this be the issue ?
```
$ docker rmi -f daprio/dapr 
Error: No such image: daprio/dapr
$ # Exits without error 

$ podman rmi -f daprio/dapr 
Error: daprio/dapr: image not known
$ # Exits with error 
```

pravinpushkar:
> > @mukundansundar With podman also, sometime `podman machine start` gets stuck. But with re-run the tests couple of times, it works. RN in the current PR I have configured the macos tests to use podman. As you can see all tests except uninstall is passing. I am facing a peculiar issue with uninstall whereas it is passing locally but failing in github action.
> 
> Could this be the issue ?
> 
> ```
> $ docker rmi -f daprio/dapr 
> Error: No such image: daprio/dapr
> $ # Exits without error 
> 
> $ podman rmi -f daprio/dapr 
> Error: daprio/dapr: image not known
> $ # Exits with error 
> ```
> 
> @pravinpushkar

Actually the issue was we are using `import dockerClient "github.com/docker/docker/client"` in tests for verifying the containers and it expects api forwarding to listen on `/var/run/docker.sock` but with podman api forwarding listens on `/Users/runner/.local/share/containers/podman/machine/podman-machine-default/podman.sock`. There are two approach I can think of -
1.  There is a utility that comes with podman `podman-mac-helper install` which makes api forwarding to listen same as docker.
2. Using slim mode for macos tests.

I have made the changes using 1st approach i.e. `podman-mac-helper install`. 
@mukundansundar

.github/workflows/self_hosted_e2e.yaml
```diff
@@ -61,14 +61,20 @@ jobs:
...
-        if: matrix.os == 'macos-10.15'
+      - name: Install podman - MacOS
+        if: matrix.os == 'macos-latest'
         run: |
-          brew install --cask docker
-          sudo /Applications/Docker.app/Contents/MacOS/Docker --unattended --install-privileged-components
-          open -a /Applications/Docker.app --args --unattended --accept-license
-          while ! /Applications/Docker.app/Contents/Resources/bin/docker info &>/dev/null; do sleep 1; done
-          docker --version
+          sudo rm -rf `brew --cache`
```
mukundansundar, line 67: just wondering if we should also run podman tests in linux/windows?
@pravinpushkar @shubham1172 thoughts?

pkg/standalone/docker.go
```diff
@@ -25,7 +25,8 @@ import (
 )
 
 func runDockerLoad(in io.Reader) error {
```
mukundansundar, line 27:
```suggestion
func loadContainerFromReader(in io.Reader) error {
```

pkg/standalone/docker.go
```diff
@@ -122,10 +124,11 @@ func parseDockerError(component string, err error) error {
 }
```
mukundansundar, line 124: do these error conditions remain the same? if so can we rename the function to parseContainerRuntimeError or something like that?

pkg/standalone/docker.go
mukundansundar: Also change the following function name from `loadDocker` to `loadContainer`?

pkg/standalone/docker.go
```diff
@@ -25,7 +25,8 @@ import (
 )
```
mukundansundar, line 25: Change file name to `container.go`?

pkg/standalone/standalone.go
```diff
@@ -145,15 +145,16 @@ func isBinaryInstallationRequired(binaryFilePrefix, installDir string) (bool, er
...
 	var err error
+	utils.SetContainerRuntime(containerRuntime)
 	var bundleDet bundleDetails
 	fromDir = strings.TrimSpace(fromDir)
 	// AirGap init flow is true when fromDir var is set i.e. --from-dir flag has value.
 	setAirGapInit(fromDir)
 	if !slimMode {
 		// If --slim installation is not requested, check if docker is installed.
-		dockerInstalled := utils.IsDockerInstalled()
+		dockerInstalled := utils.IsDockerInstalled() || utils.IsPodmanInstalled()
```
mukundansundar, line 157:
```suggestion
		conatinerRuntimeAvailable := utils.IsDockerInstalled() || utils.IsPodmanInstalled()
```

pkg/standalone/uninstall.go
```diff
@@ -97,8 +98,9 @@ func Uninstall(uninstallAll bool, dockerNetwork string) error {
 		print.WarningStatusEvent(os.Stdout, "WARNING: could not delete dapr bin dir: %s", daprBinDir)
 	}
 
+	utils.SetContainerRuntime(containerRuntime)
 	dockerInstalled := false
-	dockerInstalled = utils.IsDockerInstalled()
+	dockerInstalled = utils.IsDockerInstalled() || utils.IsPodmanInstalled()
```
mukundansundar, line 103: same as above

utils/utils.go
```diff
@@ -35,6 +35,35 @@ import (
...
 )
 
+type ContainerRuntime string
+
+const (
+	DOCKER ContainerRuntime = "docker"
+	PODMAN ContainerRuntime = "podman"
+)
+
+var selectedContainerRuntime = DOCKER
```
mukundansundar, line 45: can we make these changes without a global var?

utils/utils.go
```diff
@@ -35,6 +35,35 @@ import (
...
+
+func SetContainerRuntime(containerRuntime string) {
+	if containerRuntime == string(PODMAN) {
+		selectedContainerRuntime = PODMAN
+	} else {
+		selectedContainerRuntime = DOCKER
+	}
```
mukundansundar, line 52: `else` not needed ... should always default to docker

mukundansundar (commented):
@pravinpushkar added some comments ...

.github/workflows/self_hosted_e2e.yaml
shubham1172: We can run it for Windows to test the full suite (instead of just slim).

mukundansundar: I still believe running slim is also good ... we could matrix on both slim and full installs ...

pravinpushkar: > just wondering if we should also run podman tests in linux/windows? @pravinpushkar @shubham1172 thoughts?

Let me check if we can run windows full suite using podman.

> we could matrix on both slim and full installs

If we are able to run full suite the do we need to run for slim mode separately ?

mukundansundar: Would like to make sure that slim install works with the limitations that we say it works with. We do have to add actor e2e but still have E2E using slim install is something that should do IMO.

shubham1172: Yes, slim mode has some extra tests, so it will be good to test both.

pkg/standalone/docker.go
pravinpushkar: Error code remains same see here - https://docs.podman.io/en/latest/markdown/podman-remote.1.html#exit-status

.github/workflows/self_hosted_e2e.yaml
pravinpushkar: Enabled slim and complete install matrix for linux and macos.

pravinpushkar: @mukundansundar @shubham1172 Tried installing the podman on windows but could not make it work, see install podman step in windows workflow here - https://github.com/pravinpushkar/cli/runs/7957959381?check_suite_focus=true

mukundansundar (approved):
lgtm

VinuelMancent:
Is there any update to implement this on dapr for windows as well?

pravinpushkar:
> Is there any update to implement this on dapr for windows as well?

Are you facing any issue while running dapr with Podman on Windows ? 
I believe if Podman can be installed on windows then dapr should work as expected. Kindly check Podman docs for installation on windows.

Issue #1098, opened 2022-10-11
Validate and limit container runtime to only podman and docker


shubham1172:
## Describe the proposal

Today any arbitrary container runtime can be passed to Dapr CLI. This proposal is to limit its value to the two supported container runtime today, `podman` and `docker`.

## Release Note

RELEASE NOTE: **ADD** limitation for container runtime in `dapr init` and `dapr uninstall`

shubham1172:
Closed by #1099

PR #1099, opened 2022-10-11
Validate and limit container runtime configuration in init and uninstall commands

shubham1172:
# Description

This PR limits the container runtime configuration to `docker` and `podman` only. 

## Issue reference

Please reference the issue this PR will close: #1098

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/tests/e2e/standalone/commands.go
+++ b/tests/e2e/standalone/commands.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/dapr/cli/tests/e2e/common"
 	"github.com/dapr/cli/tests/e2e/spawn"
+	"github.com/dapr/cli/utils"
 )
 
 // cmdDashboard runs the Dapr dashboard and blocks until it is started.
@@ -101,7 +102,7 @@ func cmdPublish(appId, pubsub, topic, unixDomainSocket string, args ...string) (
 
 // cmdRun runs a Dapr instance and returns the command output and error.
 func cmdRun(unixDomainSocket string, args ...string) (string, error) {
-	runArgs := append([]string{"run"})
+	runArgs := []string{"run"}
 
 	if unixDomainSocket != "" {
 		runArgs = append(runArgs, "--unix-domain-socket", unixDomainSocket)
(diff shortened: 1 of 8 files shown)


pravinpushkar (approved):
lgtm

tests/e2e/standalone/uninstall_test.go
```diff
@@ -30,21 +30,29 @@ import (
...
-	require.NoDirExists(t, daprPath, "Directory %s does not exist", daprPath)
+		// verify that .dapr directory does not exist.
+		homeDir, err := os.UserHomeDir()
+		require.NoError(t, err, "failed to get user home directory")
 
-	verifyNoContainers(t)
+		daprPath := filepath.Join(homeDir, ".dapr")
+		require.NoDirExists(t, daprPath, "Directory %s does not exist", daprPath)
+
+		verifyNoContainers(t)
```
mukundansundar, line 54: @shubham1172 In line 61, do not skip the test (it is showing the whole test as skipped.) just log that you are skipping and return.

tests/e2e/standalone/init_test.go
```diff
@@ -62,6 +62,15 @@ func TestStandaloneInit(t *testing.T) {
...
+	t.Run("init should error out if container runtime is not valid", func(t *testing.T) {
+		// Ensure a clean environment
+		must(t, cmdUninstall, "failed to uninstall Dapr")
+
+		output, err := cmdInit(daprRuntimeVersion, "--container-runtime", "invalid")
+		require.Error(t, err, "expected error if container runtime is invalid")
+		require.Contains(t, output, "Invalid container runtime")
+	})
+
 	t.Run("init", func(t *testing.T) {
```
mukundansundar, line 74: @shubham1172 same for line 121 .... do not skip .... just return ... since without that verify binaries and verify configs will not be run for slim init ...

mukundansundar (changes requested):

tests/e2e/standalone/init_test.go
shubham1172: Good catch!

mukundansundar (approved):

tests/e2e/standalone/commands.go
```diff
@@ -119,12 +120,19 @@ func cmdStop(appId string, args ...string) (string, error) {
 }
 
 // cmdUninstall uninstalls Dapr with --all flag and returns the command output and error.
-func cmdUninstall() (string, error) {
+func cmdUninstall(args ...string) (string, error) {
+	uninstallArgs := []string{"uninstall", "--all"}
```
DeepanshuA, line 124: Can we pre-allocate the size for uninstallArgs? I think we should be able to and if un-used due to if-else branches, we can keep a counter `n` and return uninstallArgs[:n]. As far as I understand, append further will not have O(1) time complexity: https://medium.com/vendasta/golang-the-time-complexity-of-append-2177dcfb6bad

mukundansundar: I agree that this should be done.... 
But I belive this can be done later in master.
This is specifically test code...

shubham1172: Since the performant benefit won't be much for this scenario, I would lean towards readability/maintainability over making tiny performance improvement in this e2e test. Please let me know your thoughts!

DeepanshuA: Though I don't agree with the readability part, but I agree with being in e2e test and number of params being small, so will go ahead and approve and merge

DeepanshuA (approved):

Issue #1117, opened 2022-10-20
❌  Failed to upgrade Dapr: release name is invalid:


isayden:
## Expected Behavior

Successfully upgraded version through cli

## Actual Behavior

# dapr --version
CLI version: 1.9.1
Runtime version: 1.4.3
An error was reported when executing this upgrade command.

#dapr upgrade --runtime-version 1.8.5   -k
ℹ️  Container images will be pulled from Docker Hub
ℹ️  Dapr control plane version 1.6.0 detected in namespace dapr-system
ℹ️  Starting upgrade...
❌  Failed to upgrade Dapr: release name is invalid:

## Steps to Reproduce the Problem

Use the following command
#dapr upgrade --runtime-version 1.8.5   -k

isayden:
It was installed too long ago. The use 
#dapr init -k
I didn't modify the release name somewhere，
dapr-cli 1.9 I just upgraded because I want to upgrade the version of dapr
![image](https://user-images.githubusercontent.com/89994987/196915715-479c77ad-4770-4238-aa66-c0ffffdc1f45.png)

yaoyaoio:
because  in `helm list`  not found dapr 
dapr cli upgrades dapr clusters that already exist in helm

shubham1172:
@isayden can you please also share the output of `helm list -A`?

isayden:
@shubham1172  sure  
![image](https://user-images.githubusercontent.com/89994987/197097666-9b9ac8d3-2446-4240-8811-984ccb565a1e.png)

yaoyaoio:
@shubham1172
dapr upgrate calls the helm.NewList() get about releases list.
function using filters dapr with status = deployed and status = failed
if status !  = deployed | failed .  return empty releases list
@isayden  he is my firend
his  dapr cluster in helm status = pending_upgrade.  so using `helm list --all -n dapr-system` not using `helm list -A` or `helm list -n dapr-system`.

yaoyaoio:
I thinks dapr status = (deployed | failed)  in helm 。To meet the upgrade requirements

isayden:
![image](https://user-images.githubusercontent.com/89994987/197100613-0d58bd9e-1f66-4cae-8378-f1db94117f8f.png)

yaoyaoio:
I think the user needs to be alerted

https://github.com/dapr/cli/pull/1118

PR #1118, opened 2022-10-20
In a certain scenario

yaoyaoio:
# Description

Dapr is installed in kubernetes clsuter.
But it's not in `helm list`
Because using dapr cli upgrade will report an Failed
So I think the user needs to be alerted.
## Issue reference

look issue: https://github.com/dapr/cli/issues/1117

Please reference the issue this PR will close: #_[1117]_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


hueifeng:
@yaoyaoio please fix the DCO.

yaoyaoio:
@hueifeng hi good evening
I handled the DCO and you have a look

hueifeng (approved):

pkg/kubernetes/common.go
```diff
@@ -51,6 +51,9 @@ func GetDaprHelmChartName(helmConf *helm.Configuration) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	if len(releases) == 0 {
+		return "", errors.New("dapr is not installed in your helm")
```
pravinpushkar, line 55: nit: Just thinking of better error message here -
"Could not find release name dapr in the your helm releases" @shubham1172 WDYT?

shubham1172: Yes, we should inform users transparently. +1

pravinpushkar (changes requested):
Could you please correct the error message.

pravinpushkar (approved):
@mukundansundar This change looks fine to me and we can keep this check.
I pushed a commit to change the error message to make it more informative.

PR #1126, opened 2022-11-04
update golang version

mukundansundar:
Signed-off-by: Mukundan Sundararajan <65565396+mukundansundar@users.noreply.github.com>

# Description

update golang version

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/dapr_cli.yaml
+++ b/.github/workflows/dapr_cli.yaml
@@ -29,7 +29,7 @@ jobs:
     name: Build ${{ matrix.target_os }}_${{ matrix.target_arch }} binaries
     runs-on: ${{ matrix.os }}
     env:
-      GOVER: 1.19.2
+      GOVER: "^1.19.3"
       GOLANG_CI_LINT_VER: v1.49.0
       GOOS: ${{ matrix.target_os }}
       GOARCH: ${{ matrix.target_arch }}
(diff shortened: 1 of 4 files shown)


.github/workflows/dapr_cli.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
     name: Build ${{ matrix.target_os }}_${{ matrix.target_arch }} binaries
     runs-on: ${{ matrix.os }}
     env:
-      GOVER: 1.19.2
+      GOVER: "^1.19.3"
```
shubham1172, line 32: Why this change, shouldn't we pin to a known version? This will take up all minor and patch upgrades automatically. I propose we stick to pinning a version to avoid any unforeseen problems.

shubham1172 (changes requested):

.github/workflows/dapr_cli.yaml
mukundansundar: see https://github.com/npm/node-semver#caret-ranges-123-025-004

pravinpushkar (approved):

.github/workflows/dapr_cli.yaml
shubham1172: Yes, so this means >=1.19.3 and <2.0.0-0, right?

Issue #1130, opened 2022-11-17
Fix self-hosted e2e tests on macOS


shubham1172:
## Description

Our self-hosted e2e test suite has broken on macOS due to upgrade to macOS-12 runner. The new runner does not include `helm`, and we fail to fetch the latest runtime version. Example, https://github.com/dapr/cli/actions/runs/3485014051/jobs/5831664262

![image](https://user-images.githubusercontent.com/22132944/202372832-6cb34355-4873-4d76-8cc7-8889eb296736.png)

Reference issue https://github.com/actions/runner-images/issues/6384

## Potential fix

1. Remove dependency from Helm
1. Install Helm before using it

Should we also consider pinning our runner versions to not run into issues like these?

## Release Note

RELEASE NOTE: **FIX** self-hosted e2e tests on macOS-12

PR #1131, opened 2022-11-17
Remove helm dependency from e2e test suite

shubham1172:

# Description

Remove dependency from Helm since it's no longer available in the latest macOS.

Uses bash magic to get version for both dashboard and runtime from GitHub releases.

## Issue reference

Please reference the issue this PR will close: #1130 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/self_hosted_e2e.yaml
+++ b/.github/workflows/self_hosted_e2e.yaml
@@ -1,3 +1,5 @@
+--- # ------------------------------------------------------------
+
 # ------------------------------------------------------------
 # Copyright 2021 The Dapr Authors
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -9,8 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# ------------------------------------------------------------
-
 name: E2E - Self-hosted
 
 on:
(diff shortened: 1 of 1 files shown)


mukundansundar (changes requested):
can this be done for the dashboard check as well ?

shubham1172:
> can this be done for the dashboard version as well ?

Yes indeed, good idea. Note that it was earlier taking the version from 1.9.0's helm charts. Now it's pointing to the latest.

mukundansundar:
> > can this be done for the dashboard version as well ?
> 
> Yes indeed, good idea. Note that it was earlier taking the version from 1.9.0's helm charts. Now it's pointing to the latest.

Good point. Is there any scenario where the latest version will not be in the latest Dapr runtime version also? 

Also what happens with RC releases with this command?

shubham1172:
> Good point. Is there any scenario where the latest version will not be in the latest Dapr runtime version also?

It is safe to assume that we will always make that change. It's a part of our endgame check-list too.

> Also what happens with RC releases with this command?

The sorting is done correctly even with RCs using sed magic.

mukundansundar:
it would really be nice if podman becomes a default with macOS :)

pravinpushkar (approved):

shubham1172:
@mukundansundar can we merge it?

mukundansundar (approved):

Issue #1105, opened 2022-10-14
Add E2E to test plain dapr init


mukundansundar:
## Describe the proposal
Add E2E to test plain dapr init instead of `dapr init --runtime-version`

## Release Note

RELEASE NOTE:

PR #1135, opened 2022-11-23
dapr init tests without runtime-version flag

pravinpushkar:

## Issue reference

Please reference the issue this PR will close: #1105


shubham1172 (dismissed):

tests/e2e/standalone/init_test.go
```diff
@@ -110,6 +127,26 @@ func TestStandaloneInit(t *testing.T) {
...
+
+		homeDir, err := os.UserHomeDir()
+		require.NoError(t, err, "failed to get user home directory")
+
+		daprPath := filepath.Join(homeDir, ".dapr")
+		require.DirExists(t, daprPath, "Directory %s does not exist", daprPath)
+
+		verifyContainers(t, daprRuntimeVersion)
+		verifyBinaries(t, daprPath, daprRuntimeVersion, daprDashboardVersion)
+		verifyConfigs(t, daprPath)
```
mukundansundar, line 148: This will fail whenever there is a new patch release, unless the DAPR_RUNTIME_VERSION field in workflow file is updated ...

mukundansundar (changes requested):
We need to either modify tests such that on each release we can get the latest version injected into the tests

tests/e2e/standalone/init_test.go
shubham1172: Don't we query the latest version in the workflow file though, with the cURL magic?

mukundansundar: not for normal scheduled runs ... only for PRs... That is done specifically such that for release branches we always have a specific version to target and not latest version ...

shubham1172: Got it, yes. Then we should check "greater-than-equal" for versions in the test, WDYT?

mukundansundar: `init` as of now is dependent on latest release in GH. There are two scenarios here: 
- without specifying a runtime version, it should always get the latest version. we can follow the `cURL` magic and injection into env there? 
- when runtime version is specified, it has to be done in workflow file and that version should be checked. 
IMO we should not relax the test assertion.

shubham1172: Just to clarify, the "greater-than-equal" was only for `init` without runtime version.

Yes, in the workflow, we can set additional variables to always point to the latest version. The scheduled runs can set the variables, but still be pinned to the release specified.

pravinpushkar: I see, let me check this.

.github/workflows/self_hosted_e2e.yaml
```diff
@@ -42,6 +42,8 @@ jobs:
       ARCHIVE_OUTDIR: dist/archives
       DAPR_RUNTIME_VERSION: "1.9.4"
       DAPR_DASHBOARD_VERSION: 0.11.0
+      DAPR_LATEST_RUNTIME_VERSION:
```
shubham1172, line 45: For consistency
```suggestion
      DAPR_RUNTIME_LATEST_VERSION:
```

.github/workflows/kind_e2e.yaml
```diff
@@ -133,14 +133,14 @@ jobs:
       if: github.base_ref == 'master'
       run: |
         helm repo add dapr https://dapr.github.io/helm-charts/ && helm repo update && export RUNTIME_VERSION=$(helm search repo dapr/dapr --devel --versions | awk '/dapr\/dapr/ {print $3; exit}' )
```
shubham1172, line 135: Unrelated, but shall we also change this to use cURL like self hosted e2e?

.github/workflows/kind_e2e.yaml
```diff
@@ -51,8 +51,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GOVER: 1.19.3
-      DAPR_RUNTIME_VERSION: 1.9.0
-      DAPR_DASHBOARD_VERSION: 0.11.0
+      DAPR_RUNTIME_PINNED_VERSION: 1.9.0
```
shubham1172, line 54: Shall we update this to 1.9.4?

.github/workflows/self_hosted_e2e.yaml
```diff
@@ -40,8 +40,10 @@ jobs:
       GOARCH: ${{ matrix.target_arch }}
       GOPROXY: https://proxy.golang.org
       ARCHIVE_OUTDIR: dist/archives
-      DAPR_RUNTIME_VERSION: "1.9.4"
-      DAPR_DASHBOARD_VERSION: 0.11.0
+      DAPR_RUNTIME_PINNED_VERSION: "1.9.0"
```
shubham1172, line 43: Same, 1.9.4?

.github/workflows/self_hosted_e2e.yaml
```diff
@@ -77,25 +79,33 @@ jobs:
           podman machine start
           echo "CONTAINER_RUNTIME=podman" >> $GITHUB_ENV
       - name: Determine latest Dapr Runtime version including Pre-releases
-        if: github.base_ref == 'master'
         run: |
           export RUNTIME_VERSION=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" https://api.github.com/repos/dapr/dapr/releases | grep tag_name | awk -F':' '{print $2}' | tr -d '", ' | sed '/-/! s/$/_/' | sort -V | sed 's/_$//' | tr -d 'v' | tail -1)
+          echo "DAPR_RUNTIME_LATEST_VERSION=$RUNTIME_VERSION" >> $GITHUB_ENV
           if [[ -z "$RUNTIME_VERSION" ]]; then
-            echo "Could not fetch the latest Dapr Runtime version. Using default version $DAPR_RUNTIME_VERSION"
+            echo "Could not fetch the latest Dapr Runtime version. Using default version $DAPR_RUNTIME_PINNED_VERSION"
```
shubham1172, line 86: You also need to set `DAPR_RUNTIME_LATEST_VERSION` to `DAPR_RUNTIME_PINNED_VERSION`, right?

tests/e2e/common/common.go
```diff
@@ -73,18 +73,23 @@ type TestCase struct {
 
 // GetVersionsFromEnv will return values from required environment variables,
 // if environment variables are not set it fails the test.
-func GetVersionsFromEnv(t *testing.T) (string, string) {
+func GetVersionsFromEnv(t *testing.T, latest bool) (string, string) {
```
shubham1172, line 76: Maybe we can create another function `GetLatestVersionsFromEnv` instead. If not, let's update the docs comment to mention what it does.

.github/workflows/kind_e2e.yaml
pravinpushkar: IMO we can leave this as it since this looks cleaner than cURL magic, but I am fine to change as well. @mukundansundar

.github/workflows/kind_e2e.yaml
pravinpushkar: Yes, I did this for testing. Will change it back to 1.9.4

tests/e2e/common/common.go
pravinpushkar: Will update the doc

.github/workflows/self_hosted_e2e.yaml
pravinpushkar: No, not needed. These two will be used separately to track the pinned(directly set via `env` in workflow files) and latest versions.

.github/workflows/kind_e2e.yaml
mukundansundar: if we can make it consistent lets do it. But in a separate PR please.

.github/workflows/self_hosted_e2e.yaml
shubham1172: What will this imply then "Using default version $DAPR_RUNTIME_PINNED_VERSION"?
The idea here was that if the cURL request fails due to GitHub API being down or throttling, `$DAPR_RUNTIME_LATEST_VERSION` will still contain a sane value instead of empty.

mukundansundar (changes requested):
@pravinpushkar  The `dapr init -k` command should also be tested not only self hosted mode command without `--runtime-version` flag.

pravinpushkar:
> @pravinpushkar The `dapr init -k` command should also be tested not only self hosted mode command without `--runtime-version` flag.

ok.. Then let me add for that also.

.github/workflows/self_hosted_e2e.yaml
pravinpushkar: Corrected.

.github/workflows/kind_e2e.yaml
shubham1172: This is done via #1140

mukundansundar:
@pravinpushkar All E2E tests are failing can you take a look ?

shubham1172 (approved):

.github/workflows/kind_e2e.yaml
```diff
@@ -130,25 +132,35 @@ jobs:
       run: |
         docker image prune -a -f
     - name: Determine latest Dapr Runtime version including Pre-releases
-      if: github.base_ref == 'master'
```
mukundansundar, line 133: By removing this, this will always be run, for running against release branches, it will always use the latest release version and not the pinned version.

mukundansundar (changes requested):
@pravinpushkar changes need to be made to make sure that when workflow is run against a release branch, only the pinned version is used, otherwise it can query the latest version.

.github/workflows/kind_e2e.yaml
pravinpushkar: @mukundansundar The behavior is unchanged from the previous. This check is not removed but it has been moved down by few lines. 

The change here is-  
- it will  always calculate the "latest version" (to test dapr init without `runtime-version`) for all runs(schedule/prs)
- Run all other tests with "Pinned version" for all schedule runs.
- Run all other tests with "latest version" for all prs.

mukundansundar (approved):

Issue #1125, opened 2022-11-04
Add `--image-variant` flag to mtls renew-certificate command 


pravinpushkar:
##Description

Renew certificate command will fail if the control plane is using mariner variant of the docker image.

➜  cli git:(master) ✗ dapr status -k
  NAME                   NAMESPACE    HEALTHY  STATUS   REPLICAS  VERSION        AGE  CREATED
  dapr-operator          dapr-system  True     Running  1         1.8.0-mariner  7m   2022-11-04 14:13.16
  dapr-dashboard         dapr-system  True     Running  1         0.10.0         7m   2022-11-04 14:13.16
  dapr-sentry            dapr-system  True     Running  1         1.8.0-mariner  7m   2022-11-04 14:13.16
  dapr-placement-server  dapr-system  True     Running  1         1.8.0-mariner  7m   2022-11-04 14:13.18
  dapr-sidecar-injector  dapr-system  True     Running  1         1.8.0-mariner  7m   2022-11-04 14:13.16

➜  cli git:(master) ✗ dapr mtls renew-certificate -k --valid-until 5
⌛  Starting certificate rotation
ℹ️  generating fresh certificates
ℹ️  Updating certifcates in your Kubernetes cluster
ℹ️  Dapr control plane version 1.8.0-mariner detected in namespace dapr-system
❌  certificate rotation failed: chart "dapr" version "1.8.0-mariner" not found in https://dapr.github.io/helm-charts repository

pravinpushkar:
Another question - Should we actually show `1.8.0-mariner` in the `version` cloumn for `dapr status` ?

mukundansundar:
> Another question - Should we actually show `1.8.0-mariner` in the `version` cloumn for `dapr status` ?

I think showing 1.8.0-mariner is ok in the version field. Unless you want to add a variant field which will either be mariner or empty for a deployment. 
I think having it in version is ok since it is the actual docker container tag ...

PR #1139, opened 2022-11-25
mariner renew-certificate fix

pravinpushkar:

# Description

In case, where control plane has mariner variant images, the renew-cert will fail while getting the helm chart for current version(1.9.0-mariner) in this case. 
Here, the version is read from `Status` command which appends `mariner` to the actual version. So, we need extract the version from complete tag.

## Issue reference

Please reference the issue this PR will close: #1125 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/renew_certificate.go
+++ b/cmd/renew_certificate.go
@@ -39,7 +39,7 @@ func RenewCertificateCmd() *cobra.Command {
 	command := &cobra.Command{
 		Use:     "renew-certificate",
 		Aliases: []string{"renew-cert", "rc"},
-		Short:   "Rotates Dapr root certificate of your kubernetes cluster",
+		Short:   "Rotates the Dapr root certificate on your Kubernetes cluster",
 
 		Example: `
 # Generates new root and issuer certificates for kubernetes cluster
--- a/cmd/renew_certificate.go
+++ b/cmd/renew_certificate.go
@@ -38,7 +38,7 @@ var (
 func RenewCertificateCmd() *cobra.Command {
 	command := &cobra.Command{
 		Use:     "renew-certificate",
-		Aliases: []string{"renew-cert", "rc"},
+		Aliases: []string{"renew-cert", "rnc"},
 		Short:   "Rotates the Dapr root certificate on your Kubernetes cluster",
 
 		Example: `
@@ -50,6 +50,13 @@ dapr mtls renew-certificate -k --private-key myprivatekey.key --valid-until <no 
 
 # Rotates certificate of your kubernetes cluster with provided ca.cert, issuer.crt and issuer.key file path
 dapr mtls renew-certificate -k --ca-root-certificate <root.pem> --issuer-private-key <issuer.key> --issuer-public-certificate <issuer.pem> --restart
+
+# Generates new root and issuer certificates for kubernetes cluster with provided image variant
+dapr mtls renew-certificate -k --valid-until <no of days> --image-variant mariner --restart
+
+# Use alias to renew certificate command
+dapr mtls rnc -k --valid-until <no of days> --restart
+dapr mtls renew-cert -k --valid-until <no of days> --restart
 
 # See more at: https://docs.dapr.io/getting-started/
 `,
--- a/cmd/renew_certificate.go
+++ b/cmd/renew_certificate.go
@@ -37,8 +37,9 @@ var (
 
 func RenewCertificateCmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:   "renew-certificate",
-		Short: "Rotates Dapr root certificate of your kubernetes cluster",
+		Use:     "renew-certificate",
+		Aliases: []string{"renew-cert", "rc"},
+		Short:   "Rotates Dapr root certificate of your kubernetes cluster",
 
 		Example: `
 # Generates new root and issuer certificates for kubernetes cluster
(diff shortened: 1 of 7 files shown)


cmd/renew_certificate.go
```diff
@@ -37,8 +37,9 @@ var (
 
 func RenewCertificateCmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:   "renew-certificate",
-		Short: "Rotates Dapr root certificate of your kubernetes cluster",
+		Use:     "renew-certificate",
+		Aliases: []string{"renew-cert", "rc"},
+		Short:   "Rotates Dapr root certificate of your kubernetes cluster",
```
shubham1172, line 42:
```suggestion
		Short:   "Rotates the Dapr root certificate of your kubernetes cluster",
```

shubham1172, line 42:
```suggestion
		Short:   "Rotates the Dapr root certificate on your Kubernetes cluster",
```

utils/utils.go
```diff
@@ -310,3 +310,13 @@ func GetVariantVersion(version, imageVariant string) string {
 	}
 	return fmt.Sprintf("%s-%s", version, imageVariant)
 }
+
+// Returns image version and variant.
+// Expected imageTag format: <version>-<variant>, i.e. 1.0.0-mariner.
+func GetVersionAndImageVariant(imageTag string) (string, string) {
```
shubham1172, line 316: What if the version is 1.9.0-rc.2-mariner?

shubham1172 (changes requested):

tests/e2e/common/common.go
```diff
@@ -437,7 +437,17 @@ func GenerateNewCertAndRenew(details VersionDetails, opts TestOptions) func(t *t
 		err := exportCurrentCertificate(daprPath)
 		require.NoError(t, err, "expected no error on certificate exporting")
 
-		output, err := spawn.Command(daprPath, "mtls", "renew-certificate", "-k", "--valid-until", "20", "--restart")
+		args := []string{
+			"mtls",
+			"renew-certificate",
+			"-k",
+			"--valid-until", "20",
```
shubham1172, line 444: Should this be configurable?

utils/utils.go
pravinpushkar: yes, Thanks for catching. I think we should use `LastIndex` here also, right ?

shubham1172: Yes, that should work.

utils/utils_test.go
```diff
@@ -117,3 +121,45 @@ func TestContains(t *testing.T) {
...
+			expectedImageVariant: "mariner",
+		},
+		{
+			name:                 "image tag contains only version",
+			input:                "1.9.0",
+			expectedVersion:      "1.9.0",
+			expectedImageVariant: "",
+		},
+		{
+			name:                 "image tag contains only version",
```
mukundansundar, line 145:
```suggestion
			name:                 "image tag contains only rc version and variant",
```

utils/utils_test.go
```diff
@@ -117,3 +121,45 @@ func TestContains(t *testing.T) {
...
+			input:                "1.9.0-rc.1-mariner",
+			expectedVersion:      "1.9.0-rc.1",
+			expectedImageVariant: "mariner",
+		},
+		{
+			name:                 "image tag contains only version",
```
shubham1172, line 151:
```suggestion
			name:                 "image tag contains only rc version",
```

shubham1172 (approved):

pkg/kubernetes/renew_certificate.go
```diff
@@ -86,14 +88,23 @@ func parseCertificateFiles(rootCert, issuerCert, issuerKey string) ([]byte, []by
...
 	status, err := GetDaprResourcesStatus()
 	if err != nil {
 		return err
 	}
-	daprVersion := GetDaprVersion(status)
+	daprVersion = GetDaprVersion(status)
 	print.InfoStatusEvent(os.Stdout, "Dapr control plane version %s detected in namespace %s", daprVersion, status[0].Namespace)
 
+	// Get the current version if image variant is provided.
+	if imageVariant != "" {
```
pravinpushkar, line 101: `imageVariant` is just needed to extract the version from this string 1.x.x-mariner and do some validation on that. It is not needed in actual cert rotation logic as this will be handled by this logic - `upgradeClient.ReuseValues = true` below at line 118. It will reuse all previous flags if not passed.

mukundansundar: Can you add that as a comment in the code itself?

cmd/renew_certificate.go
```diff
@@ -37,8 +37,9 @@ var (
 
 func RenewCertificateCmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:   "renew-certificate",
-		Short: "Rotates Dapr root certificate of your kubernetes cluster",
+		Use:     "renew-certificate",
+		Aliases: []string{"renew-cert", "rc"},
```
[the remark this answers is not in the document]
shubham1172: Since it's suffixed with `mtls`, i.e. `dapr mtls rc`, I guess it should be contextual. I am ok with keeping or removing it.

pkg/kubernetes/renew_certificate.go
shubham1172: Unrelated, but I believe image variants should not be passed around as strings. We should create a new type and use that instead.

cmd/renew_certificate.go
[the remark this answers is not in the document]
pravinpushkar: I added it because I have personally felt pain in writing complete `renew-certificate` always while testing. I am in favour of keeping it. 
or we can put "rnc" instead of rc?

pkg/kubernetes/renew_certificate.go
mukundansundar: Agreed

cmd/renew_certificate.go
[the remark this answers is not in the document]
mukundansundar: let's go with rc itself ... 
Can you add usage examples for this with the alias also. 
@msfussell Any issue with having an alias like rc so that we can call `dapr mtls rc` instead of `dapr mtls renew-certificate` ?

mukundansundar (approved):
lgtm

Issue #1148, opened 2022-12-08
Rename `components` directory to `resources` under .dapr directory


pravinpushkar:
## Describe the proposal

The issue #953 layout a proposal for using `resources-path` flag with `dapr run` instead of `components-path` flag and slowly deprecating it.
In the similar context it would make sense to rename `components` directory to `resources` and slowly deprecate `components` directory use. 

Since the `components` directory is heavily used and exists from beginning so, migrating from `components` to `resources` directly should be done in a phased manner.

We can use this issue to discuss different approaches for this issue.

## Release Note

RELEASE NOTE:

pravinpushkar:
@mukundansundar @yaron2 @shubham1172 @msfussell

shubham1172:
Sounds good!
1. When running a `dapr init`, use `resources` instead of `components`.
1. When running `dapr run`, order of precedence can be `resources` > `components`. If `components` is used, a warning log can be added with a link to the GitHub issue with more details on the deprecation.

mukundansundar:
> Sounds good!
> 
> 1. When running a `dapr init`, use `resources` instead of `components`.
> 2. When running `dapr run`, order of precedence can be `resources` > `components`. If `components` is used, a warning log can be added with a link to the GitHub issue with more details on the deprecation.

Agreed. Only when default dir is used, will the deprecation notice be printed i.e. for `~/.dapr/components` dir.
Also when `--components-path` is used, then also a deprecation notice needs to be printed.

shubham1172:
> There might be a preexisting ~/.dapr/components if dapr uninstall --all is not used before dapr init

In this case, shall we warn and copy files to `resources`?

pravinpushkar:
> Out of scope for this issue?

It is being covered in #1012

mukundansundar:
> Based on the above suggestions -
> 
> 1. When User just upgrades his `dapr cli` without uninstalling -> all usage of `dapr run --components-path ""` will start giving deprecation message but continues to work.
> 2. User does `dapr init` after `dapr uninstall` or `dapr uninstall  --all` -> Create `resources` directory always and copy the files from `components` directory if `components` dir exists.
> 3. In all cases `dapr run` without either of the above flags, will default to looking `resources` dir first then `components` dir.

this looks good to me.
@yaron2 thoughts?

mukundansundar:
Have triaged this to be completed in 1.11

PR #1149, opened 2022-12-09
Feat/resources dir

pravinpushkar:
# Description

- this pr changes `dapr init` behaviour to create `resources` dir instead of `components`
- if there is any pre-existing `components` dir, `dapr init` will also moves its contents into `resources` dir
- `dapr run --resources-path` -> uses provided path in through flag
- `dapr run --components-path` -> uses provided path in through flag 
- `dapr run` -> precedence order is resources dir then components dir.

## Issue reference

Please reference the issue this PR will close: #1148 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -377,7 +377,7 @@ func init() {
 	RunCmd.Flags().StringVarP(&logLevel, "log-level", "", "info", "The log verbosity. Valid values are: debug, info, warn, error, fatal, or panic")
 	RunCmd.Flags().IntVarP(&maxConcurrency, "app-max-concurrency", "", -1, "The concurrency level of the application, otherwise is unlimited")
 	RunCmd.Flags().StringVarP(&protocol, "app-protocol", "P", "http", "The protocol (gRPC or HTTP) Dapr uses to talk to the application")
-	RunCmd.Flags().StringVarP(&componentsPath, "components-path", "d", standalone.DefaultComponentsDirPath(), "The path for components directory")
+	RunCmd.Flags().StringVarP(&componentsPath, "components-path", "d", standalone.DefaultResourcesDirPrecedence(), "The path for components directory")
 	RunCmd.Flags().StringVarP(&resourcesPath, "resources-path", "", "", "The path for resources directory")
 	// TODO: Remove below line once the flag is removed in the future releases.
 	// By marking this as deprecated, the flag will be hidden from the help menu, but will continue to work. It will show a warning message when used.
--- a/pkg/standalone/common.go
+++ b/pkg/standalone/common.go
@@ -17,6 +17,8 @@ import (
 	"os"
 	path_filepath "path/filepath"
 	"runtime"
+
+	"github.com/dapr/cli/pkg/print"
 )
 
 const (
@@ -52,6 +54,8 @@ func DefaultResourcesDirPath() string {
 	return path_filepath.Join(defaultDaprDirPath(), defaultResourcesDirName)
 }
 
+// when either `components-path` or `resources-path` flags are not present then preference is given to resources dir and then components dir.
+// TODO: Remove this function and use `DefaultResourcesDirPath` when `--components-path` flag is removed.
 func DefaultResourcesDirPrecedence() string {
 	defaultResourcesDirPath := DefaultResourcesDirPath()
 	if _, err := os.Stat(defaultResourcesDirPath); os.IsNotExist(err) {
@@ -62,4 +66,30 @@ func DefaultResourcesDirPrecedence() string {
 
 func DefaultConfigFilePath() string {
 	return path_filepath.Join(defaultDaprDirPath(), defaultConfigFileName)
+}
+
+// It used to copy existing resources from components dir to resources dir.
+// TODO: Remove this function when `--components-path` flag is removed.
+func moveFilesFromComponentsToResourcesDir(componentsDirPath, resourcesDirPath string) error {
+	if _, err := os.Stat(componentsDirPath); err == nil {
+		files, err := os.ReadDir(componentsDirPath)
+		if err != nil {
+			return err
+		}
+		if len(files) > 0 {
+			print.InfoStatusEvent(os.Stdout, "Moving files from %q to %q", componentsDirPath, resourcesDirPath)
+			for _, file := range files {
+				content, err := os.ReadFile(componentsDirPath + "/" + file.Name())
+				if err != nil {
+					return err
+				}
+				// #nosec G306
+				err = os.WriteFile(resourcesDirPath+"/"+file.Name(), content, 0o644)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
 }
--- a/pkg/standalone/common.go
+++ b/pkg/standalone/common.go
@@ -23,6 +23,7 @@ const (
 	defaultDaprDirName       = ".dapr"
 	defaultDaprBinDirName    = "bin"
 	defaultComponentsDirName = "components"
+	defaultResourcesDirName  = "resources"
 	defaultConfigFileName    = "config.yaml"
 )
 
@@ -45,6 +46,18 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
 
 func DefaultComponentsDirPath() string {
 	return path_filepath.Join(defaultDaprDirPath(), defaultComponentsDirName)
+}
+
+func DefaultResourcesDirPath() string {
+	return path_filepath.Join(defaultDaprDirPath(), defaultResourcesDirName)
+}
+
+func DefaultResourcesDirPrecedence() string {
+	defaultResourcesDirPath := DefaultResourcesDirPath()
+	if _, err := os.Stat(defaultResourcesDirPath); os.IsNotExist(err) {
+		return DefaultComponentsDirPath()
+	}
+	return defaultResourcesDirPath
 }
 
 func DefaultConfigFilePath() string {
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -244,7 +244,7 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 	defer stopSpinning(print.Failure)
 
 	// Make default components directory.
-	err = makeDefaultComponentsDir()
+	err = makeDefaultResourcesDir()
 	if err != nil {
 		return err
 	}
@@ -297,7 +297,8 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		}
 		for _, container := range dockerContainerNames {
 			containerName := utils.CreateContainerName(container, dockerNetwork)
-			ok, err := confirmContainerIsRunningOrExists(containerName, true, runtimeCmd)
+			var ok bool
+			ok, err = confirmContainerIsRunningOrExists(containerName, true, runtimeCmd)
 			if err != nil {
 				return err
 			}
@@ -306,6 +307,11 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 			}
 		}
 		print.InfoStatusEvent(os.Stdout, "Use `%s ps` to check running containers.", runtimeCmd)
+	}
+	// TODO: remove below method when components-path flag is removed.
+	err = moveFilesFromComponentsToResourcesDir(DefaultComponentsDirPath(), DefaultResourcesDirPath())
+	if err != nil {
+		return err
 	}
 	return nil
 }
@@ -659,14 +665,14 @@ func createComponentsAndConfiguration(wg *sync.WaitGroup, errorChan chan<- error
 	var err error
 
 	// Make default components directory.
-	componentsDir := DefaultComponentsDirPath()
+	resourcesDir := DefaultResourcesDirPath()
 
-	err = createRedisPubSub(redisHost, componentsDir)
+	err = createRedisPubSub(redisHost, resourcesDir)
 	if err != nil {
 		errorChan <- fmt.Errorf("error creating redis pubsub component file: %w", err)
 		return
 	}
-	err = createRedisStateStore(redisHost, componentsDir)
+	err = createRedisStateStore(redisHost, resourcesDir)
 	if err != nil {
 		errorChan <- fmt.Errorf("error creating redis statestore component file: %w", err)
 		return
@@ -693,19 +699,19 @@ func createSlimConfiguration(wg *sync.WaitGroup, errorChan chan<- error, info in
 	}
 }
 
-func makeDefaultComponentsDir() error {
+func makeDefaultResourcesDir() error {
 	// Make default components directory.
-	componentsDir := DefaultComponentsDirPath()
+	resourcesDir := DefaultResourcesDirPath()
 	//nolint
-	_, err := os.Stat(componentsDir)
+	_, err := os.Stat(resourcesDir)
 	if os.IsNotExist(err) {
-		errDir := os.MkdirAll(componentsDir, 0o755)
+		errDir := os.MkdirAll(resourcesDir, 0o755)
 		if errDir != nil {
 			return fmt.Errorf("error creating default components folder: %w", errDir)
 		}
 	}
 
-	os.Chmod(componentsDir, 0o777)
+	os.Chmod(resourcesDir, 0o777)
 	return nil
 }
 


tests/e2e/standalone/componentsDir_to_resourcesDir_mig_test.go
```diff
@@ -0,0 +1,130 @@
+//go:build e2e
```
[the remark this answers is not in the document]
pravinpushkar: how about `resources_dir_migration_test.go` ?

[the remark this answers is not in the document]
shubham1172: Can it be `init_migration_test.go`?

pkg/standalone/standalone.go
```diff
@@ -297,7 +297,8 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		}
 		for _, container := range dockerContainerNames {
 			containerName := utils.CreateContainerName(container, dockerNetwork)
-			ok, err := confirmContainerIsRunningOrExists(containerName, true, runtimeCmd)
+			var ok bool
```
akhilac1, line 300: move declaration to outside for loop or use shortcut to declare and init ':='

pkg/standalone/standalone.go
```diff
@@ -307,6 +308,11 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		}
 		print.InfoStatusEvent(os.Stdout, "Use `%s ps` to check running containers.", runtimeCmd)
 	}
+	// TODO: remove below method when components-path flag is removed.
+	err = moveFilesFromComponentsToResourcesDir(DefaultComponentsDirPath(), DefaultResourcesDirPath())
```
akhilac1, line 312: Since this is init, a standard set of files will be setup in the components-path. Deleting the path should be sufficient. is there a need to copy files from old dir to new dir for init?

pkg/standalone/standalone.go
```diff
@@ -693,19 +699,19 @@ func createSlimConfiguration(wg *sync.WaitGroup, errorChan chan<- error, info in
...
-	componentsDir := DefaultComponentsDirPath()
+	resourcesDir := DefaultResourcesDirPath()
 	//nolint
-	_, err := os.Stat(componentsDir)
+	_, err := os.Stat(resourcesDir)
 	if os.IsNotExist(err) {
-		errDir := os.MkdirAll(componentsDir, 0o755)
+		errDir := os.MkdirAll(resourcesDir, 0o755)
 		if errDir != nil {
 			return fmt.Errorf("error creating default components folder: %w", errDir)
```
akhilac1, line 710: Update the error message to indicate resource directory instead of components folder, since components folder is marked for deprecation

tests/e2e/standalone/utils.go
```diff
@@ -142,7 +142,7 @@ func ensureDaprInstallation(t *testing.T) {
 	// Slim mode does not have any components by default.
 	// Install the components required by the tests.
 	if isSlimMode() {
-		err = createSlimComponents(filepath.Join(daprPath, "components"))
+		err = createSlimComponents(filepath.Join(daprPath, "resources"))
```
akhilac1, line 145: use the const declared in common.go instead of string literal for "resources"

akhilac1 (changes requested):

cmd/run.go
```diff
@@ -377,7 +377,7 @@ func init() {
 	RunCmd.Flags().StringVarP(&logLevel, "log-level", "", "info", "The log verbosity. Valid values are: debug, info, warn, error, fatal, or panic")
 	RunCmd.Flags().IntVarP(&maxConcurrency, "app-max-concurrency", "", -1, "The concurrency level of the application, otherwise is unlimited")
 	RunCmd.Flags().StringVarP(&protocol, "app-protocol", "P", "http", "The protocol (gRPC or HTTP) Dapr uses to talk to the application")
-	RunCmd.Flags().StringVarP(&componentsPath, "components-path", "d", standalone.DefaultComponentsDirPath(), "The path for components directory")
+	RunCmd.Flags().StringVarP(&componentsPath, "components-path", "d", standalone.DefaultResourcesDirPrecedence(), "The path for components directory")
 	RunCmd.Flags().StringVarP(&resourcesPath, "resources-path", "", "", "The path for resources directory")
```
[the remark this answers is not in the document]
pravinpushkar: We can not use same shortcut for two flags. We will be using the existing shortcut once we move out completely from `--components-path`

pkg/standalone/common.go
```diff
@@ -47,6 +50,56 @@ func DefaultComponentsDirPath() string {
 	return path_filepath.Join(defaultDaprDirPath(), defaultComponentsDirName)
 }
 
+func DefaultResourcesDirPath() string {
+	return path_filepath.Join(defaultDaprDirPath(), defaultResourcesDirName)
+}
+
+// when either `components-path` or `resources-path` flags are not present then preference is given to resources dir and then components dir.
```
shubham1172, line 57:
```suggestion
// DefaultResourcesDirPrecedence returns the path to the resources directory if it exists, or otherwise the components directory.
```

pkg/standalone/standalone.go
```diff
@@ -244,7 +244,7 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 	defer stopSpinning(print.Failure)
 
 	// Make default components directory.
```
shubham1172, line 246:
```suggestion
	// Make default resources directory.
```

pkg/standalone/common.go
```diff
@@ -47,6 +50,56 @@ func DefaultComponentsDirPath() string {
...
+}
+
 func DefaultConfigFilePath() string {
 	return path_filepath.Join(defaultDaprDirPath(), defaultConfigFileName)
 }
+
+// It used to copy existing resources from components dir to resources dir.
+// TODO: Remove this function when `--components-path` flag is removed.
+func moveFilesFromComponentsToResourcesDir(componentsDirPath, resourcesDirPath string) error {
+	if _, err := os.Stat(componentsDirPath); err == nil {
```
shubham1172, line 74: Invert this please.

```
if ...; err != nil {
    return err
}
// proceed with logic
```
The caller should ensure that the src exists before calling this, otherwise we should rename the method and update the doc accordingly.

pkg/standalone/standalone.go
shubham1172: @akhilac1 we need the existing components to exist in `resources` directory. If the user modified anything beyond the standard set of files in `components`, it will be lost with a simple delete and fresh init.

shubham1172 (changes requested):

tests/e2e/standalone/utils.go
shubham1172: +1, I would also suggest using typed values instead of string in CLI (that needs some refactoring too). But for example, imageVariant, path, etc., everything is a string today. Method signatures like these can get confusing.
```go
func method(var1, var2, var3 string)
```
Instead this is better
```go
func method(var1 t1, var2 t2, var3 t3)
```

pravinpushkar: changing this to be something like- 
```
type DirName string

const (
DEFAULTCOMPONENTSDIRNAME DirName = "components"
DEFAULTRESOURCESDIRNAME  DirName = "resources"
)
```

shubham1172: Or maybe `DaprDir`, which makes the intent further clear.

```go
// DaprDir is a directory in the Dapr CLI installation path.
type DaprDir string

const (
BIN_DAPR_DIR DaprDir = "bin"
COMPONENTS_DAPR_DIR DaprDir = "components"
RESOURCES_DAPR_DIR  DaprDir = "resources"
)
```
@mukundansundar thoughts?

pkg/standalone/standalone.go
pravinpushkar: did not know why I changed this. Reverted.

pkg/standalone/standalone.go
```diff
@@ -307,6 +308,11 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		}
 		print.InfoStatusEvent(os.Stdout, "Use `%s ps` to check running containers.", runtimeCmd)
 	}
+	// TODO: remove below method when components-path flag is removed.
+	err = emptyAndCopyFiles(DefaultComponentsDirPath(), DefaultResourcesDirPath())
```
shubham1172, line 312: Should also handle here when the components directory does not exist, since `emptyAndCopyFiles` will throw an error.

tests/e2e/standalone/utils.go
pravinpushkar: Using the const from the commons.go gives error `panic: proto: file "dapr/proto/common/v1/common.proto" is already registered`

pkg/standalone/standalone.go
pravinpushkar: aah.. thats why I had the logic this way in `emptyAndCopyFiles` - 

```
if _, err := os.Stat(componentsDirPath); err == nil
```
otherwise I was returning nil..

shubham1172: Yes, but we should error out for anything other than `os.IsNotExist`.

tests/e2e/standalone/componentsDir_to_resourcesDir_mig_test.go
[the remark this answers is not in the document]
pravinpushkar: keeping it as `upgrade_test.go`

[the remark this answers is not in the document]
pravinpushkar: @mukundansundar

tests/e2e/standalone/utils.go
pravinpushkar: should we add `export GOLANG_PROTOBUF_REGISTRATION_CONFLICT=warn/ignore` for workflows ?

pravinpushkar: moved the varibales to utils.go

utils/utils.go
```diff
@@ -41,6 +41,9 @@ const (
 	DOCKER ContainerRuntime = "docker"
 	PODMAN ContainerRuntime = "podman"
 
+	DefaultComponentsDirName = "components"
```
shubham1172, line 44: Is the `default` prefix needed here? This is not a configurable option anyway, right?

pravinpushkar: I think you are right. We can change it to `DaprComponentsDirName`?

shubham1172: Ok, so `dapr init` always installs at `components`/`resources`, but overrides can happen via `dapr run`. So it makes sense to keep it as default then.

README.md
```diff
@@ -444,7 +448,7 @@ Launch Dapr and your app:
 dapr run --app-id nodeapp --app-port 3000 node app.js
 ```
 
-Note: To choose a non-default components folder, use the --resources-path(--components-path is deprecated) option.
+Note: To choose a non-default resources folder, use the --resources-path(--components-path is deprecated) option.
```
shubham1172, line 451:
```suggestion
Note: To choose a non-default resources folder, use the --resources-path option (--components-path is deprecated, see <link to github issue> for more).
```

tests/e2e/standalone/upgrade_test.go
```diff
@@ -0,0 +1,136 @@
...
+	uninstallWithoutAllFlag(t)
+
+	// dapr init should duplicate files from components dir to resources dir.
+	initTest(t)
+
+	// copy a in memomy state store component to resources dir.
+	copyInMemStateStore(t)
+
+	// check dapr run precedence order, resources dir 1st then components dir.
+	// 2nd paramter is false to indicate that both components and resources dir are present.
```
shubham1172, line 69:
```suggestion
	// 2nd parameter is false to indicate that both components and resources dir are present.
```

pkg/standalone/common.go
```diff
@@ -44,9 +46,64 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
+	return path_filepath.Join(defaultDaprDirPath(), utils.DefaultComponentsDirName)
+}
+
+func DefaultResourcesDirPath() string {
+	return path_filepath.Join(defaultDaprDirPath(), utils.DefaultResourcesDirName)
+}
+
+// DefaultResourcesDirPrecedence returns the path to the resources directory if it exists, or otherwise the components directory.
+// TODO: Remove this function and use `DefaultResourcesDirPath` when `--components-path` flag is removed.
+func DefaultResourcesDirPrecedence() string {
```
shubham1172, line 58:
```suggestion
// GetResourcesDir returns the path where the resources are stored.
func GetResourcesDir() string {
    // TODO: clean up this logic and return `DefaultResourcesDirPath` when `--components-path` flag is removed.
```

This will abstract the logic for getting the resources dir. We can try to unexport `DefaultResourcesDirPath` and `DefaultComponentsDirPath`.

pkg/standalone/common.go
```diff
@@ -44,9 +46,64 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
+	return defaultResourcesDirPath
 }
 
 func DefaultConfigFilePath() string {
 	return path_filepath.Join(defaultDaprDirPath(), defaultConfigFileName)
 }
+
+// emptyAndCopyFiles copies files from src to dest. It deletes the existing files in dest before copying from src.
+// TODO: Remove this function when `--components-path` flag is removed.
+func emptyAndCopyFiles(src, dest string) error {
```
shubham1172, line 72: Can we add unit tests for this?

pravinpushkar: most of the logic like -  "when components dir exist, when component dir does not exist" are covered under E2E tests. So, IMO we can ignore the UT because we can not verify the actual behaviour with UT here and also this function is temporary.

shubham1172 (dismissed):
LGTM, thanks @pravinpushkar!!

pkg/standalone/common.go
```diff
@@ -44,9 +46,64 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
+			return nil
+		}
+		return err
+	}
+	files, err := os.ReadDir(dest)
+	if err != nil {
+		return err
+	}
+	for _, file := range files {
+		err = os.Remove(dest + "/" + file.Name())
```
mukundansundar, line 85: yes filepath.Join ....

pkg/standalone/common.go
```diff
@@ -44,9 +46,64 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
+		}
+	}
+	files, err = os.ReadDir(src)
+	if err != nil {
+		return err
+	}
+	if len(files) > 0 {
+		print.InfoStatusEvent(os.Stdout, "Moving files from %q to %q", src, dest)
+		for _, file := range files {
+			content, err := os.ReadFile(src + "/" + file.Name())
```
mukundansundar, line 97: same as above.

mukundansundar (changes requested):

pkg/standalone/common.go
mukundansundar: Please add UTs for this function also.

mukundansundar:
@msfussell @yaron2 @greenie-msft 
There is one concern that I have with this PR. This PR introduces a `resources` directory instead of `components` directory. On init, it changes the directory to `resources`, if a prior directory `components` was there in the folder `~/.dapr`, then all the component definitions in that folder is moved/copied to `resources` dir and that is the directory used as the source of truth for default components from the next `dapr run`. 

But if there is a scenario like the following:
- user uses `dapr init` to initialize dapr  --- creates resources dir and moves all component defs to that folder
- user has an automated script that adds component CRDs to `~/.dapr/components` folder. 
- user runs `dapr run` with default arg for --resources-path/--components-path

The expectation from the user is that the newly added def in `~/.dapr/components` will also be picked up by `dapr run`, but the precedence is changed in this PR to give more precedence to `resources` dir, so only the older `resources` dir set of CRDs will be used by `dapr`. 

This is kind of a breaking change from that perspective. Should we wait for a couple of releases after `---resources-path` flag introduction to default to `resources` directory instead of `components` directory? Or should we have this change now itself?
My thought was to have it after the 1.10 release, due to the scenario described above.

akhilac1:
> @msfussell @yaron2 @greenie-msft There is one concern that I have with this PR. This PR introduces a `resources` directory instead of `components` directory. On init, it changes the directory to `resources`, if a prior directory `components` was there in the folder `~/.dapr`, then all the component definitions in that folder is moved/copied to `resources` dir and that is the directory used as the source of truth for default components from the next `dapr run`.
> 
> But if there is a scenario like the following:
> 
> * user uses `dapr init` to initialize dapr  --- creates resources dir and moves all component defs to that folder
> * user has an automated script that adds component CRDs to `~/.dapr/components` folder.
> * user runs `dapr run` with default arg for --resources-path/--components-path
> 
> The expectation from the user is that the newly added def in `~/.dapr/components` will also be picked up by `dapr run`, but the precedence is changed in this PR to give more precedence to `resources` dir, so only the older `resources` dir set of CRDs will be used by `dapr`.
> 
> This is kind of a breaking change from that perspective. Should we wait for a couple of releases after `---resources-path` flag introduction to default to `resources` directory instead of `components` directory? Or should we have this change now itself? My thought was to have it after the 1.10 release, due to the scenario described above. In the meantime make users aware that such is change is in the works and will be added in a future release. And let them for now specify a `--resources-path` directory manually or default to `~/.dapr/components` folder only.

Using a symbolic link should solve this issue vs. creating a separate directory @mukundansundar @pravinpushkar

pravinpushkar:
I need to try this out and for the transition period, even for the fresh `dapr init`(no components dir exists) we can maintain this symbolic link.

msfussell:
@mukundansundar @pravinpushkar - Yes, we need to prevent any breaking change at least over a period of two release and have a deprecation policy. There needs to be a deprecation announcement to go along with this and maybe we can also have something in the CLI is you use the `--components-path` flag

shubham1172:
Based on the different feedback in this thread, can we do the following during `dapr init`?

1. Copy files from `components` to `resources` directory
2. Delete `components` directory
3. Create a symlink to `resources` called `components`

- This takes care of the scenario mentioned by @mukundansundar since when an automated script copies anything to `components`, they end up copying it to `resources`. 
- The `resources` directory ultimately contains all the files (as we want it to), and clean up will only require deleting the symlink `components`.
- Won't be a breaking change for users.

pravinpushkar:
> Based on the different feedback in this thread, can we do the following during `dapr init`?
> 
> 1. Copy files from `components` to `resources` directory
> 2. Delete `components` directory
> 3. Create a symlink to `resources` called `components`
> 
> * This takes care of the scenario mentioned by @mukundansundar since when an automated script copies anything to `components`, they end up copying it to `resources`.
> * The `resources` directory ultimately contains all the files (as we want it to), and clean up will only require deleting the symlink `components`.
> * Won't be a breaking change for users.

One problem still is what if user deletes the resources dir(this might happen, as it is newly introduced and users not aware of it). This might happen other way also if `resources` dir is the symlink and some user deletes the component directory.

If we are fine with the above limitation then we could go the suggested approach of having `components` as symlink for `resources` directory. 
@mukundansundar @msfussell @yaron2 thoughts ?

shubham1172:
> One problem still is what if user deletes the resources dir(this might happen, as it is newly introduced and users not aware of it). This might happen other way also if `resources` dir is the symlink and some user deletes the component directory.

Isn't this the same problem as user deleting the `components` directory today? :) Our release notes have to highlight that it's an important directory. Deleting anything inside `.dapr` might have consequences for users. Also after a few releases, we will stop creating the symlink for `components` too.

msfussell:
> > Based on the different feedback in this thread, can we do the following during `dapr init`?
> > 
> > 1. Copy files from `components` to `resources` directory
> > 2. Delete `components` directory
> > 3. Create a symlink to `resources` called `components`
> > 
> > 
> > * This takes care of the scenario mentioned by @mukundansundar since when an automated script copies anything to `components`, they end up copying it to `resources`.
> > * The `resources` directory ultimately contains all the files (as we want it to), and clean up will only require deleting the symlink `components`.
> > * Won't be a breaking change for users.
> 
> One problem still is what if user deletes the resources dir(this might happen, as it is newly introduced and users not aware of it). This might happen other way also if `resources` dir is the symlink and some user deletes the component directory.
> 
> If we are fine with the above limitation then we could go the suggested approach of having `components` as symlink for `resources` directory. @mukundansundar @msfussell @yaron2 thoughts ?

Yes, this is a fine way to go. I would also like to include the deprecation warning on using the --components-path flag and the suggestion to use --resources-path

pravinpushkar:
>  I would also like to include the deprecation warning on using the --components-path flag and the suggestion to use --resources-path

this is already there.

pkg/standalone/common.go
```diff
@@ -44,9 +47,94 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
 }
 
 func DefaultConfigFilePath() string {
 	return path_filepath.Join(defaultDaprDirPath(), defaultConfigFileName)
 }
+
+// copyFilesAndCreateSymlink copies files from src to dest. It deletes the existing files in dest before copying from src.
+// this method also deletes the components dir and makes it as a symlink to resources directory.
+// please see this comment for more details:https://github.com/dapr/cli/pull/1149#issuecomment-1364424345
+// TODO: Remove this function when `--components-path` flag is removed.
```
shubham1172, line 74: I have a suggestion to break it into two methods. (1) moveDir(src, dest) (which will delete files in dest, copy things from src to dest, and delete src - essentially a `mv` operation) , and (2) createSymLink (which you already have). Easier to read and maintain. The caller will then invoke these two methods serially.

Also we can replace references to `resources` and `components` in these functions, only use `src` and `dest` terminologies.

shubham1172 (changes requested):

pkg/standalone/common_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+import (
+	"os"
+	path_filepath "path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateSymLink(t *testing.T) {
```
shubham1172, line 24: Same comment for these tests, and can we use `src` and `dest` instead of `resources` and `components`?

tests/e2e/standalone/upgrade_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+func copyInMemStateStore(t *testing.T) {
+	filePath := filepath.Join("../testdata/resources", "test-statestore.yaml")
+	content, err := os.ReadFile(filePath)
+	assert.NoError(t, err, "cannot read testdata/resources/test-statestore.yaml file")
+	err = os.WriteFile(filepath.Join(defaultResourcesDirPath, "test-statestore.yaml"), content, 0644)
+	assert.NoError(t, err, "cannot write testdata/resources/test-statestore.yaml file to resources directory")
+}
+
+func checkDaprRunPrecedenceTest(t *testing.T, inMemoryCompPresent bool, flags ...string) {
```
shubham1172, line 73:
```suggestion
func testDaprRunOutput(t *testing.T, InMemCompLoaded bool, flags ...string) {
```

tests/e2e/standalone/upgrade_test.go
```diff
@@ -0,0 +1,89 @@
...
+	// Ensure a clean environment.
+	must(t, cmdUninstall, "failed to uninstall Dapr")
+
+	// install dapr -> installs dapr, creates resources dir and symlink components dir.
+	ensureDaprInstallation(t)
+
+	// check dapr run -> should not load in-memory component.
+	checkDaprRunPrecedenceTest(t, false)
+
+	// copy a in memomy state store component to resources dir.
```
shubham1172, line 52:
```suggestion
	// Copy an in-memory state store component to the resources directory.
```

tests/e2e/standalone/upgrade_test.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+var (
+	// DefaultComponentsDirPath is the default components directory path.
+	defaultComponentsDirPath = ""
+	defaultResourcesDirPath  = ""
+)
+
+// It covers the test flow for scenario when user does: (1) dapr uninstall (2) upgrades dapr cli (3) dapr init (4) dapr run.
+func TestCompToResrcDirMig(t *testing.T) {
```
shubham1172, line 38:
```suggestion
// Tests precedence for --components-path and --resources-path flags.
func TestComponentsAndResourcesPathPrecedence(t *testing.T) {
```

tests/e2e/standalone/upgrade_test.go
```diff
@@ -0,0 +1,89 @@
...
+	copyInMemStateStore(t)
+
+	// check dapr run -> should load in-memory component.
+	checkDaprRunPrecedenceTest(t, true)
+
+	// dapr run with --components-path flag -> should load in-memory component.
+	checkDaprRunPrecedenceTest(t, true, "--components-path", defaultComponentsDirPath)
+
+	// dapr run with --resources-path flag -> should load in-memory component.
+	checkDaprRunPrecedenceTest(t, true, "--resources-path", defaultResourcesDirPath)
```
shubham1172, line 62: Can we also add a test where both flags are given?

tests/e2e/standalone/upgrade_test.go
```diff
@@ -0,0 +1,89 @@
...
+	// install dapr -> installs dapr, creates resources dir and symlink components dir.
+	ensureDaprInstallation(t)
+
+	// check dapr run -> should not load in-memory component.
+	checkDaprRunPrecedenceTest(t, false)
+
+	// copy a in memomy state store component to resources dir.
+	copyInMemStateStore(t)
+
+	// check dapr run -> should load in-memory component.
```
shubham1172, line 55: Can you add where is the component coming from here?

pkg/standalone/common.go
```diff
@@ -44,9 +47,104 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
+				return fmt.Errorf("error reading file %s: %w", file.Name(), err)
+			}
+			// #nosec G306
+			err = os.WriteFile(path_filepath.Join(dest, file.Name()), content, 0o644)
+			if err != nil {
+				return fmt.Errorf("error writing file %s: %w", file.Name(), err)
+			}
+		}
+	}
+	// delete the components dir and make it as a symlink to resources directory.
```
shubham1172, line 129: Reference to components/resources - please remove

pkg/standalone/common.go
```diff
@@ -44,9 +47,104 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
+	destFiles, err := os.ReadDir(dest)
+	if err != nil {
+		return fmt.Errorf("error reading files from %s: %w", dest, err)
+	}
+	srcFiles, err := os.ReadDir(src)
+	if err != nil {
+		return fmt.Errorf("error reading files from %s: %w", src, err)
+	}
+	if len(srcFiles) > 0 {
+		// delete the existing files in dest before copying from src iff there are files in src.
```
shubham1172, line 108:
```suggestion
		// delete the existing files in dest before copying from src if there are files in src.
```

pkg/standalone/common.go
```diff
@@ -44,9 +47,104 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
+	if err = moveDir(src, dest); err != nil {
+		return err
+	}
+	if err = createSymLink(dest, src); err != nil {
+		return err
+	}
+	return nil
+}
+
+// moveDir moves files from src to dest. If there are files in src, it deletes the existing files in dest before copying from src.
```
shubham1172, line 97: Please mention that `src` directory is deleted.

pkg/standalone/common.go
```diff
@@ -44,9 +47,104 @@ func binaryFilePath(binaryDir string, binaryFilePrefix string) string {
...
+		return fmt.Errorf("error reading directory %s: %w", src, err)
+	}
+	if err = moveDir(src, dest); err != nil {
+		return err
+	}
+	if err = createSymLink(dest, src); err != nil {
+		return err
+	}
+	return nil
+}
```
shubham1172, line 95: Is this required here now? Can this be a part of `standalone.go` directly? This is no more a file abstraction, it's directly a bunch of operations now.

shubham1172 (approved):

tests/e2e/standalone/utils.go
```diff
@@ -156,3 +162,23 @@ func containerRuntime() string {
...
+
+func uninstallDapr(uninstallArgs ...string) (string, error) {
+	daprContainerRuntime := containerRuntime()
+
+	// Add --container-runtime flag only if daprContainerRuntime is not empty, or overridden via args.
+	// This is only valid for non-slim mode.
+	if !isSlimMode() && daprContainerRuntime != "" && !utils.Contains(uninstallArgs, "--container-runtime") {
+		uninstallArgs = append(uninstallArgs, "--container-runtime", daprContainerRuntime)
+	}
+	return spawn.Command(common.GetDaprPath(), uninstallArgs...)
```
shubham1172, line 183: Why not reuse `cmdUninstall`, similar to `installDapr` function above?

tests/e2e/standalone/commands.go
```diff
@@ -125,20 +124,18 @@ func cmdStop(appId string, args ...string) (string, error) {
...
+func cmdUninstallAll(args ...string) (string, error) {
 	uninstallArgs := []string{"uninstall", "--all"}
-
-	daprContainerRuntime := containerRuntime()
-
-	// Add --container-runtime flag only if daprContainerRuntime is not empty, or overridden via args.
-	// This is only valid for non-slim mode.
-	if !isSlimMode() && daprContainerRuntime != "" && !utils.Contains(args, "--container-runtime") {
-		uninstallArgs = append(uninstallArgs, "--container-runtime", daprContainerRuntime)
-	}
```
shubham1172, line 138: Why did we remove this instead of moving it to `cmdUninstall` and using this instead of the new function in `utils`?

mukundansundar:
@pravinpushkar Please fix conflicts in this PR. 

@msfussell @yaron2 I think this PR can go in 1.11. (Renaming default `components` dir to `resources` dir) Thoughts?

Issue #1142, opened 2022-12-06
Implement the new precedence rules for `resources_dir` for `dapr run -f`


mukundansundar:
## Describe the proposal
- Implement the new precedence rules for `resources_dir` for `dapr run -f`

Add capability to look for resources under given input directory and fallback to default if not present with UTs.
See the folder organization structure  in dapr/proposals#6.

related to #1123 

## Release Note

RELEASE NOTE:

PR #1158, opened 2022-12-30
[WIP] Precedence order for resources directory and config file

pravinpushkar:
# Description

Just a draft PR. It will be worked upon the template parsing PR. I will update the description when it is time.

## Issue reference

Please reference the issue this PR will close: #1142 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pravinpushkar:
Please check the `.gitkeep` files and associated structure. I decided to keep this test directory structure as part of the testdata itself. This is because one thing it is very clear what is begin tested and can be verified visually also. 

We can not use `temp` directory as directories are created as mentioned in the test config file and are looked up w.r.t test file. Alternatively we can create the structure as it is there in run config file but it will be hard to figure out what is expected from the test. thoughts ?

pkg/standalone/testdata/runfileconfig/app/resources/.gitkeep
```diff
@@ -0,0 +1 @@
+# dummy file
```
shubham1172, line 1:
```suggestion
# empty file
```

pkg/standalone/runfileconfig/run_file_config_parser_test.go
```diff
@@ -154,3 +163,51 @@ func TestGetBasePathFromAbsPath(t *testing.T) {
...
+	assert.Nil(t, err)
+	err = backendConfigFile.Close()
+	assert.Nil(t, err)
+
+	err = os.MkdirAll(webappAppResourcesPath, 0o777)
+	assert.Nil(t, err)
+
+	webAppConfigFile, err := os.Create(webAppAppConfigPath)
+	assert.Nil(t, err)
+	err = webAppConfigFile.Close()
```
shubham1172, line 202: Maybe create a helper function and pass a slice of directories to create/cleanup?

pkg/standalone/run.go
```diff
@@ -68,6 +74,7 @@ type SharedRunConfig struct {
 	AppHealthTimeout   int    `arg:"app-health-probe-timeout" ifneq:"0" yaml:"app_health_probe_timeout"`
 	AppHealthThreshold int    `arg:"app-health-threshold" ifneq:"0" yaml:"app_health_threshold"`
 	EnableAPILogging   bool   `arg:"enable-api-logging" yaml:"enable_api_logging"`
+	DaprPathCmdFlag    string `yaml:"dapr_path_cmd_flag"`
```
mukundansundar, line 77:
```suggestion
	DaprPathCmdFlag    string `yaml:"dapr_path"`
```

pkg/standalone/run.go
```diff
@@ -44,8 +50,8 @@ type RunConfig struct {
 	MetricsPort      int      `env:"DAPR_METRICS_PORT" arg:"metrics-port" yaml:"metrics_port"`
 	UnixDomainSocket string   `arg:"unix-domain-socket" yaml:"unix_domain_socket"`
 	InternalGRPCPort int      `arg:"dapr-internal-grpc-port" yaml:"dapr_internal_grpc_port"`
-	DaprPathCmdFlag  string   `yaml:"dapr_path_cmd_flag"`
 	SharedRunConfig  `yaml:",inline"`
+	Env              []EnvItems `yaml:"env"`
```
mukundansundar, line 54: please move back Env inside the runfileconfig for now ...

pkg/standalone/run.go
```diff
@@ -44,8 +50,8 @@ type RunConfig struct {
 	MetricsPort      int      `env:"DAPR_METRICS_PORT" arg:"metrics-port" yaml:"metrics_port"`
 	UnixDomainSocket string   `arg:"unix-domain-socket" yaml:"unix_domain_socket"`
 	InternalGRPCPort int      `arg:"dapr-internal-grpc-port" yaml:"dapr_internal_grpc_port"`
-	DaprPathCmdFlag  string   `yaml:"dapr_path_cmd_flag"`
 	SharedRunConfig  `yaml:",inline"`
```
shubham1172, line 53: Quick question, should `SharedRunConfig` be embedded in the `RunConfig` instead of being a field?

pkg/standalone/runfileconfig/run_file_config_parser_test.go
```diff
@@ -133,3 +159,73 @@ func TestGetBasePathFromAbsPath(t *testing.T) {
...
 	}
 }
+
+// Expected directory and file structures from this method:
+// 1) testdata/runfileconfig/app/resources.
+// 2) testdata/runfileconfig/backend/.dapr/resources.
+// 3) testdata/runfileconfig/backend/.dapr/config.yaml.
+// 4) testdata/runfileconfig/webapp/resources.
+// 5) testdata/runfileconfig/webapp/config.yaml.
+func testSetup(t *testing.T) func(t *testing.T) {
```
mukundansundar, line 169: Is there a reason to remove .gitkeep from the empty dirs and recreate them on the fly here?

pkg/standalone/run.go
shubham1172: Ok never mind, missed it. @pravinpushkar can you please add this to the top of the struct, embedded values by convention appear on the top.

pkg/standalone/runfileconfig/run_file_config_parser_test.go
shubham1172: Isn't this better than keeping .gitkeep with empty folders? The tests are self-sustained now. We don't even need to use testdata, we can use /tmp.

mukundansundar: IMO, the less side effects(os calls) that we have with UT, the better. File and dir creation should ideally be performed in E2E.

mukundansundar: Again even it .gitkeep, this does not affect the tests much ...
And we cannot use /tmp since the yaml file must point to the resources dir. So getting a random temp dir is not going to be possible.

shubham1172: >  And we cannot use /tmp since the yaml file must point to the resources dir. So getting a random temp dir is not going to be possible.

Makes sense!

> IMO, the less side effects(os calls) that we have with UT, the better. File and dir creation should ideally be performed in E2E.

True, ideally we should test with some file system abstraction.

pkg/standalone/run.go
pravinpushkar: By making it as part of Runconfig, we can reuse the same logic of passing the envs into child process.

pkg/standalone/runfileconfig/run_file_config_parser_test.go
pravinpushkar: There will be a separate PR for file system abstraction.

pkg/standalone/runfileconfig/run_file_config.go
```diff
@@ -26,18 +26,11 @@ type RunFileConfig struct {
...
 	standalone.RunConfig `yaml:",inline"`
-	AppDirPath           string     `yaml:"app_dir_path"`
-	Env                  []EnvItems `yaml:"env"`
+	AppDirPath           string `yaml:"app_dir_path"`
 }
 
 // Common represents the configuration options for the common section in the run file.
 type Common struct {
-	Env                        []EnvItems `yaml:"env"`
+	Env                        []standalone.EnvItems `yaml:"env"`
```
mukundansundar, line 34: why can't `Env []EnvItems` be part of SharedRunConfig? 
That was Common will only be the embedded struct SharedRunConfig?

pkg/standalone/run.go
```diff
@@ -48,30 +55,30 @@ type RunConfig struct {
...
+	AppSSL             bool       `arg:"app-ssl" yaml:"app_ssl"`
+	MaxRequestBodySize int        `arg:"dapr-http-max-request-size" yaml:"dapr_http_max_request_size" default:"-1"`
+	HTTPReadBufferSize int        `arg:"dapr-http-read-buffer-size" yaml:"dapr_http_read_buffer_size" default:"-1"`
+	EnableAppHealth    bool       `arg:"enable-app-health-check" yaml:"enable_app_health_check"`
+	AppHealthPath      string     `arg:"app-health-check-path" yaml:"app_health_check_path"`
+	AppHealthInterval  int        `arg:"app-health-probe-interval" ifneq:"0" yaml:"app_health_probe_interval"`
+	AppHealthTimeout   int        `arg:"app-health-probe-timeout" ifneq:"0" yaml:"app_health_probe_timeout"`
+	AppHealthThreshold int        `arg:"app-health-threshold" ifneq:"0" yaml:"app_health_threshold"`
+	EnableAPILogging   bool       `arg:"enable-api-logging" yaml:"enable_api_logging"`
+	DaprPathCmdFlag    string     `yaml:"dapr_path"`
```
mukundansundar, line 80:
```suggestion
	DaprdInstallPath    string     `yaml:"dapr_path"`
```

pkg/standalone/runfileconfig/run_file_config_parser_test.go
```diff
@@ -27,74 +27,96 @@ var (
...
-
-	firstAppConfig := appsRunConfig.Apps[0]
-	secondAppConfig := appsRunConfig.Apps[1]
-	assert.Equal(t, "", firstAppConfig.AppID)
-	assert.Equal(t, "GRPC", secondAppConfig.AppProtocol)
-	assert.Equal(t, 8080, firstAppConfig.AppPort)
-	assert.Equal(t, "", firstAppConfig.UnixDomainSocket)
-}
+func TestRunConfigFile(t *testing.T) {
+	t.Run("test run config parser", func(t *testing.T) {
```
mukundansundar, line 31:
```suggestion
	t.Run("test parse valid run template", func(t *testing.T) {
```

pkg/standalone/runfileconfig/run_file_config_parser_test.go
```diff
@@ -27,74 +27,96 @@ var (
...
+		// test resources_path and config_path after precedence order logic.
+		assert.Equal(t, filepath.Join(apps[0].AppDirPath, "resources"), apps[0].ResourcesPath)
+		assert.Equal(t, filepath.Join(apps[1].AppDirPath, ".dapr", "resources"), apps[1].ResourcesPath)
+
+		assert.Equal(t, filepath.Join(apps[0].AppDirPath, "config.yaml"), apps[0].ConfigFile)
+		assert.Equal(t, filepath.Join(apps[1].AppDirPath, ".dapr", "config.yaml"), apps[1].ConfigFile)
+
+		// temporarily set apps[0].ResourcesPath to empty string to test it is getting picked from common section.
+		apps[0].ResourcesPath = ""
+		config.resolveResourcesAndConfigFilePaths()
```
mukundansundar, line 76: please add a separate UT for this function

pkg/standalone/runfileconfig/run_file_config_parser_test.go
```diff
@@ -27,74 +27,96 @@ var (
...
+
+		// test merged envs from common and app sections.
+		assert.Equal(t, 2, len(apps[0].Env))
+		assert.Equal(t, 2, len(apps[1].Env))
+		assert.Equal(t, "DEBUG", apps[0].Env[0].Name)
+		assert.Equal(t, "false", apps[0].Env[0].Value)
+		assert.Equal(t, "DEBUG", apps[1].Env[0].Name)
+		assert.Equal(t, "true", apps[1].Env[0].Value)
+	})
+	t.Run("test validate config files", func(t *testing.T) {
```
mukundansundar, line 88:
```suggestion
	t.Run("test validate run config", func(t *testing.T) {
```

mukundansundar (changes requested):
small changes

pkg/standalone/runfileconfig/run_file_config_parser_test.go
mukundansundar: add UTs that test all different use cases

pkg/standalone/runfileconfig/run_file_config_parser_test.go
```diff
@@ -27,74 +29,173 @@ var (
...
+					config.Apps[tc.appIndex].ConfigFile = ""
+				}
+				if tc.setDaprdInstallPath {
+					config.Apps[tc.appIndex].DaprdInstallPath = config.Apps[1].AppDirPath // set to apps[1].AppDirPath teporarily.
+				}
+				// test precedence logic for resources_path and config_path.
+				config.resolveResourcesAndConfigFilePaths()
+				assert.Equal(t, tc.expectedResourcesPath, config.Apps[tc.appIndex].ResourcesPath)
+				assert.Equal(t, tc.expectedConfigFilePath, config.Apps[tc.appIndex].ConfigFile)
+			})
```
mukundansundar, line 163: as discussed offline, add test for default resources path resolution.

pkg/standalone/run.go
```diff
@@ -37,8 +37,15 @@ const (
 	defaultStructTagKey  = "default"
 )
 
+// EnvItems represents the env configuration options that are present in commmon and/or individual app's section.
+type EnvItems struct {
+	Name  string `yaml:"name"`
+	Value string `yaml:"value"`
+}
+
```
shubham1172, line 45: Discussed with Pravin to change this to a simple `map[string]string`, which will (a) simplify the YAML, (b) make the merging b/w common and app environment variables much simpler.

shubham1172 (changes requested):

pkg/standalone/common.go
```diff
@@ -20,10 +20,12 @@ import (
 )
 
 const (
-	defaultDaprDirName       = ".dapr"
+	DefaultDaprDirName    = ".dapr"
+	DefaultConfigFileName = "config.yaml"
+	ResourcesDirName      = "resources"
```
mukundansundar, line 25: rename to `DefaultResourcesDirName`

mukundansundar (changes requested):

mukundansundar (approved):
lgtm

pkg/standalone/runfileconfig/run_file_config_parser.go
```diff
@@ -155,3 +162,83 @@ func (a *RunFileConfig) resolvePathToAbsAndValidate(baseDir string, paths ...*st
...
+				a.Apps[i].Env[k] = v
+			}
+		}
+	}
+}
+
+// resolveResourcesFilePath resolves the resources path for the app.
+// Precedence order for resources_path -> apps[i].resources_path > apps[i].app_dir_path/.dapr/resources > common.resources_path > dapr default resources path.
+func (a *RunFileConfig) resolveResourcesFilePath(app *Apps) error {
+	if app.ResourcesPath == "" {
```
shubham1172, line 203: Please invert.
```go
if app.ResourcesPath != "" {
    return nil
}
```

pkg/standalone/runfileconfig/run_file_config_parser_test.go
```diff
@@ -133,3 +261,22 @@ func TestGetBasePathFromAbsPath(t *testing.T) {
...
+func getResourcesAndConfigFilePaths(t *testing.T, daprInstallPath string) []string {
+	t.Helper()
+	result := make([]string, 2)
+	var daprDirPath string
+	var err error
+	if daprInstallPath == "" {
+		daprDirPath, err = standalone.GetDaprPath(daprInstallPath)
+		assert.NoError(t, err)
+		result[0] = standalone.GetDaprComponentsPath(daprDirPath)
+		result[1] = standalone.GetDaprConfigPath(daprDirPath)
```
shubham1172, line 275: Isn't this always being overwritten?

pravinpushkar: hmm.. typo it is.

shubham1172 (approved):
LGTM, thanks @pravinpushkar!!

mukundansundar (approved):
good to me it looks :)

PR #1159, opened 2023-01-02
Add e2e timeout to Dockerhub runs and update Makefile default

shubham1172:

# Description

Earlier e2e test timeout was only applicable for GHCR cron runs, now it's also extended to Dockerhub.
The Makefile is also updated to always use a default value if the environment variable is unset.

## Issue reference

NA

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,9 @@ export GOOS ?= $(TARGET_OS_LOCAL)
 export BINARY_EXT ?= $(BINARY_EXT_LOCAL)
 
 TEST_OUTPUT_FILE ?= test_output.json
-E2E_SH_TEST_TIMEOUT ?= 10m
+
+# Set the default timeout for tests to 10 minutes
+E2E_SH_TEST_TIMEOUT ?= $(if $(E2E_SH_TEST_TIMEOUT), $(E2E_SH_TEST_TIMEOUT), 10m)
 
 # Use the variable H to add a header (equivalent to =>) to informational output
 H = $(shell printf "\033[34;1m=>\033[0m")
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,9 @@ export BINARY_EXT ?= $(BINARY_EXT_LOCAL)
 TEST_OUTPUT_FILE ?= test_output.json
 
 # Set the default timeout for tests to 10 minutes
-E2E_SH_TEST_TIMEOUT ?= $(if $(E2E_SH_TEST_TIMEOUT), $(E2E_SH_TEST_TIMEOUT), 10m)
+ifndef E2E_SH_TEST_TIMEOUT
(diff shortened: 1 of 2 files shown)


shubham1172:
MacOS

gotestsum --jsonfile /Users/runner/work/cli/cli/test-e2e-standalone.json --format standard-verbose -- -timeout **20m** -count=1 -tags=e2e ./tests/e2e/standalone/...


Linux

gotestsum --jsonfile /home/runner/work/cli/cli/test-e2e-standalone.json --format standard-verbose -- -timeout **10m** -count=1 -tags=e2e ./tests/e2e/standalone/...

Makefile
```diff
@@ -70,7 +70,11 @@ export GOOS ?= $(TARGET_OS_LOCAL)
 export BINARY_EXT ?= $(BINARY_EXT_LOCAL)
 
 TEST_OUTPUT_FILE ?= test_output.json
-E2E_SH_TEST_TIMEOUT ?= 10m
+
+# Set the default timeout for tests to 10 minutes
```
pravinpushkar, line 74: is not this same as `E2E_SH_TEST_TIMEOUT ?= 10m.

shubham1172: Turns out, no. `?=` only sets a variable if it's not already set. In our case, it was set, but to `""`.

pravinpushkar (approved):

mukundansundar (commented):
@shubham1172 one suggestion ... for the self hosted e2e alone keep the timeout at 15 mins .... It is only timeout and that should not affect the actual tests ... 
The point is do we need the complex conditions to set different values based on arch?

shubham1172:
> @shubham1172 one suggestion ... for the self hosted e2e alone keep the timeout at 15 mins .... It is only timeout and that should not affect the actual tests ... The point is do we need the complex conditions to set different values based on arch?

It's better to fail early if something goes wrong, isn't it? The difference is mostly due to podman vs docker in Mac and other runners. We CAN revert this to use a single value, but let's do it in a separate PR since a part of this logic is already merged to `master`, but I am OK with keeping different timeouts too. Let me know your thoughts.

mukundansundar (approved):

mukundansundar:
> > @shubham1172 one suggestion ... for the self hosted e2e alone keep the timeout at 15 mins .... It is only timeout and that should not affect the actual tests ... The point is do we need the complex conditions to set different values based on arch?
> 
> It's better to fail early if something goes wrong, isn't it? The difference is mostly due to podman vs docker in Mac and other runners. We CAN revert this to use a single value, but let's do it in a separate PR since a part of this logic is already merged to `master`, but I am OK with keeping different timeouts too. Let me know your thoughts.

IMO it can be done as a separate PR.

PR #1169, opened 2023-01-18
Dapr stop and list changes for dapr run -f

### LINKED ISSUE #1153: [Proposal] `dapr list --output json` includes related Dapr Compose configuration file metadata
## Describe the proposal

This proposal  is related to the [Dapr Compose](https://github.com/dapr/cli/issues/1123) feature work currently underway.

`dapr list` shows all of the Dapr sidecars created across the local system, whether part of any Dapr Compose operation or started independently.  In order for tooling to generate a complete model of a system of applications started by Dapr Compose, in particular when *not* started by that tooling itself, it needs to be able to correlate individual sidecars back to its associated Dapr Compose configuration (if any).  This will allow tooling to present a more wholistic view of the application system, rather than a simple list of sidecars.

I propose that the `dapr list --output json` include a property that references the path to the Dapr Compose configuration file (if any) that caused that sidecar to be created.

For example, the addition of a `composeFilePath` property:

```bash
$ dapr list --output json
{
  "appId": "test",
  "httpPort": 59514,
  "grpcPort": 59515,
  "appPort": 0,
  "metricsEnabled": true,
  "command": "",
  "age": "3s",
  "created": "2022-12-16 09:39.45",
  "daprdPid": 43496,
  "cliPid": 43491,
  "maxRequestBodySize": -1,
  "httpReadBufferSize": -1,
  "composeFilePath": "/Users/user/src/app/dapr.yaml"
}
```

If the sidecar is not associated with any Dapr Compose operation, then the property may be `null` or omitted.

## Release Note

RELEASE NOTE:

### LINKED ISSUE #1165: [Proposal] Enhance `dapr stop` functionality to handle stopping multiple apps
## Describe the proposal

Related to  #1143, this proposal talks about enhancing the `dapr stop` experience to accommodate (a) stopping all apps, and (b) stopping a single app, started by the `dapr run -f` command.

1. When an application is started individually (using simple `dapr run`), `dapr stop --app-id <APP-ID>` should stop the application along with the CLI process associated with it.
1. When an application is started using `dapr run -f` (along with other apps), `dapr stop --app-id <APP-ID>` should 
    1. Warn the user that this application was started along with other apps (log the app-ids)
    1. Stop the application process
    1. If there are no other applications associated with the same CLI process that spawned this application, kill the CLI process; else, skip it.
1. When multiple applications are started with `dapr run -f dapr.yaml`, `dapr stop -f dapr.yaml` should stop all the running apps (based on app-id) along with the associated CLI process.

## Release Note

RELEASE NOTE: **ADD** support to `dapr stop` to handle applications spawned by `dapr run -f`

pravinpushkar:
# Description

## Issue reference

Please reference the issue this PR will close: 
closes #1165 
closes #1153

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/dapr.go
+++ b/cmd/dapr.go
@@ -46,6 +46,12 @@ type daprVersion struct {
 	RuntimeVersion string `json:"Runtime version"`
 }
 
+type osType string
+
+const (
+	WINDOWS osType = "windows"
+)
+
 var (
 	daprVer   daprVersion
 	logAsJSON bool
--- a/cmd/dapr.go
+++ b/cmd/dapr.go
@@ -49,7 +49,7 @@ type daprVersion struct {
 type osType string
 
 const (
-	WINDOWS osType = "windows"
+	windowsOsType osType = "windows"
 )
 
 var (
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -106,7 +106,7 @@ dapr run --run-file /path/to/directory
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(runFilePath) > 0 {
-			if runtime.GOOS == string(WINDOWS) {
+			if runtime.GOOS == string(windowsOsType) {
 				print.FailureStatusEvent(os.Stderr, "The run command with run file is not supported on Windows")
 				os.Exit(1)
 			}
@@ -140,7 +140,7 @@ dapr run --run-file /path/to/directory
 
 		if unixDomainSocket != "" {
 			// TODO(@daixiang0): add Windows support.
-			if runtime.GOOS == string(WINDOWS) {
+			if runtime.GOOS == string(windowsOsType) {
 				print.FailureStatusEvent(os.Stderr, "The unix-domain-socket option is not supported on Windows")
 				os.Exit(1)
 			} else {
@@ -510,13 +510,13 @@ func executeRun(runFilePath string, apps []runfileconfig.App) (bool, error) {
 		runStates = append(runStates, runState)
 
 		// Metadata API is only available if app has started listening to port, so wait for app to start before calling metadata API.
-		_ = putCLIProcessIDInMeta(runState, os.Getpid())
+		putCLIProcessIDInMeta(runState, os.Getpid())
 
 		// Update extended metadata with run file path.
-		_ = putRunFilePathInMeta(runState, runFilePath)
+		putRunFilePathInMeta(runState, runFilePath)
 
 		if runState.AppCMD.Command != nil {
-			_ = putAppCommandInMeta(runConfig, runState)
+			putAppCommandInMeta(runConfig, runState)
 		}
 		print.StatusEvent(runState.DaprCMD.OutputWriter, print.LogSuccess, "You're up and running! Dapr logs will appear here.\n")
 		logInformationalStatusToStdout(app)
@@ -894,43 +894,37 @@ func killAppProcess(runE *runExec.RunExec) error {
 }
 
 // putCLIProcessIDInMeta puts the CLI process ID in metadata so that it can be used by the CLI to stop the CLI process.
-func putCLIProcessIDInMeta(runE *runExec.RunExec, pid int) error {
-	// For now putting this as 0, since we do not want the dapr stop command for a single to kill the CLI process,
-	// thereby killing all the apps that are running via dapr run -f.
+func putCLIProcessIDInMeta(runE *runExec.RunExec, pid int) {
+	print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogInfo, "Updating metadata for cliPID: %d", pid)
 	err := metadata.Put(runE.DaprHTTPPort, "cliPID", strconv.Itoa(pid), runE.AppID, unixDomainSocket)
 	if err != nil {
 		print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogWarning, "Could not update sidecar metadata for cliPID: %s", err.Error())
-		return err
 	}
-	return nil
 }
 
 // putAppCommandInMeta puts the app command in metadata so that it can be used by the CLI to stop the app.
-func putAppCommandInMeta(runConfig standalone.RunConfig, runState *runExec.RunExec) error {
+func putAppCommandInMeta(runConfig standalone.RunConfig, runE *runExec.RunExec) {
 	appCommand := strings.Join(runConfig.Command, " ")
-	print.StatusEvent(runState.DaprCMD.OutputWriter, print.LogInfo, "Updating metadata for app command: %s", appCommand)
-	err := metadata.Put(runState.DaprHTTPPort, "appCommand", appCommand, runState.AppID, runConfig.UnixDomainSocket)
+	print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogInfo, "Updating metadata for app command: %s", appCommand)
+	err := metadata.Put(runE.DaprHTTPPort, "appCommand", appCommand, runE.AppID, runConfig.UnixDomainSocket)
 	if err != nil {
-		print.StatusEvent(runState.DaprCMD.OutputWriter, print.LogWarning, "Could not update sidecar metadata for appCommand: %s", err.Error())
-		return err
+		print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogWarning, "Could not update sidecar metadata for appCommand: %s", err.Error())
+		return
 	}
-	print.StatusEvent(runState.DaprCMD.OutputWriter, print.LogSuccess, "You're up and running! Dapr logs will appear here.\n")
-	return nil
+	print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogSuccess, "You're up and running! Dapr logs will appear here.\n")
 }
 
 // putRunFilePathInMeta puts the absolute path of run file in metadata so that it can be used by the CLI to stop all apps started by this run file.
-func putRunFilePathInMeta(runE *runExec.RunExec, runFilePath string) error {
+func putRunFilePathInMeta(runE *runExec.RunExec, runFilePath string) {
 	runFilePath, err := filepath.Abs(runFilePath)
 	if err != nil {
 		print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogWarning, "Could not get absolute path for run file: %s", err.Error())
-		return err
+		return
 	}
 	err = metadata.Put(runE.DaprHTTPPort, "runFile", runFilePath, runE.AppID, unixDomainSocket)
 	if err != nil {
 		print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogWarning, "Could not update sidecar metadata for runFile: %s", err.Error())
-		return err
 	}
-	return nil
 }
 
 // getRunFilePath returns the path to the run file.
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -921,7 +921,7 @@ func putRunFilePathInMeta(runE *runExec.RunExec, runFilePath string) {
 		print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogWarning, "Could not get absolute path for run file: %s", err.Error())
 		return
 	}
-	err = metadata.Put(runE.DaprHTTPPort, "runFile", runFilePath, runE.AppID, unixDomainSocket)
+	err = metadata.Put(runE.DaprHTTPPort, "runTemplatePath", runFilePath, runE.AppID, unixDomainSocket)
 	if err != nil {
 		print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogWarning, "Could not update sidecar metadata for runFile: %s", err.Error())
 	}
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -445,7 +445,7 @@ func init() {
 	RunCmd.Flags().IntVar(&appHealthThreshold, "app-health-threshold", 0, "Number of consecutive failures for the app to be considered unhealthy")
 	RunCmd.Flags().BoolVar(&enableAPILogging, "enable-api-logging", false, "Log API calls at INFO verbosity. Valid values are: true or false")
 	RunCmd.Flags().StringVar(&apiListenAddresses, "dapr-listen-addresses", "", "Comma separated list of IP addresses that sidecar will listen to")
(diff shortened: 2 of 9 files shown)


pravinpushkar:
I will cover the `stop` and `list` enhancement through E2E tests once this PR gets merged -   #1170

pkg/standalone/stop.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // Stop terminates the application process.
-func Stop(appID string) error {
+func Stop(appID string, cliPIDToNoOfApps map[int]int) error {
```
pravinpushkar, line 26: This particular parameter is added to keep track of how many applications launched using the same `CLIPID`. This will be helpful in deciding when to kill the CLIPID.

cmd/stop.go
```diff
@@ -30,13 +33,39 @@ var StopCmd = &cobra.Command{
...
+				os.Exit(1)
+			}
+			runFilePath, err = getRunFilePath(runFilePath)
+			if err != nil {
+				print.FailureStatusEvent(os.Stderr, "Failed to get run file path: %v", err)
+				os.Exit(1)
+			}
+			err = executeStopWithRunFile(runFilePath)
+			if err != nil {
+				print.FailureStatusEvent(os.Stderr, "failed to stop Dapr and app processes: %s", err)
```
mukundansundar, line 57:
```suggestion
				print.FailureStatusEvent(os.Stderr, "Failed to stop Dapr and app processes: %s", err)
```

cmd/dapr.go
```diff
@@ -46,6 +46,12 @@ type daprVersion struct {
 	RuntimeVersion string `json:"Runtime version"`
 }
 
+type osType string
+
+const (
+	WINDOWS osType = "windows"
+)
```
shubham1172, line 53: nit: move it to a common place?

```suggestion
type osType string

const (
	WindowsOsType osType = "windows"
)
```

If it's only used for `runtime.GOOS`, we can also skip this change and directly use `"windows"` string.

cmd/run.go
```diff
@@ -503,8 +510,10 @@ func executeRun(runFilePath string, apps []runfileconfig.App) (bool, error) {
 		runStates = append(runStates, runState)
 
 		// Metadata API is only available if app has started listening to port, so wait for app to start before calling metadata API.
-		// The PID is put as 0, so as to not kill CLI process when any one of the apps is stopped.
-		_ = putCLIProcessIDInMeta(runState, 0)
+		_ = putCLIProcessIDInMeta(runState, os.Getpid())
+
+		// Update extended metadata with run file path.
+		_ = putRunFilePathInMeta(runState, runFilePath)
```
shubham1172, line 516: Are we discarding the errors intentionally? If yes, we can also discard the L.H.S. altogether.

cmd/run.go
```diff
@@ -909,6 +918,21 @@ func putAppCommandInMeta(runConfig standalone.RunConfig, runState *runExec.RunEx
 	return nil
 }
 
+// putRunFilePathInMeta puts the absolute path of run file in metadata so that it can be used by the CLI to stop all apps started by this run file.
+func putRunFilePathInMeta(runE *runExec.RunExec, runFilePath string) error {
```
shubham1172, line 922: `putAppCommandInMeta` still calls `runE` as `runState`, can you please update that?

cmd/stop.go
```diff
@@ -30,13 +33,39 @@ var StopCmd = &cobra.Command{
...
+dapr stop --run-file dapr.yaml
+
+# Stop multiple apps by providing a directory path containing the run config file(dapr.yaml)
+dapr stop --run-file /path/to/directory
 `,
 	Run: func(cmd *cobra.Command, args []string) {
+		var err error
+		if len(runFilePath) > 0 {
+			if runtime.GOOS == string(WINDOWS) {
+				print.FailureStatusEvent(os.Stderr, "The stop command with run file is not supported on Windows")
```
shubham1172, line 47:
```suggestion
				print.FailureStatusEvent(os.Stderr, "Stop command with run file is not supported on Windows")
```

shubham1172: Can we also add a tracking issue here if it's in the eventual roadmap?

pkg/standalone/list.go
```diff
@@ -156,3 +160,16 @@ func getIntArg(argMap map[string]string, argKey string, argDef int) int {
 	}
 	return argDef
 }
+
+// GetCLiPIDCountMap returns a map of cliPID to count of apps started with the same PPID.
+func GetCLiPIDCountMap() map[int]int {
```
shubham1172, line 165:
```suggestion
func GetCLIPIDCountMap() map[int]int {
```

pkg/standalone/list.go
```diff
@@ -156,3 +160,16 @@ func getIntArg(argMap map[string]string, argKey string, argDef int) int {
 	}
 	return argDef
 }
+
+// GetCLiPIDCountMap returns a map of cliPID to count of apps started with the same PPID.
```
shubham1172, line 164:
```suggestion
// GetCLIPIDCountMap returns a map of CLI PIDs to number of apps started with it.
```

cmd/stop.go
```diff
@@ -30,13 +33,39 @@ var StopCmd = &cobra.Command{
...
+				print.FailureStatusEvent(os.Stderr, "Failed to stop Dapr and app processes: %s", err)
+			} else {
+				print.SuccessStatusEvent(os.Stdout, "Dapr and app processes stopped successfully")
+			}
+			return
+		}
 		if stopAppID != "" {
 			args = append(args, stopAppID)
 		}
+		cliPIDToNoOfApps := standalone.GetCLiPIDCountMap()
```
shubham1172, line 66: Why not get this map directly in `standalone.stop` instead of `cmd/stop`?

cmd/stop.go
```diff
@@ -48,6 +77,15 @@ dapr stop --app-id <ID>
...
 	StopCmd.Flags().StringVarP(&stopAppID, "app-id", "a", "", "The application id to be stopped")
+	StopCmd.Flags().StringVarP(&runFilePath, "run-file", "f", "", "Path to the configuration file for the apps to run")
 	StopCmd.Flags().BoolP("help", "h", false, "Print this help message")
 	RootCmd.AddCommand(StopCmd)
 }
+
+func executeStopWithRunFile(runFilePath string) error {
+	absFilePath, err := filepath.Abs(runFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to get abosulte file path for %s: %w", runFilePath, err)
```
shubham1172, line 88:
```suggestion
		return fmt.Errorf("failed to get absolute file path for %s: %w", runFilePath, err)
```

pkg/standalone/stop.go
```diff
@@ -34,8 +34,9 @@ func Stop(appID string) error {
 			var pid string
 			// Kill the Daprd process if Daprd was started without CLI, otherwise
 			// kill the CLI process which also kills the associated Daprd process.
-			if a.CliPID == 0 {
+			if a.CliPID == 0 || cliPIDToNoOfApps[a.CliPID] > 1 {
 				pid = fmt.Sprintf("%v", a.DaprdPID)
+				cliPIDToNoOfApps[a.CliPID]--
```
shubham1172, line 39: If CliPID is 0, is this still valid?

shubham1172 (changes requested):

pkg/standalone/stop.go
pravinpushkar: right, it is valid.

cmd/stop.go
pravinpushkar: I placed here to handle the scenario like this `dapr stop app1 app2 app3`. We need to pass the original count. Relying on list command to return updated values will not work as it prcesses takes some time to get deleted.

cmd/stop.go
pravinpushkar: It will be tracked as per the feature impl for windows

cmd/dapr.go
pravinpushkar: it is being used in `runtime.GOOS` only but at some places it is being used as constant and and some places as string. So would like to keep as a const for cmd package only. Will make this unexported. let  me know is this does not sound ok.

cmd/run.go
pravinpushkar: yes. Actually no need to return an err from the method itself, because we just want to log and continue.

pkg/standalone/list.go
```diff
@@ -41,6 +41,7 @@ type ListOutput struct {
 	CliPID             int    `csv:"CLI PID"   json:"cliPid"             yaml:"cliPid"`
 	MaxRequestBodySize int    `csv:"-"         json:"maxRequestBodySize" yaml:"maxRequestBodySize"` // Additional field, not displayed in table.
 	HTTPReadBufferSize int    `csv:"-"         json:"httpReadBufferSize" yaml:"httpReadBufferSize"` // Additional field, not displayed in table.
+	RunFile            string `csv:"RUN_FILE"  json:"runFile"            yaml:"runFile"`
```
mukundansundar, line 44:
```suggestion
	RunTemplatePath            string `csv:"RUN_TEMPLATE_PATH"  json:"runTemplatePath"            yaml:"runTemplatePath"`
```

cmd/run.go
```diff
@@ -885,28 +894,37 @@ func killAppProcess(runE *runExec.RunExec) error {
...
+}
+
+// putRunFilePathInMeta puts the absolute path of run file in metadata so that it can be used by the CLI to stop all apps started by this run file.
+func putRunFilePathInMeta(runE *runExec.RunExec, runFilePath string) {
+	runFilePath, err := filepath.Abs(runFilePath)
+	if err != nil {
+		print.StatusEvent(runE.DaprCMD.OutputWriter, print.LogWarning, "Could not get absolute path for run file: %s", err.Error())
+		return
+	}
+	err = metadata.Put(runE.DaprHTTPPort, "runFile", runFilePath, runE.AppID, unixDomainSocket)
```
mukundansundar, line 924:
```suggestion
	err = metadata.Put(runE.DaprHTTPPort, "runTemplatePath", runFilePath, runE.AppID, unixDomainSocket)
```

mukundansundar (commented):
@pravinpushkar as a follow up to this PR and #1170 , need to add E2E for the stop scenario as well as list.

cmd/stop.go
```diff
@@ -48,6 +77,15 @@ dapr stop --app-id <ID>
 
 func init() {
 	StopCmd.Flags().StringVarP(&stopAppID, "app-id", "a", "", "The application id to be stopped")
+	StopCmd.Flags().StringVarP(&runFilePath, "run-file", "f", "", "Path to the configuration file for the apps to run")
```
mukundansundar, line 80:
```suggestion
	StopCmd.Flags().StringVarP(&runFilePath, "run-file", "f", "", "Path to the run template file for the list of apps to stop.")
```

cmd/run.go
```diff
@@ -18,6 +18,7 @@ import (
 	"fmt"
```
mukundansundar, line 18: Change `RunCmd.Flags().StringVarP(&runFilePath, "run-file", "f", "", "Path to the configuration file for the apps to run")`
```suggestion
	RunCmd.Flags().StringVarP(&runFilePath, "run-file", "f", "", "Path to the run template file for the list of apps to run")
```

pravinpushkar:
> @pravinpushkar as a follow up to this PR and #1170 , need to add E2E for the stop scenario as well as list.

yes, thats the plan.

pkg/standalone/stop.go
mukundansundar, line 39: this line is not needed right? 
As the command is run only once and then the `dapr stop` command exits with the return statement below.

pkg/standalone/list.go
```diff
@@ -156,3 +160,16 @@ func getIntArg(argMap map[string]string, argKey string, argDef int) int {
 	}
 	return argDef
 }
+
+// GetCLIPIDCountMap returns a map of CLI PIDs to number of apps started with it.
+func GetCLIPIDCountMap() map[int]int {
+	cliPIDCountMap := make(map[int]int)
```
mukundansundar, line 166: move to next line and change to 
```suggestion
	cliPIDCountMap := make(map[int]int, len(apps))
```

pkg/standalone/list.go
```diff
@@ -156,3 +160,16 @@ func getIntArg(argMap map[string]string, argKey string, argDef int) int {
...
+	apps, err := List()
+	if err != nil {
+		return nil
```
mukundansundar, line 169: why is the error swallowed here?

pkg/standalone/stop.go
pravinpushkar: it can be triggered like this -  dapr stop app1 app2 app3 (all 3 apps are started by dapr run -f )

pkg/standalone/list.go
pravinpushkar: this error was not that useful here because we do not want the stop to exit here and again the same logic to get the list of apps was there in `Stop` method, which would have handled the failure case.

So, refactored `Stop` method a little to avoid the duplication of logic. Let me know your thoughts on this.

cmd/stop.go
```diff
@@ -48,6 +82,15 @@ dapr stop --app-id <ID>
...
 	StopCmd.Flags().BoolP("help", "h", false, "Print this help message")
 	RootCmd.AddCommand(StopCmd)
 }
+
+func executeStopWithRunFile(runFilePath string) error {
+	absFilePath, err := filepath.Abs(runFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to get absolute file path for %s: %w", runFilePath, err)
+	}
+	return standalone.StopAppsWithRunFile(absFilePath)
```
mukundansundar, line 95: can pass in the list of apps from the run function above. 
Query List() only once?

pkg/standalone/list.go
mukundansundar: Ok ... but have the List() queried only once then and use if for either normal stop or stop run template ...

cmd/stop.go
pravinpushkar: it is now querying only once.
for `dapr stop -f ` -> see method `StopAppsWithRunFile`
for normal `dapr stop appID` -> see line no 66 in cmd/stop.go.

mukundansundar (approved):
lgtm

PR #1175, opened 2023-01-20
fix bug in precendence log wrt --dapr-path flag

pravinpushkar:

# Description
 Fixed an issue wrt `--dapr-path` flag. It does not create a `.dapr folder` in the given install path.
related to https://github.com/dapr/cli/pull/1158


mukundansundar:
@pravinpushkar can you provide more details about this issue?

pravinpushkar:
> @pravinpushkar can you provide more details about this issue?

Initially my understanding was  -> `dapr init --dapr-path installfolder` will create a `.dapr` directory inside `installfolder`. Which is not the case. It creates bin dir, components dir and config.yaml  directly under `installfolder`.

Actually the PR was according to the correct logic only but somehow I changed this logic y'day.

mukundansundar (approved):

PR #1181, opened 2023-01-25
dapr stop -f fix and E2E tests for dapr stop and list

pravinpushkar:

# Description

- Modify stop logic to kill process group instead of `CLIPID`
- This will stop all child processes including nested children PID.
- Added E2E for dapr stop -f 

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/ossyscall.go
+++ b/cmd/ossyscall.go
@@ -20,8 +20,17 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+
+	"github.com/dapr/cli/pkg/print"
 )
 
 func setupShutdownNotify(sigCh chan os.Signal) {
 	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
+}
+
+// createProcessGroupID creates a process group id for the current process.
+func createProcessGroupID() {
+	if err := syscall.Setpgid(0, 0); err != nil {
+		print.WarningStatusEvent(os.Stdout, "Failed to create process group id: %s", err.Error())
+	}
 }
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -22,7 +22,6 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -459,10 +458,7 @@ func executeRun(runFilePath string, apps []runfileconfig.App) (bool, error) {
 
 	runStates := make([]*runExec.RunExec, 0, len(apps))
 	// Create a process group id for all the apps and daprd processes started.
-	err := syscall.Setpgid(0, 0)
-	if err != nil {
-		print.WarningStatusEvent(os.Stdout, "Failed to create process group id: %s", err.Error())
-	}
+	createProcessGroupID()
 	for _, app := range apps {
 		print.StatusEvent(os.Stdout, print.LogInfo, "Validating config and starting app %q", app.RunConfig.AppID)
 		// Set defaults if zero value provided in config yaml.
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -460,7 +460,7 @@ func executeRun(runFilePath string, apps []runfileconfig.App) (bool, error) {
 	runStates := make([]*runExec.RunExec, 0, len(apps))
 
 	// Creates a separate process group ID for current process i.e. "dapr run -f".
-	// All the subprocess and their grand children inherits this PGID.
+	// All the subprocess and their grandchildren inherit this PGID.
 	// This is done to provide a better grouping, which can be used to control all the proceses started by "dapr run -f".
 	daprsyscall.CreateProcessGroupID()
 
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -32,7 +32,7 @@ import (
 	runExec "github.com/dapr/cli/pkg/runexec"
 	"github.com/dapr/cli/pkg/standalone"
 	"github.com/dapr/cli/pkg/standalone/runfileconfig"
-	dapr_syscall "github.com/dapr/cli/pkg/syscall"
+	daprsyscall "github.com/dapr/cli/pkg/syscall"
 	"github.com/dapr/cli/utils"
 )
 
@@ -192,7 +192,7 @@ dapr run --run-file /path/to/directory
 		// TODO: In future release replace following logic with the refactored functions seen below.
 
 		sigCh := make(chan os.Signal, 1)
-		dapr_syscall.SetupShutdownNotify(sigCh)
+		daprsyscall.SetupShutdownNotify(sigCh)
 
 		daprRunning := make(chan bool, 1)
 		appRunning := make(chan bool, 1)
@@ -455,14 +455,14 @@ func executeRun(runFilePath string, apps []runfileconfig.App) (bool, error) {
 
 	// setup shutdown notify channel.
 	sigCh := make(chan os.Signal, 1)
-	dapr_syscall.SetupShutdownNotify(sigCh)
+	daprsyscall.SetupShutdownNotify(sigCh)
 
 	runStates := make([]*runExec.RunExec, 0, len(apps))
 
 	// Creates a separate process group ID for current process i.e. "dapr run -f".
 	// All the subprocess and their grand children inherits this PGID.
 	// This is done to provide a better grouping, which can be used to control all the proceses started by "dapr run -f".
-	dapr_syscall.CreateProcessGroupID()
+	daprsyscall.CreateProcessGroupID()
 
 	for _, app := range apps {
 		print.StatusEvent(os.Stdout, print.LogInfo, "Validating config and starting app %q", app.RunConfig.AppID)
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -32,6 +32,7 @@ import (
 	runExec "github.com/dapr/cli/pkg/runexec"
 	"github.com/dapr/cli/pkg/standalone"
 	"github.com/dapr/cli/pkg/standalone/runfileconfig"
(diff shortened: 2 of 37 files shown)


tests/e2e/standalone/commands.go
```diff
@@ -137,6 +137,11 @@ func cmdStop(appId string, args ...string) (string, error) {
 	return spawn.Command(common.GetDaprPath(), stopArgs...)
 }
 
+func cmdStopWithRunTemplate(runTemplateFile string, args ...string) (string, error) {
+	stopArgs := append([]string{"stop", "--log-as-json", "-f", runTemplateFile}, args...)
+	return spawn.Command(common.GetDaprPath(), stopArgs...)
+}
+
```
mukundansundar, line 144: refactor `cmdStop` and `cmdStopRunTemplate` into a single function and then have both `cmdStop` and `cmdStopRunTemplate` use that refactored function ...

pkg/standalone/stop.go
```diff
@@ -52,7 +53,15 @@ func StopAppsWithRunFile(runTemplatePath string) error {
...
-			_, err := utils.RunCmdAndWait("kill", fmt.Sprintf("%v", a.CliPID))
+			// Get the process group id of the CLI process.
+			pgid, err := syscall.Getpgid(a.CliPID)
+			if err != nil {
+				// Fall back to cliPID if pgid is not available.
+				_, err = utils.RunCmdAndWait("kill", fmt.Sprintf("%v", a.CliPID))
+				return err
+			}
+			// Kill the whole process group.
+			_, err = utils.RunCmdAndWait("kill", "-SIGTERM", "--", fmt.Sprintf("-%v", pgid))
```
mukundansundar, line 64: Do not use SIGTERM, use SIGINT

tests/e2e/standalone/utils.go
```diff
@@ -156,3 +157,7 @@ func containerRuntime() string {
 	}
 	return ""
 }
+
+func daprStop(stopArgs ...string) (string, error) {
+	return spawn.Command(common.GetDaprPath(), stopArgs...)
+}
```
mukundansundar, line 163: move this function into the commands.go file itself.

cmd/ossyscall.go
```diff
@@ -20,8 +20,17 @@ import (
...
 	"syscall"
+
+	"github.com/dapr/cli/pkg/print"
 )
 
 func setupShutdownNotify(sigCh chan os.Signal) {
 	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
 }
+
+// createProcessGroupID creates a process group id for the current process.
```
shubham1172, line 31: Since this is not a command, can we move these to `pkg/syscall/syscall.go` and `pkg/syscall/syscall_windows.go`?

shubham1172, line 31:
```suggestion
// createProcessGroupID creates a process group ID for the current process.
```

Also please explain what it does and why.

cmd/run.go
```diff
@@ -457,6 +457,8 @@ func executeRun(runFilePath string, apps []runfileconfig.App) (bool, error) {
 	setupShutdownNotify(sigCh)
 
 	runStates := make([]*runExec.RunExec, 0, len(apps))
+	// Create a process group id for all the apps and daprd processes started.
```
shubham1172, line 460: Please explain why.

shubham1172: Also does it not create the Process Group ID for the current process only?

cmd/ossyscall.go
```diff
@@ -20,8 +20,17 @@ import (
...
+func createProcessGroupID() {
+	if err := syscall.Setpgid(0, 0); err != nil {
```
shubham1172, line 33: Can point to the manpage and explain the significance of `0, 0` params.

cmd/ossyscall.go
pravinpushkar: right, we can move. @mukundansundar any issue with this ?

pkg/syscall/syscall_windows.go
```diff
@@ -41,7 +41,7 @@ func setupShutdownNotify(sigCh chan os.Signal) {
 	}()
 }
 
-// createProcessGroupID creates a process group id for the current process.
+// createProcessGroupID creates a process group ID for the current process.
```
shubham1172, line 44: Methods in this file should also be exported?

cmd/run.go
pravinpushkar: have put better a description now. All children inherits the parent process group ID by default, until changed by following two - 
1. `Setpgid`
2. `&syscall.SysProcAttr{Setpgid: true}`

We are using the 1st option to our use and setting a different process group ID for `dapr run -f`. This will make all the children to inherit the same process group.

option 2 can also be used to create a different process group ID but it is not useful in our case as we want to set the process group ID for the calling process.

cmd/ossyscall.go
mukundansundar: No issues

pkg/standalone/stop.go
```diff
@@ -52,7 +53,15 @@ func StopAppsWithRunFile(runTemplatePath string) error {
...
-			_, err := utils.RunCmdAndWait("kill", fmt.Sprintf("%v", a.CliPID))
+			// Get the process group id of the CLI process.
+			pgid, err := syscall.Getpgid(a.CliPID)
+			if err != nil {
+				// Fall back to cliPID if pgid is not available.
+				_, err = utils.RunCmdAndWait("kill", fmt.Sprintf("%v", a.CliPID))
+				return err
+			}
+			// Kill the whole process group.
+			_, err = utils.RunCmdAndWait("kill", "-SIGINT", "--", fmt.Sprintf("-%v", pgid))
```
mukundansundar, line 64: should `syscall.Kill(-pgid, syscall.SIGINT)` be used instead of this?

pravinpushkar: changed to `syscall.Kill`. Seems both same. Did not notice any difference regarding `cmd.start` and `cmd.wait` in `RunCmdAndWait `.

pkg/syscall/syscall_windows.go
```diff
@@ -40,3 +40,9 @@ func setupShutdownNotify(sigCh chan os.Signal) {
 		sigCh <- os.Interrupt
 	}()
 }
+
+// CreateProcessGroupID creates a process group ID for the current process.
```
shubham1172, line 44: Also want to add the man link here?

shubham1172, line 44: Probably you can add the comment here itself that it's a no-op in Windows

cmd/run.go
```diff
@@ -454,9 +455,15 @@ func executeRun(runFilePath string, apps []runfileconfig.App) (bool, error) {
 
 	// setup shutdown notify channel.
 	sigCh := make(chan os.Signal, 1)
-	setupShutdownNotify(sigCh)
+	daprsyscall.SetupShutdownNotify(sigCh)
 
 	runStates := make([]*runExec.RunExec, 0, len(apps))
+
+	// Creates a separate process group ID for current process i.e. "dapr run -f".
+	// All the subprocess and their grand children inherits this PGID.
```
shubham1172, line 463:
```suggestion
	// All the subprocess and their grand children inherit this PGID.
```

cmd/ossyscall.go
shubham1172: @pravinpushkar can you please add this?

mukundansundar:
> @pravinpushkar can you please add this?

What should be added? @shubham1172

shubham1172:
@mukundansundar looks like a GH bug, I added the comment to my original comment thread 😛 

image

pkg/syscall/syscall.go
```diff
@@ -14,14 +14,24 @@ See the License for the specific language governing permissions and
...
+	"github.com/dapr/cli/pkg/print"
 )
 
-func setupShutdownNotify(sigCh chan os.Signal) {
+func SetupShutdownNotify(sigCh chan os.Signal) {
 	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
 }
+
+// CreateProcessGroupID creates a process group ID for the current process.
+// Reference - https://man7.org/linux/man-pages/man2/setpgid.2.html.
```
pravinpushkar, line 32: see here @shubham1172

cmd/ossyscall.go
pravinpushkar: added

pkg/syscall/syscall.go
```diff
@@ -31,6 +31,11 @@ func SetupShutdownNotify(sigCh chan os.Signal) {
 // CreateProcessGroupID creates a process group ID for the current process.
 // Reference - https://man7.org/linux/man-pages/man2/setpgid.2.html.
 func CreateProcessGroupID() {
+	// Below is some excerpt from the above link Setpgid() -
+	// setpgid() sets the PGID of the process specified by pid to pgid.
+	// If pid is zero, then the process ID of the calling process is
+	// used.  If pgid is zero, then the PGID of the process specified by
+	// pid is made the same as its process ID.
```
shubham1172, line 38: This is from the manpage, but rather it should be in context of the Dapr CLI.

shubham1172 (approved):
One small comment but not blocking anything, LGTM.

pkg/syscall/syscall.go
pravinpushkar: just thinking what to put because putting in context of dapr run won't be right as it is now kind of utility can be extended or modified based on further usecases.

we have put some information wrt `dapr run -f` from where this gets invoked.

mukundansundar (approved):

Issue #1188, opened 2023-01-27
dashboard fails to start if port 8080 in use


ryan-z:
## Expected Behavior

An error message saying this port was already in use, and maybe suggest the command to resolve it (`dapr dashboard -p 9999`)

## Actual Behavior

It reports that it has started running, but then confusingly questions if dapr is installed:
```
Dapr Dashboard running on http://localhost:8080
❌  Dapr dashboard not found. Is Dapr installed?
```

## Steps to Reproduce the Problem

1. Bind anything to port 8080 (e.g. `npx serve -l 8080`)
2. Run `dapr dashboard`
3. Observe the confusing error message

## Release Note

RELEASE NOTE: **FIX** Error messaging in dashboard command.

philliphoff:
Nice to see this was fixed!  The random port selection will help clean up logic in the VS Code extension when starting the dashboard.

PR #1189, opened 2023-01-28
Fix dashboard port conflict message and allow starting on random port

pravinpushkar:

## Issue reference

Please reference the issue this PR will close: #1188

### DIFF
--- a/cmd/dashboard.go
+++ b/cmd/dashboard.go
@@ -65,6 +65,9 @@ dapr dashboard
 # Start dashboard locally in a specified port 
 dapr dashboard -p 9999
 
+# Start dashboard locally on a random port which is free.
+dapr dashboard -p 0
+
 # Port forward to dashboard in Kubernetes 
 dapr dashboard -k 
 
@@ -96,6 +99,12 @@ dapr dashboard -k -p 0
 
 		if dashboardLocalPort < 0 {
 			print.FailureStatusEvent(os.Stderr, "Invalid port: %v", dashboardLocalPort)
+			os.Exit(1)
+		}
+
+		if err := utils.CheckIfPortAvailable(dashboardLocalPort); err != nil {
+			print.FailureStatusEvent(os.Stderr, "Please select a different port with %q flag: %s", "-p", err)
+			print.InfoStatusEvent(os.Stdout, "You can also use port 0 to select a random free port.")
 			os.Exit(1)
 		}
 
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -266,6 +266,15 @@ func IsAddressLegal(address string) bool {
 	return isLegal
 }
 
+func CheckIfPortAvailable(port int) error {
+	ln, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+	if err != nil {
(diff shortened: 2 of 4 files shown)


cmd/dashboard.go
```diff
@@ -99,6 +102,12 @@ dapr dashboard -k -p 0
 			os.Exit(1)
 		}
 
+		if err := utils.CheckIfPortAvailable(dashboardLocalPort); err != nil {
+			print.FailureStatusEvent(os.Stderr, "Please select a different port with %q flag: %s", "-p", err)
```
shubham1172, line 106: Will this error message inform the user that the port is busy?

shubham1172: Do we need "%q" with "-p", can we use the flag in string directly?

utils/utils.go
```diff
@@ -266,6 +266,15 @@ func IsAddressLegal(address string) bool {
 	return isLegal
 }
 
+func CheckIfPortAvailable(port int) error {
```
shubham1172, line 269: Please add doc comment, can we also add tests?

shubham1172 (changes requested):

utils/utils.go
pravinpushkar: I think for test we will have to depend on some random port which might be occupied in some cases and test will be error prone. thoughts?
Also, this method just calls the `net.Listen` and does not have much business logic.

cmd/dashboard.go
pravinpushkar: > Will this error message inform the user that the port is busy?

yes.

> Do we need "%q" with "-p", can we use the flag in string directly?

yes, we will have to use %q here.

utils/utils.go
shubham1172: We can `net.Listen` on the same port before calling this function to test. The port can be fixed or generated from freeport pkg. 

Ok with skipping the test, but it's good to have. /cc @mukundansundar

shubham1172 (approved):

utils/utils.go
shubham1172: As discussed, this test can be tracked separately.

mukundansundar (approved):

pravinpushkar:
Below is the output - 
@mukundansundar @shubham1172 

#### Valid port/ default port is unoccupied - 

➜  cli git:(dashboard) ldapr dashboard
Dapr Dashboard running on http://localhost:8080

#### When default port is occupied - 
➜  ~ ldapr dashboard
❌  Please select a different port with "-p" flag: listen tcp :8080: bind: address already in use
ℹ️  You can also use port 0 to select a random free port.

#### When passed port is 0 - 
➜  ~ ldapr dashboard -p 0
Dapr Dashboard running on http://localhost:50190

utils/utils.go
pravinpushkar: There is no E2E for dashboard. We can try to add one and cover this there.
tracking issue - https://github.com/dapr/cli/issues/1192

shubham1172: Thanks @pravinpushkar

Issue #698, opened 2021-04-23
Multiple instances of service for load balancing


satishviswanathan:
## Ask your question here
https://docs.dapr.io/developing-applications/building-blocks/service-invocation/service-invocation-overview/

Following this link I'm trying to run multiple instances of cart service. i.e when I start using the dapr I'm trying to start multiple instances using the same app id "cart" which results in the following exception.

**_dapr with ID cart is already running_**

Please let me know if my understanding is correct.

haohaifeng002:
a similar problem like https://github.com/dapr/dapr/issues/625

mukundansundar:
@yaron2 Can you add some thoughts on this? I believe that this is currently not supported by the CLI. Is that correct?

PR #1197, opened 2023-02-01
Make app id unique across apps started by dapr run

pravinpushkar:

# Description

- Make appID unique across apps started by dapr run. Not uniqueness was introduced by mistake in rel-1.6. There is a diff issues opened for supporting this correctly - #698. 
- dapr run -f will exit completely if AppIDs are not unique in `dapr.yaml`


pkg/standalone/run.go
```diff
@@ -84,6 +84,15 @@ func (meta *DaprMeta) newAppID() string {
 	}
 }
 
+func (config *RunConfig) validateAppID(meta *DaprMeta) error {
+	if config.AppID == "" {
+		config.AppID = meta.newAppID()
+	} else if meta.idExists(config.AppID) {
+		return fmt.Errorf("invalid configuration for AppID. App id %q is not available", config.AppID)
```
shubham1172, line 91:
```suggestion
		return fmt.Errorf("invalid configuration for appID. App ID %q is already in use", config.AppID)
```

shubham1172 (dismissed):
Small nit

mukundansundar (changes requested):
add a test with normal run as well, running an app with app id and the next run `executeWhileRunning` should fail.

Change it to point to `release-1.10`

mukundansundar (dismissed):
lgtm

shubham1172 (approved):

mukundansundar:
AppID uniqueness was unintentionally removed in 1.6, and there is no complete support for it yet. Will work on https://github.com/dapr/cli/issues/698 to allow appID duplication.

mukundansundar (approved):

shubham1172 (dismissed):

mukundansundar (approved):

shubham1172 (approved):

artursouza (approved):

pravinpushkar:
one suggested improvement that can be tried for 1.11 until #698 is implemented - 

> You are checking based on the list of processes. It is not guaranteed but it is better than we have now. This can also be achieved by a file lock, but not sure how to do that on Windows. It can be in 1.11 since it would avoid race conditions.

mukundansundar:
@yaron2 had some concerns with this PR breaking some of the users and also that this unintentional change has been there from 1.6. 

@yaron2 thoughts on this?

yaron2:
> @yaron2 had some concerns with this PR breaking some of the users and also that this unintentional change has been there from 1.6.
> 
> @yaron2 thoughts on this?

Technically, it's not forbidden to run two instances of Dapr with the same app Id - this is what happens naturally on K8s with multiple pods all having the same app Id. So unless there is something in the local dev experience that is degraded by having this, maybe it's best to leave it as is, because:

1. It has been allowed in the CLI for the last 3 releases
2. It allows developers to easily test things like two consumers in a consumer group with pub/sub, and multiple consumers of a distributed Lock
3. There may have been users who use the CLI to bootstrap non containerized envs with Dapr who rely on this, although unlikely.

I feel like we can go either way, but would like to revert this based on something concrete.

pravinpushkar:
@yaron2 
> Technically, it's not forbidden to run two instances of Dapr with the same app Id - this is what happens naturally on K8s with multiple pods all having the same app Id. So unless there is something in the local dev experience that is degraded by having this, maybe it's best to leave it as is

If we are going to have this as is then we have to correct the `dapr stop` behaviour.
Currently `dapr stop appID` will stop any one of the many instances of apps having same `appID`.
`dapr stop` iterates through the apps list obtained via `dapr list` and exist after getting the 1st match for the provided app ID. 

And similar behaviour happen for `dapr run -f`.

mukundansundar:
The behavior of the following CLI commands also will be affected right? `dapr stop`, `dapr publish`, `dapr invoke` etc. 
Since all of them use `list` to list the apps and based on the `appID` select a particular app to perform the function of invoke, publish or to stop

mukundansundar:
@dapr/maintainers-cli Based on the discussions that we have had offline, this change will not be done.

PR #1237, opened 2023-02-17
fix vulnerable dependencies

pravinpushkar:
# Description

Related to below vulnerability failure - 

https://github.com/dapr/cli/actions/runs/4201647149/jobs/7288859657


go.mod
```diff
@@ -228,8 +228,6 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v14.2.0+incompatible
-
-	github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
```
pravinpushkar, line 232: Removing this replace. Looks like it is related to helm dependency that has been upgraded recently.
@artursouza @mukundansundar  Do you see any issue with this?

mukundansundar: LGTM. 

@pravinpushkar Can you see if the docker client can be upgraded also? 
The replace in the line below.

mukundansundar (approved):

go.mod
mukundansundar: Can be done as a separate PR.

pravinpushkar: I will check that..

Issue #1245, opened 2023-02-23
dapr run with resourcesPath: ~/some/folder  doesn't resolve ~ to the user's home folder


sethwenzel:
## Expected Behavior

```yaml
version: 1
common:
  resourcesPath: ~/some/external/folder/path
apps:
  ....
```

should look for the components folder under the user's home folder -- `/home/<user>/some/external/folder/path`

## Actual Behavior

`resourcesPath: ~/some/folder` will cause the error:

> ❌  Error getting apps from config file: error in getting the file info for /home/\<user\>/projects/dapr-project/\~/some/folder: stat /home/\<user\>/projects/dapr-project/\~/some/folder: no such file or directory

## Steps to Reproduce the Problem

use "~" in the `resourcesPath` in a dapr.yaml file

## Release Note

RELEASE NOTE:

pravinpushkar:
It is a bug. 
Currently `resourcesPath` can be given in absolute path or relative path (relative to the yaml file).
Paths starting with `~` needs to be handled.

Thanks for reporting.

sethwenzel:
`$HOME` appears to not work either.  I tried both `$HOME/some/folder` and `/$HOME/some/folder`

tharun634:
Hi,
I want to work on this. I was thinking of introducing a util function that would take in the file path and resolve the `~/` or `$HOME/` prefix, should it exist, to the absolute path of the user's home directory and use this util function at the following places:
https://github.com/dapr/cli/blob/a6f8c7694d3bd79f1ff2d515330b71ed42e28ec1/pkg/standalone/runfileconfig/run_file_config_parser.go#L153-L157
https://github.com/dapr/cli/blob/a6f8c7694d3bd79f1ff2d515330b71ed42e28ec1/pkg/standalone/runfileconfig/run_file_config_parser.go#L209-L211
https://github.com/dapr/cli/blob/a6f8c7694d3bd79f1ff2d515330b71ed42e28ec1/pkg/standalone/runfileconfig/run_file_config_parser.go#L230-L232

Please do let me know in case I am missing something, and I'd love to raise a draft [WIP] PR with the above changes to receive feedback

mukundansundar:
@tharun634 please feel free to create a PR!

pravinpushkar:
> `$HOME` appears to not work either. I tried both `$HOME/some/folder` and `/$HOME/some/folder`

I meant the actual value of `$HOME` variable.  

[...]

PR #1247, opened 2023-02-26
handle home directory prefix

tharun634:
# Description

Introduced a util function that would take in the file path and resolve the `~/`  prefix, should it exist, to the absolute path of the user's home directory.

## Issue reference

Please reference the issue this PR will close: #1245 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


utils/utils.go
```diff
@@ -350,6 +350,26 @@ func GetAbsPath(baseDir, path string) string {
 	return absPath
 }
 
+// ResolveHomeDir resolves prefix of the given path, if present, to the user's home directory and returns it
+func ResolveHomeDir(filePath string) (string, error) {
+	if filePath == "" {
+		return "", nil
+	}
+	homeDirPrefixes := []string{"~/", "$HOME/"}
```
pravinpushkar, line 358: AFAIK in windows this variable `$HOME` is called `Userprofile` and this logic might not be able to handle it. Should we avoid this part ?

pravinpushkar:
@tharun634 Please fix the lint error to check the behaviour of your UTs on supported OSs.

utils/utils.go
tharun634: Yes you are correct, I'll remove support for `$HOME` and support only `~/`

pravinpushkar (approved):

utils/utils.go
```diff
@@ -350,6 +350,22 @@ func GetAbsPath(baseDir, path string) string {
 	return absPath
 }
 
+// ResolveHomeDir resolves prefix of the given path, if present, to the user's home directory and returns it.
+func ResolveHomeDir(filePath string) (string, error) {
+	if filePath == "" {
+		return "", nil
+	}
+	if homeDirPrefix := "~/"; strings.HasPrefix(filePath, homeDirPrefix) {
```
shubham1172, line 358: Can `homeDirPrefix := "~/"` be a constant?

mukundansundar: also is there a way to check this only for linux and posix like systems? As in Windows this will not be the case.

shubham1172: @mukundansundar on Windows also users can refer to %USERPROFILE%, do you want to not consider "~/" syntax? I propose to keep it the same as Linux and Mac.

image

mukundansundar:
@tharun634 on an unrelated note, can you stop the automatic action run in your fork? I am getting notifications of the failures of those runs for example dapr-bot-schedule etc.....

mukundansundar (approved):

Issue #1209, opened 2023-02-06
Dapr init logging should show the correct container runtime


shubham1172:
## Expected Behavior

When initializing `dapr` with Podman, if Podman is not running, console message should say "could not connect to Podman. Podman may not be installed or running"

## Actual Behavior

Console message still mentions Docker.

## Steps to Reproduce the Problem

```
PS C:\Users\shubham1172> dapr init --container-runtime podman
Making the jump to hyperspace...
could not connect to Docker. Docker may not be installed or running
```

## Fix

This code should be updated to use the actual container runtime.
https://github.com/dapr/cli/blob/55bc454ad970b1c4ac0e50b7f122ad8418ea8499/pkg/standalone/standalone.go#L163-L168

## Release Note

RELEASE NOTE: **UPDATE** Log message to reflect the container runtime

ThorstenHans:
Can I take this one?

PR #1250, opened 2023-02-28
Show correct container runtime in init logging

ThorstenHans:
# Description

Init logging now logs using container runtime specified by the user, instead of always logging "Docker".

## Issue reference

Please reference the issue this PR will close: #1209

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -32,6 +32,8 @@ import (
 	"time"
 
 	"github.com/fatih/color"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	"gopkg.in/yaml.v2"
 
 	"github.com/dapr/cli/pkg/print"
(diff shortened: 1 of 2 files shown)


pkg/standalone/standalone.go
```diff
@@ -164,7 +166,8 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 		// If --slim installation is not requested, check if docker is installed.
 		conatinerRuntimeAvailable := utils.IsDockerInstalled() || utils.IsPodmanInstalled()
 		if !conatinerRuntimeAvailable {
-			return errors.New("could not connect to Docker. Docker may not be installed or running")
+			caser := cases.Title(language.English)
+			return fmt.Errorf("could not connect to %s. %s may not be installed or running", caser.String(containerRuntime), caser.String(containerRuntime))
```
shubham1172, line 170: Do we need this? Based on the docs it seems that the two lines convert `containerRuntime` to Title-Case in English. It should be ok to display them as-is without adding two experimental libraries.

ThorstenHans: @shubham1172 I can refactor it easily to print container runtime values as they are. However `"golang.org/x/text` is already a dependency of the CLI itself

shubham1172 (approved):

pkg/standalone/standalone_test.go
```diff
@@ -17,6 +17,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/dapr/cli/utils"
```
shubham1172, line 20: Can you please fix the lint?

ThorstenHans: sure

shubham1172 (approved):

pravinpushkar (approved):

mukundansundar (approved):
lgtm

Issue #1210, opened 2023-02-06
Define `container-runtime` via environment variable


artursouza:
## Describe the proposal

Instead of requiring user to pass `--container-runtime` when using Podman, an environment variable override (`DAPR_CONTAINER_RUNTIME`) should be enough. This should allow Quickstarts to show the same commands regardless of user using Docker or Podman, while having automated validation to be done with Podman.

This proposal is to allow offer `DAPR_CONTAINER_RUNTIME` env in addition to `--container-runtime` arg.

## Release Note

RELEASE NOTE: **ADDED** `DAPR_CONTAINER_RUNTIME` env in addition to `--container-runtime` arg.

PR #1251, opened 2023-02-28
Allow setting container-runtime via env var for `dapr init`

ThorstenHans:
# Description

The container runtime used by `dapr init` reads potential value from `DAPR_CONTAINER_RUNTIME` env var. 

## Issue reference

Please reference the issue this PR will close: #1210

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


cmd/init.go
```diff
@@ -179,6 +179,13 @@ func init() {
 		defaultDashboardVersion = dashboardVersionEnv
 	}
 
+	defaultContainerRuntime := "docker"
```
pravinpushkar, line 182: Constant present in utils/utils.go might be used.
Normally we try to prefix env variables with `dapr`. So can we keep something like `dapr_container_runtime` ?

ThorstenHans: @pravinpushkar you mean the local variable name which is now called `defaultContainerRuntime` should be renamed? Because the env var taken into context is called `DAPR_CONTAINER_RUNTIME`. 

I'll update the PR to pull "docker" from `utils/utils.go`

cmd/init.go
```diff
@@ -179,6 +179,13 @@ func init() {
 		defaultDashboardVersion = dashboardVersionEnv
 	}
 
+	defaultContainerRuntime := string(utils.DOCKER)
+	viper.BindEnv("container_runtime_override", "DAPR_CONTAINER_RUNTIME")
```
mukundansundar, line 183: I think this is not needed. These lines https://github.com/dapr/cli/blob/master/cmd/dapr.go#L100-L102 should already have a bound string, so getting that string should be good.
```
viper.GetString("container-runtime")
```

mukundansundar (changes requested):
@ThorstenHans can you please check if the automatic env bind works as expected and use that var string instead?

cmd/init.go
ThorstenHans: > @ThorstenHans can you please check if the automatic env bind works as expected and use that var string instead?

Actually, it does not work. The `init` func in `cmd/init.go` is invoked before `initConfig` from `dapr/dapr.go` is called as part of `cobra.OnInitialize`.

mukundansundar: You can resolve it in the `Run` function of the command as `OnInitialize` is called before `Run` is called.
For example see how `network` value is got in the same `cmd/init.go` file.

ThorstenHans: That's clear. 

However, I was a bit worried about other configuration data loaded from env vars. So, should I relocate loading data from other env vars (`DAPR_RUNTIME_VERSION`, `DAPR_DASHBOARD_VERSION`) to `Run` too?

mukundansundar: I agree, they can be moved to Run as well.

ThorstenHans: @mukundansundar I moved loading configuration values from 

1. default values
2. env vars
3. flags

into a dedicated func `getConfigurationValue` and hooked it up in `PreRun` as there was already some code that is relied to configuration data.

pravinpushkar:
@ThorstenHans Please resolve the conflicts.

ThorstenHans:
> @ThorstenHans Please resolve the conflicts.

👍🏼 done

mukundansundar (changes requested):
@ThorstenHans these changes seem to be breaking the tests. Can you change the way the overrides are bind to be similar to viper.BindEnv("runtime_version_override", "DAPR_RUNTIME_VERSION")
 	runtimeVersionEnv := viper.GetString("runtime_version_override")
 	if runtimeVersionEnv != "" {
 	}

mukundansundar (approved):

PR #1255, opened 2023-03-07
output help information when dapr has no parameters

tianyax:
# Description

Fixed the problem that the output of the dapr did not match the documentation after installation.

## Issue reference

Please reference the issue this PR will close: #1254

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/cmd/dapr.go
+++ b/cmd/dapr.go
@@ -39,9 +39,11 @@ var RootCmd = &cobra.Command{
 
 ===============================
 Distributed Application Runtime`,
-	Run: func(cmd *cobra.Command, args []string) {
+	Run: func(cmd *cobra.Command, _ []string) {
 		if versionFlag {
 			printVersion()
+		} else {
+			cmd.Help()
 		}
 	},
 }


shubham1172 (changes requested):
The `--version` flag should be the same as `dapr version` IMO. It is meaningful to provide both CLI and installed runtime's version.

cmd/dapr.go
```diff
@@ -39,11 +39,7 @@ var RootCmd = &cobra.Command{
 
 ===============================
 Distributed Application Runtime`,
-	Run: func(cmd *cobra.Command, args []string) {
-		if versionFlag {
-			printVersion()
-		}
```
mukundansundar, line 45: I think adding return inside the block and having command print usage outside the block should solve this issue.

tianyax:
> The `--version` flag should be the same as `dapr version` IMO. It is meaningful to provide both CLI and installed runtime's version.

@shubham1172  For `--version` flag I think in most cases it should print its own information, and it overlaps with the `dapr version` function, I respect the original design and will make a change.😊

cmd/dapr.go
tianyax: 👍，Thanks

mukundansundar (approved):
LGTM

mukundansundar:
@tianyax Thanks for the contribution, welcome to the Dapr community.

Issue #1216, opened 2023-02-08
Support passing multiple `--resources-path` flags in `dapr run`


ItalyPaleAle:
## Describe the proposal

dapr/dapr#5909 is adding the ability to pass multiple `--resources-path` flags in daprd

## Release Note

RELEASE NOTE: **ADD** Support for multiple `--resources-path` flags

mukundansundar:
Once the runtime changes are made, the changes need to be made in CLI as well. This should be changed for the multi-app run template as well.

mukundansundar:
For E2E #904 needs to completed first.

mukundansundar:
This would potentially break multi-app run template as both the template and the run CLI flags share the same underlying struct. 

One potential solution is, continue to keep resourcesPath as string a string variable in the underlying struct. 
```go
ResourcesPath      string            `arg:"resources-path" yaml:"resourcesPath"`
```
So in `dapr run`, users would be able to pass in multiple `--resources-path` flags (`dapr run --resource-path path1 --resources-path path2 --resources-path path3`), but internally represented as `path1,path2,path3` in the struct var `ResourcesPath`. 

This would also mean that in the multi-app run template, the `resourcesPath` field would then accept a `,` delimited string of paths instead of a single path.  (Ideally this should be an array of strings and not a comma separated value of strings)

Or if it is ok to break the template compatibility since this is a preview feature, then we can change this struct value to handle an array of strings as expected. 

In either case, my suggestion for 1.12 will be to completely decouple `dapr run` underlying struct from the multi-app run template file underlying struct and have a translation layer in between. 

cc @ItalyPaleAle @philliphoff @yaron2

ItalyPaleAle:
Could you change the struct used by the multi-app template so `resourcesPath` accepts both a string and an array of strings? This may require changing the type of the property to `any` and then perform run-time type assertions to ensure it's either `string` or `[]string`.

mukundansundar:
> Could you change the struct used by the multi-app template so resourcesPath accepts both a string and an array of strings? This may require changing the type of the property to any and then perform run-time type assertions to ensure it's either string or []string.

Would prefer to have clear types in any struct fields that are going to be unmarshaled from a user input template.

> Would it be crazy to have the template support both the existing resourcesPath string property and a new resourcesPaths array property? The user can choose to use one or the other if there's only a single path, and preserves backward compatibility. (To me, it feels better as a collection should always use plural, even if just specifying one.)

[...]

PR #1276, opened 2023-04-25
Change for multiple resources paths

pravinpushkar:
# Description

1. changes for normal `dapr run`
example :- `dapr run --resources-path p1 --resources-path p2`

2. changes done for multi app run template

## Issue reference

Please reference the issue this PR will close: #1216 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pkg/standalone/runfileconfig/run_file_config_parser.go
```diff
@@ -56,6 +56,22 @@ func (a *RunFileConfig) validateRunConfig(runFilePath string) error {
...
+		for i := range a.Common.ResourcesPaths {
+			err := a.resolvePathToAbsAndValidate(baseDir, &a.Common.ResourcesPaths[i])
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// Merge common's section ResourcesPaths and ResourcePath. ResourcesPaths will be single source of truth for resources to be loaded.
+	if len(strings.TrimSpace(a.Common.ResourcesPath)) > 0 && !utils.Contains(a.Common.ResourcesPaths, a.Common.ResourcesPath) {
```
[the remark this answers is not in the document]
pravinpushkar: @philliphoff Thanks for pointing this out. As runtime does not enforces uniqueness. As in user can very well run this, So we can get rid of duplicate checking -

`daprd --app-id test --resources-path p1 --resources-path p1`. Both `resources-path` are pointing to same paths.
@mukundansundar

pkg/standalone/run.go
```diff
@@ -85,16 +85,17 @@ func (meta *DaprMeta) newAppID() string {
 }
 
 func (config *RunConfig) validateResourcesPath() error {
-	dirPath := config.ResourcesPath
-	if dirPath == "" {
-		dirPath = config.ComponentsPath
+	dirPath := config.ResourcesPaths
+	if len(dirPath) == 0 {
+		dirPath = []string{config.ComponentsPath}
```
shubham1172, line 90: Shall we add a info log here?

pkg/standalone/run.go
```diff
@@ -148,7 +149,7 @@ func (config *RunConfig) Validate() error {
 		return err
 	}
 
-	if config.ResourcesPath != "" {
+	if len(config.ResourcesPaths) > 0 {
 		config.ComponentsPath = ""
```
shubham1172, line 153: Should we add a info log here if ComponentsPath is not ""?

pkg/standalone/run.go
pravinpushkar: @shubham1172 Do you mean deprecation message ? We are already logging a deprecation warning message in cmd/run.go- 
```
RunCmd.Flags().MarkDeprecated("components-path", "This flag is deprecated and will be removed in the future releases. Use \"resources-path\" flag instead")
```
I think we can avoid logging it again here. WDYT ?

pkg/standalone/run.go
```diff
@@ -61,7 +60,8 @@ type SharedRunConfig struct {
 	MaxConcurrency     int               `arg:"app-max-concurrency" yaml:"appMaxConcurrency" default:"-1"`
 	PlacementHostAddr  string            `arg:"placement-host-address" yaml:"placementHostAddress"`
 	ComponentsPath     string            `arg:"components-path"`
```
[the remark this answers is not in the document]
pravinpushkar: in run template both `componentsPath` and `resourcesPath` is deprecated and will be replaced with `resourcePaths`.

pkg/standalone/run.go
shubham1172: If user provides both components path and resources path, it might be worth warning them again. But again, ok without it as well.

shubham1172 (approved):
LGTM!

PR #1283, opened 2023-05-17
add app channel address flag for run

mukundansundar:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -48,6 +48,9 @@ const (
 
 	windowsOsType = "windows"
 	homeDirPrefix = "~/"
+
+	// DefaultAppChannelAddress is the default local network address that user application listen on.
+	DefaultAppChannelAddress = "127.0.0.1"
 )
 
 // IsValidContainerRuntime checks if the input is a valid container runtime.
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -63,6 +63,7 @@ var (
 	enableAPILogging   bool
 	apiListenAddresses string
 	runFilePath        string
+	appChannelAddress  string
 )
 
 const (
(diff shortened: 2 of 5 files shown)


utils/utils.go
```diff
@@ -48,6 +48,9 @@ const (
 
 	windowsOsType = "windows"
 	homeDirPrefix = "~/"
+
+	// DefaultChannelAddress is the default local network address that user application listen on.
+	DefaultChannelAddress = "127.0.0.1"
```
yaron2, line 53: Not that this is likely to change, but can this be used from the dapr/dapr package? Its already defined there

mukundansundar: If possible I do not want to create another dependency in runtime. Can we move certain constants to kit and let both runtime and CLI depend on that instead?

yaron2: What's the issue with adding a package from `dapr/dapr`? In Go, as long as you don't have a package inside `internal`, it's perfectly fine to use packages. This ensures we have consistency and not minimize chances of non-synced changes. If anything, I would advocate for the CLI to rely on `dapr/dapr` packages as much as it possibly can.

yaron2:
Please also update the README file with an example.

mukundansundar:
> Please also update the README file with an example.

I am also writing an E2E for the same ... will do it

mukundansundar:
> > Please also update the README file with an example.
> 
> I am also writing an E2E for the same ... will do it

Added as part of usage help.

utils/utils.go
mukundansundar: Importing `github.com/dapr/pkg/runtime` into utils.go for using the constant defined there started creating issue wrt duplicate protobuf registration https://github.com/dapr/cli/actions/runs/5129434834/jobs/9227056977 associated commit https://github.com/dapr/cli/pull/1283/commits/0381466d446d6eca9eeda9bade0c23095046c168

for now hardcoding that constant to 127.0.0.1 directly in CLI itself.

pravinpushkar, line 53: Shall we call it this ?
```suggestion
	DefaultAppChannelAddress = "127.0.0.1"
```

mukundansundar: done

pravinpushkar (approved):
LGTM

Issue #1228, opened 2023-02-14
`dapr list` display the paths to the application and daprd logs in multi-app run


philliphoff:
## Describe the proposal

When using a run file (i.e. `dapr run -f <file>`), Dapr CLI redirects the application and Dapr logs of each application to its own file (under `<appDirPath>/.dapr/logs`).  Subsequent runs will create new log files to ensure that developers can go back to previous logs for analysis, with each log filename containing a "timestamp" of its first creation.  This is great but, after a while, these logs can accumulate and make it difficult for the the user to pick out the *right* log (e.g. the log that's currently being written to).  Tooling can help, by automating the process of finding the "latest" logs and opening those (e.g. in the current editor).  However, that automation (based on filenames) is not entirely certain; there will always be a chance some combination of files in the directory disrupts any filename timestamp sort.

Tooling is likely to care mostly about the logs being written to *right now*, and this is information that the Dapr CLI has. I propose that the Dapr CLI expose that information in the `dapr list` command.  This will provide a more deterministic means to obtain the paths to application logs (and avoid having tooling care about Dapr log naming conventions).

For example:

```json
[
  {
    "appId": "vscode-test",
    "httpPort": 64040,
    "grpcPort": 64041,
    "appPort": 3000,
    "metricsEnabled": true,
    "command": "node ./bin/www",
    "age": "7s",
    "created": "2023-01-30 10:32.27",
    "daprdPid": 18398,
    "cliPid": 18392,
    "commandPid": 18399,
    "maxRequestBodySize": -1,
    "httpReadBufferSize": -1,
    "runTemplatePath": "/Users/phoff/Source/scratch/vscode-test/dapr.yaml",
    # New path to the application log file
    "appLogPath": "/Users/phoff/Source/scratch/vscode-test/app-one/.dapr/logs/app-one_app_20230202160305.log", 
    # New path to the daprd log file
    "daprdLogPath": "/Users/phoff/Source/scratch/vscode-test/app-one/.dapr/logs/app-one_daprd_20230202160305.log" 
  }
]
```

## Release Note

RELEASE NOTE:

shivam-51:
Hi team, I am working on this!

mukundansundar:
> Hi team, I am working on this!

Please use /assign to assign this issue you yourself.

PR #1200, opened 2023-02-01
Store app PID in sidecar metadata.

philliphoff:
# Description

Adds the application's command's process ID to the Dapr sidecar metadata (alongside the sidecar and CLI process IDs).

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #1191

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation

PR #1296, opened 2023-05-28
shivam-51: Add log file paths to dapr list 1228

shivam-51:
# Description

Add paths of app and dapr logs to `dapr list` command.

## Issue reference

Please reference the issue this PR will close: #1228 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


mukundansundar:
@shivam-51 Can you please add E2E for this change?

pravinpushkar:
Multi app run template introduces `log destination` fields for daprd and app. We have to take into account if the log destination is set as `console` then what values it captures.

shivam-51:
> Multi app run template introduces `log destination` fields for daprd and app. We have to take into account if the log destination is set as `console` then what values it captures. @shivam-51

@pravinpushkar Can you share some documentation for this property? I am not able to understand it's significance.
I checked https://docs.dapr.io/developing-applications/local-development/multi-app-dapr-run/multi-app-template/

shivam-51:
> Multi app run template introduces `log destination` fields for daprd and app. We have to take into account if the log destination is set as `console` then what values it captures. @shivam-51

- For `appLogDestination: console` show `appLogPath: console`
- For `appLogDestination: file` show `appLogPath: <file path>`
- For `appLogDestination: fileAndConsole` show `appLogPath: <file path>` giving file more priority than console

@pravinpushkar @mukundansundar What do you think?

philliphoff:
@shivam-51 @pravinpushkar @mukundansundar My vote would be that, if the log is written to a file, then add/populate the `appLogDestination` metadata property.  If the log is not (e.g. being solely written to the console), then simply omit that metadata property (or set to `null` or the empty string `""`).  I would rather not have "special strings" that tooling needs to distinguish from otherwise valid paths.

pravinpushkar:
> @shivam-51 @pravinpushkar @mukundansundar My vote would be that, if the log is written to a file, then add/populate the `appLogDestination` metadata property. If the log is not (e.g. being solely written to the console), then simply omit that metadata property (or set to `null` or the empty string `""`). I would rather not have "special strings" that tooling needs to distinguish from otherwise valid paths.

+1 to update log paths only when written to only files. In other cases the logs will anyway be available to the console and tolling can use that. 

@philliphoff However one question -  if there are multiple apps writing logs to console then how will the tooling distinguish which log written by which app ?

philliphoff:
>However one question - if there are multiple apps writing logs to console then how will the tooling distinguish which log written by which app ?

@pravinpushkar If the application output is only going to the console, then it'd be unavailable to tooling anyway (unless it was the one to have started the Dapr process and captured the `stdout` itself).  If no path metadata is present, then tooling would simply assume it's being captured someplace else and not offer the user the option to view the logs.

shivam-51:
> Can you please add E2E for this change? Please see #1200 for reference.

@mukundansundar I have identified the tests need to go into `list_test.go` but I am not able to test locally. Most of the tests fail somehow.

mukundansundar:
> > Can you please add E2E for this change? Please see #1200 for reference.
> 
> @mukundansundar I have identified the tests need to go into `list_test.go` but I am not able to test locally. Most of the tests fail somehow.

Could you explain what errors that you are seeing?

cmd/run.go
```diff
@@ -540,6 +540,16 @@ func executeRun(runTemplateName, runFilePath string, apps []runfileconfig.App) (
 		// Update extended metadata with run file path.
 		putRunTemplateNameInMeta(runState, runTemplateName)
 
+		// Update extended metadata with app log file path.
+		if app.AppLogDestination != "console" {
```
mukundansundar, line 544: I think there is a constant for `console`, please use that.

mukundansundar (changes requested):

shivam-51:
> Could you explain what errors that you are seeing?

That's okay. I ran the tests on my fork using github actions.

cmd/run.go
shivam-51: Done.

mukundansundar (changes requested):
@shivam-51 Almost there, thanks for adding the negative scenarios. Can you add the positive test case as part of the run file scenarios where the list is actually populated with the file path? Inside this function here https://github.com/dapr/cli/blob/master/tests/e2e/standalone/stop_with_run_template_test.go#L121

shivam-51:
@mukundansundar can you have a look please?

mukundansundar (approved):

Issue #1300, opened 2023-06-03
dapr init --container-runtime podman


kochia7:
When there is no `podman` installed and try to run this command:
```bash
dapr init --container-runtime podman
```
Going to get this error:
```bash
❌  could not connect to Docker. Docker may not be installed or running
```

It should be:
```bash
❌  could not connect to Podman. Podman may not be installed or running
```

```bash
CLI version: 1.10.0 
Runtime version: 1.10.7
```

pravinpushkar:
@kochia7 in slim mode there is no need for a container runtime. It should have installed w/o any error. In fact you should use just this command - `dapr init --slim`.

See here for more info  - https://docs.dapr.io/operations/hosting/self-hosted/self-hosted-no-docker/

kochia7:
> @kochia7 in slim mode there is no need for a container runtime. It should have installed w/o any error. In fact you should use just this command - `dapr init --slim`.
> 
> See here for more info - https://docs.dapr.io/operations/hosting/self-hosted/self-hosted-no-docker/

Sorry for mistake, I just copied the last command and included the runtime parameter (fixed the post)

shubham1172:
Hi @kochia7, this has been fixed in the latest CLI version. Please check.

## With v1.10

```
~ ❯ dapr --version
CLI version: 1.10.0
Runtime version: n/a
```

Condition: docker is stopped and podman uninstalled.
```
~ ❯ dapr init --container-runtime podman
⌛  Making the jump to hyperspace...
❌  could not connect to Docker. Docker may not be installed or running
```

## With v1.11

```
~ ❯ dapr --version
CLI version: 1.11.0-rc.3
Runtime version: n/a
```

Condition: docker is stopped and podman uninstalled.
```
~ ❯ dapr init --container-runtime podman
⌛  Making the jump to hyperspace...
❌  could not connect to podman. podman may not be installed or running
```

pravinpushkar:
@shubham1172 yes, this makes sense. We can fix this for 1.12.

PR #1307, opened 2023-06-08
Fix checks for container runtime installation

shubham1172:
# Description

The previous logic to check if a container runtime is installed or not, used an OR logic. This resulted in issues like #1300

```go
containerRuntimeAvailable := utils.IsDockerInstalled() || utils.IsPodmanInstalled() 
```

The new fix is to check for a specific runtime based on the flag passed by the user.

## Issue reference

Please reference the issue this PR will close: #1300

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/pkg/standalone/standalone.go
+++ b/pkg/standalone/standalone.go
@@ -162,7 +162,7 @@ func Init(runtimeVersion, dashboardVersion string, dockerNetwork string, slimMod
 	setAirGapInit(fromDir)
 	if !slimMode {
 		// If --slim installation is not requested, check if docker is installed.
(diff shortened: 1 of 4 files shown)


pravinpushkar (approved):

mukundansundar:
> # Description
> The previous logic to check if a container runtime is installed or not, used an OR logic. This resulted in issues like #1300
> 
> ```go
> containerRuntimeAvailable := utils.IsDockerInstalled() || utils.IsPodmanInstalled() 
> ```
I did remember that a fix was made previously, so this is the issue :). Nice find.

Issue #1288, opened 2023-05-20
`dapr uninstall --all` should warn if container runtime is not running


shubham1172:
## Expected Behavior

When Docker (or any container runtime) is not running, `dapr uninstall --all` should throw a warning since placement, redis, and zipkin containers won't be removed.

## Actual Behavior

`dapr uninstall --all` exits with success status code and no warning.

## Steps to Reproduce the Problem

Ensure that docker is not running.

```
>> dapr uninstall --all
ℹ️  Removing Dapr from your machine...
ℹ️  Removing directory: /Users/shubham1172/.dapr/bin
ℹ️  Removing directory: /Users/shubham1172/.dapr
✅  Dapr has been removed successfully
```

## Release Note

RELEASE NOTE: **FIX** dapr uninstall --all to throw warning if Docker is not running

mcaupybugs:
Hi based on the description of the bug I can see that adding an `else if` condition and checking if `uninstallPlacementContainer` or `uninstallAll` variable is set to true and giving warning should be the solution. 
```
containerRuntimeAvailable := false
	containerRuntimeAvailable = utils.IsDockerInstalled() || utils.IsPodmanInstalled()
	if containerRuntimeAvailable {
		containerErrs = removeContainers(uninstallPlacementContainer, uninstallAll, dockerNetwork, runtimeCmd)
	}
```

If this solution works, Can I contribute to this? Or please let me know if i am missing something.

PR #1310, opened 2023-06-09
Adding warning for uninstall --all while container runtime not running

mcaupybugs:
# Description

Added an else if condition to check the case when the `dapr uninstall --all` command is used but the container runtime is not running. In such a case, this code change will throw a warning stating that it could not remove the supporting containers.

## Issue reference

Please reference the issue this PR will close: #1288

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


shubham1172 (commented):
Hi @mcaupybugs, this LGTM, could you also please add an e2e test? `dapr init --slim` followed by `dapr uninstall --all` should throw this error.

image

mcaupybugs:
Hey @shubham1172, I am not sure what E2E test should be written. Like I started to implement but after looking at the existing E2E tests I realized that each of those requires container runtime to be in the running state and is the first check done but my test case would require non-running container runtime. And if I try to manually give invalid container runtime using a --container-runtime flag then it will be blocked at the container-runtime validation step which is the `IsValidContainerRuntime` function and it checks if the runtime flag provided is either docker or podman. So not sure if i can even write E2E test for this scenario. please let me know wdyt

shubham1172:
Hi @mcaupybugs, you could stop the docker daemon using os/exec. Did you try that? Whilst doing it, we need to make sure that no other test is running in parallel. Although at this point I feel that it's non-trivial, so I am creating a separate issue to track it and approving this PR. Thanks again for the PR! :)

shubham1172 (approved):

PR #1339, opened 2023-09-14
Dapr CLI Docker Image

mindovermiles262:
# Description

This PR adds a workflow to publish Dapr CLI to docker hub when there is a new release. This PR will require help with a few things from Maintainers:

- [ ] A `daprio/cli` Docker hub repository created
- [ ] Secrets for `DOCKER_REGISTRY_ID` and
- [ ] Secret for `DOCKER_REGISTRY_PASSWORD` created.

## Issue reference

None

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


mindovermiles262:
I have a use case where I want to run a job on my kubernetes cluster using dapr cli.  Something like:

```
apiVersion: batch/v1
kind: Job
metadata:
  name: dapr-cli
spec:
  template:
    spec:
      containers:
      - name: dapr-cli
        image: daprio/cli
        command: ["dapr", "..."]
```

I did a quick search but couldn't find anything

shubham1172:
@mindovermiles262 how about a workaround like the following?

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: dapr-cli
spec:
  template:
    spec:
      containers:
      - name: dapr-container
        image: alpine:latest  # Maybe try distroless if not alpine, commands below are roughly written
        command:
          - "/bin/sh"
          - "-c"
          - |
            apk add --update --no-cache curl
            curl -fsSL https://raw.githubusercontent.com/dapr/cli/master/install/install.sh | /bin/sh
            dapr command
      restartPolicy: Never
  backoffLimit: 1

```

mindovermiles262:
Sure, that would work, but it seems like the cli should be available as a container

shubham1172:
> Sure, that would work, but it seems like the cli should be available as a container

If there are more use-cases or requirements from the community I'd be happy to vote in favor. For one specific use case it might be an overkill to maintain yet another GitHub pipeline and have yet another Docker image (which will support a single platform - linux/amd64 as per this PR).

mindovermiles262:
Fair enough. Let me move this to an Issue for future reference. Thanks

Issue #1416, opened 2024-07-04
Scheduler service is not started in self-hosted mode


mikeee:
## Expected Behavior

The scheduler should be started when using `dapr init`.

## Actual Behavior

```
C:\Users\#>dapr init --runtime-version 1.14.0-rc.1
Making the jump to hyperspace...
Container images will be pulled from Docker Hub
Installing runtime version 1.14.0-rc.1
Downloading binaries and setting up components...
Downloaded binaries and completed components set up.
daprd binary has been installed to C:\Users\#\.dapr\bin.
dapr_placement container is running.
dapr_redis container is running.
dapr_zipkin container is running.
Use `docker ps` to check running containers.
Success! Dapr is up and running. To get started, go here: https://aka.ms/dapr-getting-started

C:\Users\#>docker ps -a
CONTAINER ID   IMAGE                      COMMAND          CREATED         STATUS                     PORTS
                           NAMES
ee713c674604   kindest/node               ""               6 days ago      Exited (0) 292 years ago   80/tcp, 443/tcp, 127.0.0.1:52648->6443/tcp   kind-cluster-control-plane
03a0b3703797   daprio/dapr:1.14.0-rc.1    ""               9 minutes ago   Up 9 minutes               8080/tcp, 9090/tcp, 50005/tcp                dapr_placement
38979748ab7f   openzipkin/zipkin:latest   ""               9 minutes ago   Up 9 minutes               9411/tcp
                           dapr_zipkin
f82bcb0d8866   redis:6                    "redis-server"   9 minutes ago   Up 9 minutes               6379/tcp
                           dapr_redis

C:\Users\#>dapr uninstall --all
Removing Dapr from your machine...
Removing directory: C:\Users\#\.dapr\bin
Removing container: dapr_placement
WARNING: dapr_scheduler container does not exist
Removing container: dapr_redis
Removing container: dapr_zipkin
Removing directory: C:\Users\#\.dapr
Dapr has been removed successfully
```

## Steps to Reproduce the Problem

Mac/Windows - Podman

## Release Note

RELEASE NOTE:

mikeee:
```
dapr init --runtime-version 1.14.0-rc.2
⌛  Making the jump to hyperspace...
ℹ️  Container images will be pulled from Docker Hub
ℹ️  Installing runtime version 1.14.0-rc.2
❌  Downloading binaries and setting up components...
❌  Error: statfs /Users/#/.dapr/scheduler/data/0: no such file or directory
```

[...]

PR #1422, opened 2024-07-11
dapr_scheduler: Adds scheduler-volume flag

JoshVanL:
Expose `--scheduler-volume` to optionally specify the volume which scheduler writes to.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -45,6 +45,7 @@ var (
 	fromDir           string
 	containerRuntime  string
 	imageVariant      string
+	schedulerVolume   string
 )
 
 var InitCmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


mikeee (dismissed):
lgtm! Resolves the remaining scheduler init issue for me - #1416

yaron2 (approved):

cmd/init.go
```diff
@@ -219,6 +224,7 @@ func init() {
 	InitCmd.Flags().String("network", "", "The Docker network on which to deploy the Dapr runtime")
 	InitCmd.Flags().StringVarP(&fromDir, "from-dir", "", "", "Use Dapr artifacts from local directory for self-hosted installation")
 	InitCmd.Flags().StringVarP(&imageVariant, "image-variant", "", "", "The image variant to use for the Dapr runtime, for example: mariner")
+	InitCmd.Flags().StringVarP(&schedulerVolume, "scheduler-volume", "", "", "Self-hosted only. Optionally specify a volume for the scheduler service data directory. By default, scheduler data is not persisted and not resilient to restarts without this flag.")
```
artursouza, line 227: We should have a volume by default instead of not persisting stuff.

mikeee: https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/cmd/scheduler/options/options.go#L82

Agreed, unless the defaults are changed in both places. The mount should in its current state be created by default for the CLI 1.14.0-rc.4 release - in its current state a `dapr init` does not result in the scheduler running successfully.

marcduiker:
@holopin-bot @JoshVanL Thanks Josh!

PR #1430, opened 2024-07-19
fix: scheduler init data dir

mikeee:
# Description

Creates the directory to mount by default

## Issue reference

Please reference the issue this PR will close: #1416

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


yaron2:
cc @JoshVanL

mikeee:
Appears to work on mac but is having issues on Linux. Converting to draft

JoshVanL:
It's very rare that a user would want to create local directory which is mounted into scheduler. There is very little the user can do with the data as it's raw ETCD & proto. Container runtime local volumes are the canonical approach, which are still accessible via the filesystem, but have no user permission issues- we are even doing a 777 here which we never want to be doing.

PR #1437, opened 2024-07-23
Pin bitnami chart version for redis

### LINKED ISSUE #1436: dapr init -k --dev is in crash loop on redis init, failing redis component load
## Actual 
This quickstart repro's the issue.  
image

This appears to only happen in the GH action runner, and not on my local kind installations.
https://github.com/dapr/quickstarts/actions/runs/9702023201/job/26776912327#step:18:82

## Expected
It should look like this fully passing
image

## Release Note

RELEASE NOTE:

yaron2:
Pins the bitnami version to a stable one and removes unneeded Redis replicas for dev mode.

Fixes #1436 
and https://github.com/dapr/quickstarts/issues/1048

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -40,6 +40,7 @@ const (
 	daprReleaseName      = "dapr"
 	dashboardReleaseName = "dapr-dashboard"
 	latestVersion        = "latest"
+	bitnamiStableVersion = "17.14.5"
 
 	// dev mode constants.
 	thirdPartyDevNamespace   = "default"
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -40,6 +40,7 @@ const (
 	daprReleaseName      = "dapr"
 	dashboardReleaseName = "dapr-dashboard"
 	latestVersion        = "latest"
+	bitnamiStableVersion = "17.14.5"
```
paulyuk, line 43: Verified version

paulyuk (approved):
We'll just want to verify that the statestore CRD's creds from --dev work with this.

paulyuk:
> We'll just want to verify that the `statestore` CRD's creds from `dapr init -k --dev` work with this.

Actually i recommend CLI no longer applies statestore anymore.  This leads to too many troubleshooting conflicts.  The main value here is installing redis. not applying the statestore.

marcduiker:
@holopin-bot @yaron2 Thanks Yaron!

PR #1443, opened 2024-08-15
Remove Docker client dependency from standalone run

JoshVanL:
/


cicoyle (approved):
So that's why I couldn't repro the issue. Thanks Josh

yaron2 (approved):

PR #1445, opened 2024-08-15
Scheduler: change /var/lock to /run/lock

JoshVanL:
Change the data directory from `/var/lock` to `/run/lock` so that the scheduler data directory also works for mariner based images.


artursouza (approved):

artursouza:
Tests are failing.

PR #1478, opened 2025-01-22
Use non-deprecated flags in List operation

yaron2:
The CLI currently uses a deprecated flag in `daprd` for max request size and read buffer body. This PR sets the correct flags. The CLI can continue to exist with the user facing `--dapr` prefix.


mikeee (approved):

paulyuk (approved):
Matches supported migration.  
LGTM.  Thank you for this, because it makes our tests work.

yaron2:
@mikeee feel free to release rc.2

Issue #1503, opened 2025-03-13
Dapr ignore schedule-host-address


HenrixG:
## Expected Behavior

Ked zadam parameter --scheduler-host-address "empty" tak by sa sidecar nemala periodicky dotazovat na scheduler.
Druhy problem:
ked zadam --scheduler-host-address localhost:50007 tak by sa to malo pripojovat na IP 127.0.0.1 a nie na DockerImage IP.

## Actual Behavior

For a specific sideCar I can't turn off the scheduler (I don't use actors, jobs, workflow) even if I give the parameter
--scheduler-host-address "empty"
--scheduler-host-address ' '
it always says in the log that it's trying to connect to the scheduler.
second problem:
and when I write localhost:50006 it says in the log the IP of the dockerImage where the scheduler runs. so I can't connect to the sidecar that runs outside of docker.
## Steps to Reproduce the Problem

in VisualStudio use this launchSettings.json
    "Dapr": {
      "commandName": "Executable",
      "workingDirectory": "$(ProjectDir)",
      "executablePath": "dapr",
      "commandLineArgs": "run --app-id XXX --dapr-http-port 3500 --dapr-grpc-port 50001 --metrics-port 20000 --app-protocol https --placement-host-address localhost:50000 --scheduler-host-address ' '  --resources-path ../components/ --log-level debug --app-port 12000 --enable-api-logging -- dotnet run ."
    }

## Release Note
Dapr CLI 1.15.0
Windows 11
DockerDesktop 4.39

RELEASE NOTE:

I discussed this with @WhitWaldo(Innovian) on Discord.

mikeee:
The scheduler-host-address argument accepts "" however will be rewritten to localhost which is why it cannot be disabled at present when running in self-hosted mode unless you kick-off daprd without any scheduler-host-address(es)

https://github.com/dapr/cli/blob/release-1.15/cmd/run.go#L673:L681

antontroshin:
@HenrixG Following the Discord conversation and investigation, linking this here for documentation purposes. [Discord link](https://discord.com/channels/778680217417809931/1260511179894165515/1349827507771150508)

The first issue seems to be caused by a special case of Powershell parsing empty strings in a special way, which is different from the standard Command Prompt.
The solution to that would be to wrap the double quotes empty string with single quotes or use `=` to set the value in Powershell.
Example command:
```sh
dapr run --scheduler-host-address '""' --app-port 5000 -- go run ./app.go
# or
[...]

PR #1528, opened 2025-08-08
feat: ignore scheduler and placement connect

Gallardot:
# Description

## Issue reference

#1503 

https://docs.dapr.io/reference/arguments-annotations-overview/

Sometimes when developing locally, we don't need placement and scheduler.

When using cli, provide capabilities consistent with annotation.
```
dapr run --placement-host-address " " --scheduler-host-address " " -a app-id
```

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/pkg/standalone/run.go
+++ b/pkg/standalone/run.go
@@ -129,6 +129,11 @@ func (config *RunConfig) validatePlacementHostAddr() error {
 	if len(placementHostAddr) == 0 {
 		placementHostAddr = "localhost"
 	}
+
+	if strings.TrimSpace(placementHostAddr) == "" {
+		return nil
+	}
+
 	if indx := strings.Index(placementHostAddr, ":"); indx == -1 {
 		if runtime.GOOS == daprWindowsOS {
 			placementHostAddr += ":6050"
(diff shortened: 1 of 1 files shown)


pkg/standalone/run.go
```diff
@@ -129,6 +129,11 @@ func (config *RunConfig) validatePlacementHostAddr() error {
 	if len(placementHostAddr) == 0 {
 		placementHostAddr = "localhost"
 	}
+
+	if strings.TrimSpace(placementHostAddr) == "" {
```
yaron2, line 148: "localhost" is being assigned above in line 130 if placementHostAddr is empty, so this check could never be true

Gallardot: > "localhost" is being assigned above in line 130 if placementHostAddr is empty, so this check could never be true

Hi @yaron2 , Thanks for your reply.

Here is an example:

```dapr run --placement-host-address " " --scheduler-host-address " " -a app-id```

This is different from "empty". The part I modified is to handle such cases. I think it is OK.

When using the CLI, ensure that the provided capabilities are consistent with those of the annotations.

```
Comma separated list of addresses for Dapr Actor Placement servers.

When no annotation is set, the default value is set by the Sidecar Injector.

When the annotation is set and the value is a single space (' '), or “empty”, the sidecar does not connect to Placement server. This can be used when there are no actors running in the sidecar.

When the annotation is set and the value is not empty, the sidecar connects to the configured address. For example: 127.0.0.1:50057,127.0.0.1:50058
```
For reference, please see:
https://docs.dapr.io/reference/arguments-annotations-overview/

PR #1530, opened 2025-08-25
fix: allow explicit disable of placement with empty host

inishchith:
# Description

- This PR fixes a behavior where passing an explicitly empty placement host address should disable Placement, but instead the runtime still attempted to connect. It also aligns the CLI behavior and help so that users can reliably disable Placement without actors/workflows.

## Issue reference

Please reference the issue this PR will close: fixes: https://github.com/dapr/dapr/issues/9000

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -189,26 +189,42 @@ dapr run --run-file /path/to/directory -k
 		}
 
 		sharedRunConfig := &standalone.SharedRunConfig{
-			ConfigFile:           configFile,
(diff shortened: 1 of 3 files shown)


inishchith:
hi @JoshVanL - thanks for taking a look.

I have opened a PR (https://github.com/dapr/cli/pull/1535) as it'd be convenient with your context on this.

PR #1558, opened 2025-10-03
make it possible to use env vars in components

filintod:
# Description

This PR allows for substituting env var templated strings in components (resources) so a user can run configuration without having to change the component yaml file. Any templated value like `{{env_var_name}}` in the component file will be substituted with the env var value, if it exists, or leave as is if not.  You can also define default values in the shape of `{{env_var_name:default_value}}` where the colon `:` is the separator to indicate what is the default value if there is no env var with that name.

Note:  
- for multi-app runs there is no notion of using the `env` attribute to use for this, that is only pass to the app. The replacement does not use any of the `env` variables set in the multi-app run file.
- this solution might be replaced in the future by adding env var substitution in dapr/dapr when loading the component files content

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -210,7 +210,7 @@ dapr run --run-file /path/to/directory -k
 			SchedulerHostAddress: schedulerHostAddress,
 			DaprdInstallPath:     daprRuntimePath,
 		}
-		output, err := runExec.NewOutput(&standalone.RunConfig{
+		runConfig := &standalone.RunConfig{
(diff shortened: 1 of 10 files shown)


WhitWaldo:
Isn't this already possible via secret references (and use of the local envvar secret store)?

filintod:
oh, maybe and I just did not know, let me check this envvar secret store

PR #1564, opened 2025-11-06
Fix workflow resource path lookup

JoshVanL:
In standalone mode, if the resource path is not explicitly defined in the daprd process arguments, the workflow engine fails to locate the resource files, used to fetch the actor state store component manifest. This change updates the resource path lookup logic to use the default component path when the resource path is not defined.

### DIFF
--- a/.github/workflows/kind_e2e.yaml
+++ b/.github/workflows/kind_e2e.yaml
@@ -58,25 +58,25 @@ jobs:
       fail-fast: false # Keep running if one leg fails.
       matrix:
         k8s-version:
-        - v1.23.12
-        - v1.24.6
-        - v1.25.2
+        - v1.32.9
+        - v1.33.5
+        - v1.34.1
         mode:
         - ha
         - non-ha
         # Map between K8s and KinD versions.
         # This is attempting to make it a bit clearer what's being tested.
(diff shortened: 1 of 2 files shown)


cmd/runtime/runtime.go
```diff
@@ -0,0 +1,28 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package runtime
+
+import "github.com/spf13/cobra"
+
+var (
+	daprRuntimePath string
```
acroca, line 19: If I understand correctly, you moved this to its own package so it's reusable from both `pkg/workflow` and `cmd` packages without introducing a circular dependency, right?
Why not keep it as it was (in the `dapr.go`) and pass it as an argument when calling functions in the `pkg/workflow` like it's done for the other options like the kubernetes mode?

JoshVanL: `pkg/workflow` is called from the `cmd/workflow` sub command sub package commands. `cmd` imports `cmd/workflow` to register the workflow sub commands. 

We either need to have the flag registered and available from a third shared package, or register it multiple times for everywhere it is needed.

acroca: 🤦 right, I missed that sorry

PR #1580, opened 2026-01-29
fix(python apps): fork child + exec appcmd in child to avoid issues

sicoyle:
# Description

Fork is unsafe after threading has been used, at least for python. This is a known issue in the python community. See:
- https://www.python4data.science/en/24.3.0/performance/threading-forking-combined.html#Threading-and-forking-combined
- https://github.com/python/cpython/issues/80083
- https://github.com/python/cpython/issues/86263

Forking after threading has been used can cause deadlocks/hangs when using threading primitives like Lock, Condition, or Queue in a process created via fork. All of this occurs within the internals of durabletask-python. This means it will impact any app using the python sdk, be it dapr agents or general python workflows. I tried various python versions and they all experience hangs, at various points due to this.

This PR corrects this by using syscall.ForkExec to fork a child, then execs python or whatever app language within that child so the app language starts via exec, avoiding it knowing about the fork. I will say a downfall of this is I lose the blue APP prefix stuff on the log lines. So this means changing the UX via terminals unfortunately. I tried adding that back with pipes, but then adding the pipes to give me the log prefix then broke my python app again causing the hang.

This is what it looks like btw with this PR. Blue INFO is from the sidecar and the app logs are in white without the ===APP=== prefix that we used to have:

image

I need to test this on a windows machine to make sure I don't leave zombies, and also I think at this point this is good enough to fix other OS's. I could leave a note in the docs that running python apps on windows could lead to unpredictable behavior for now and create a follow up issue for that. Thoughts?

## Issue reference

-

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -771,7 +771,7 @@ func startDaprdAndAppProcesses(runConfig *standalone.RunConfig, commandDir strin
 		return runState, nil
 	}
 
-	if len(runConfig.Command) == 0 || strings.TrimSpace(runConfig.Command[0]) == "" {
+	if strings.TrimSpace(runConfig.Command[0]) == "" {
 		noCmdErr := errors.New("exec: no command")
 		print.StatusEvent(appErrorWriter, print.LogFailure, "Error starting app process: %s", noCmdErr.Error())
 		_ = killDaprdProcess(runState)
--- a/cmd/run_unix.go
+++ b/cmd/run_unix.go
@@ -41,6 +41,10 @@ func setDaprProcessGroupForRun(cmd *exec.Cmd) {
 // and sets output.AppCMD.Process.
 // It then runs a goroutine that waits and signals sigCh.
 func startAppProcessInBackground(output *runExec.RunOutput, binary string, args []string, env []string, sigCh chan os.Signal) error {
+	if output.AppCMD == nil || output.AppCMD.Process != nil {
+		return fmt.Errorf("app command is nil")
+	}
+
 	procAttr := &syscall.ProcAttr{
 		Env: env,
 		// stdin, stdout, and stderr inherit directly from the parent
@@ -59,9 +63,6 @@ func startAppProcessInBackground(output *runExec.RunOutput, binary string, args 
 	pid, err := syscall.ForkExec(binary, argv, procAttr)
 	if err != nil {
 		return fmt.Errorf("failed to fork/exec app: %w", err)
-	}
-	if output.AppCMD == nil {
-		return fmt.Errorf("app command is nil")
 	}
 	output.AppCMD.Process = &os.Process{Pid: pid}
 


pkg/standalone/run.go
```diff
@@ -45,6 +46,8 @@ const (
 
 	sentryDefaultAddress = "localhost:50001"
 	defaultStructTagKey  = "default"
+
+	windowsOsType osType = "windows"
```
acroca: Why? I don't see how this is better than using `daprWindowsOS`

sicoyle: this exact formatting is already used in a diff file elsewhere so i copy/pasted it here to be usable for me in this area. I did not pick this name. Happy to change, but it did exist this way so I kept for consistency🤷‍♀️

acroca: I think in this case you can use the existing variable `daprWindowsOS` like it was before. It's already in the `standalone.go` file and since you only need the string value `"windows"` it should work.

pkg/standalone/stop.go
```diff
@@ -57,13 +58,48 @@ func StopAppsWithRunFile(runTemplatePath string) error {
...
+				if err == syscall.ESRCH {
+					return nil
+				}
+				errKill := syscall.Kill(-pgid, syscall.SIGKILL)
+				// If process group doesn't exist, treat it as already stopped.
+				if errKill != nil && errKill != syscall.ESRCH {
+					return errKill
+				}
+			} else {
+				time.Sleep(1 * time.Second)
```
JoshVanL: Why do we sleep here?

pkg/standalone/stop.go
```diff
@@ -57,13 +58,48 @@ func StopAppsWithRunFile(runTemplatePath string) error {
...
+				if errCheck := syscall.Kill(-pgid, syscall.Signal(0)); errCheck == nil {
+					// Process group still exists, force kill it.
+					errKill := syscall.Kill(-pgid, syscall.SIGKILL)
+					// If process group doesn't exist, treat it as already stopped.
+					if errKill != nil && errKill != syscall.ESRCH {
+						return errKill
+					}
+				}
+				// Give additional time for port release after SIGKILL if needed
+				time.Sleep(500 * time.Millisecond)
```
JoshVanL: Same here- doesn't seem right to have arbitrary sleeps.

tests/e2e/standalone/uninstall_test.go
```diff
@@ -61,7 +61,7 @@ func verifyNoContainers(t *testing.T) {
 		return
 	}
 
-	cli, err := dockerClient.NewClientWithOpts(dockerClient.FromEnv)
+	cli, err := dockerClient.NewClientWithOpts(dockerClient.FromEnv, dockerClient.WithVersion("1.48"))
```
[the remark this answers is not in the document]
sicoyle: I added this as I was getting a compatibility issue due to versions so pinned on one that would work.

pkg/standalone/stop.go
sicoyle: was testing to see if it would make CI less flakey. I'll rm and see how it goes

pkg/standalone/stop.go
sicoyle: was testing to see if it would make CI less flakey. I'll rm and see how it goes

pkg/standalone/stop.go
```diff
@@ -57,13 +57,34 @@ func StopAppsWithRunFile(runTemplatePath string) error {
...
+				// If the process doesn't exist (ESRCH), treat it as already stopped.
+				if err == syscall.ESRCH {
+					return nil
+				}
 				// Fall back to cliPID if pgid is not available.
-				_, err = utils.RunCmdAndWait("kill", fmt.Sprintf("%v", a.CliPID)) //nolint:perfsprint
-				return err
+				_, err = utils.RunCmdAndWait("kill", "-TERM", fmt.Sprintf("%v", a.CliPID)) //nolint:perfsprint
+				if err != nil {
+					_, errKill := utils.RunCmdAndWait("kill", "-KILL", fmt.Sprintf("%v", a.CliPID)) //nolint:perfsprint
```
acroca: Can't we use `syscall.Kill` here as well?

pkg/standalone/run.go
```diff
@@ -619,9 +619,23 @@ func GetAppCommand(config *RunConfig) *exec.Cmd {
...
+		return cmd
+	}
+
+	// Use shell exec to avoid forking, which breaks Python threading on Unix
+	shArgs := []string{"-c", "exec \"$@\"", "sh", command}
+	shArgs = append(shArgs, args...)
+	cmd := exec.Command("/bin/sh", shArgs...)
 	cmd.Env = os.Environ()
 	cmd.Env = append(cmd.Env, config.getEnv()...)
+	setProcessGroup(cmd)
```
acroca, line 635: Do you think something like this would be easier to read and more maintainable?

```suggestion
	var cmd *exec.Cmd
	if runtime.GOOS == daprWindowsOS {
		// On Windows, run the executable directly (no shell).
		// TODO: In future this will likely need updates if Windows faces the same Python threading issues.
		cmd = exec.Command(command, args...)
	} else {
		shArgs := []string{"-c", "exec \"$@\"", "sh", command}
		shArgs = append(shArgs, args...)
		cmd = exec.Command("/bin/sh", shArgs...)
	}

	// Use shell exec to avoid forking, which breaks Python threading on Unix
	cmd.Env = os.Environ()
	cmd.Env = append(cmd.Env, config.getEnv()...)
	setProcessGroup(cmd)
```

PR #1598, opened 2026-02-27
use dapr v1.17.3

cicoyle:
use latest versions for official release


nelson-parente (dismissed):

.github/workflows/kind_e2e.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
     name: E2E tests for K8s (KinD)
     runs-on: ubuntu-latest
     env:
-      DAPR_RUNTIME_PINNED_VERSION: 1.17.0-rc.8
+      DAPR_RUNTIME_PINNED_VERSION: 1.17.0
```
acroca: Just a question, would we want to always run it against the latest runtime version?

JoshVanL (approved):

JoshVanL (approved):

PR #1600, opened 2026-03-01
Show command on failure

xverges:
# Description

- Make failures on subcommands more informative by displaying the command that triggered them and the exit code
- Prevent potential deadlock by using `exec.cmd.Output` rather than sequentially reading `stdout` and `stderr`

```
./dapr init --container-runtime podman
ℹ️  Container images will be pulled from Docker Hub
ℹ️  Installing runtime version 1.17.0
❌  Downloading binaries and setting up components...
❌  error running "podman run --name dapr_placement --restart always -d --entrypoint ./placement -p 50005:50005 -p 58080:8080 -p 59090:9090 docker.io/daprio/dapr:1.17.0" (exit code 126): Error: something went wrong with the request: "proxy already running\n"
```

## Issue reference

Please reference the issue this PR will close: #1594 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


utils/utils.go
```diff
@@ -192,40 +192,28 @@ func TruncateString(str string, maxLength int) string {
...
+	resp, err := cmd.Output()
 	if err != nil {
-		// in case of error, capture the exact message.
-		if len(errB) > 0 {
-			return "", errors.New(string(errB))
+		cmdStr := strings.Join(append([]string{name}, args...), " ")
+		output := ""
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			output = strings.TrimSpace(string(exitErr.Stderr))
```
acroca: What do you think about adding an early return? I think we can simplify the rest of the function to have less branches, something like this: 
```suggestion
			output = strings.TrimSpace(string(exitErr.Stderr))
			if output == "" {
				output = err.Error()
			}
			return "", fmt.Errorf("error running \"%s\" (exit code %d): %s", cmdStr, exitErr.ExitCode(), output)
```

xverges: Thanks! Much simpler, yes

utils/utils.go
```diff
@@ -192,40 +192,19 @@ func TruncateString(str string, maxLength int) string {
...
-		return "", err
-	}
-
-	err = cmd.Start()
-	if err != nil {
-		return "", err
-	}
-
-	resp, err := io.ReadAll(stdout)
+	resp, err := cmd.Output()
```
acroca, line 195: This changes a bit what ends up in the returned string because `resp` will now contain the `stderr` as well (if exit is 0), while previously we only had the `stdout` from the command if I'm not mistaken.
I think it's better but I'd like to hear from @dapr/maintainers-cli

xverges: I think that what you are describing is `CombinedOutput`. From https://pkg.go.dev/os/exec#Cmd.Output
> Output runs the command and returns its standard output. Any returned error will usually be of type [*ExitError](https://pkg.go.dev/os/exec#ExitError). If c.Stderr was nil and the returned error is of type [*ExitError](https://pkg.go.dev/os/exec#ExitError), Output populates the Stderr field of the returned error.

FWIW, I had the same understanding as you until I checked the docs a minute ago :-)

acroca: Ah cool, I was wrong then. In this case, I think the behavior is the same as previously.

acroca (approved):

PR #1626, opened 2026-04-09
fix(ci): trigger CLI build after creating release tag

nelson-parente:
## Summary

- Adds a `trigger` job to `create-release.yaml` that explicitly calls `gh workflow run dapr_cli.yaml` after the tag is created
- Same pattern already used in `dapr/dapr` ([see trigger job](https://github.com/dapr/dapr/blob/master/.github/workflows/create-release.yaml))


.github/workflows/create-release.yaml
```diff
@@ -46,4 +46,17 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: create-release
+    permissions:
+      actions: write
+    steps:
+      - name: Triggers the build
+        env:
+          GITHUB_TOKEN: ${{ secrets.DAPR_BOT_TOKEN }}
+        run: |
+          gh workflow run dapr_cli.yaml --repo dapr/cli --ref v$(echo '${{ inputs.rel_version }}' | sed -r 's/^[vV]?([0-9].+)$/\1/')
```
acroca: I think it'd be better to use the `workflow_call` approach instead.
When using `workflow_call` in the target workflow, we can invoke it with `uses: ./.github/workflows/dapr_cli.yml`, and it runs the workflow as part of the current workflow, so if the `dapr_cli` workflow fails, the release workflow will also show red.

Maybe we prefer to keep it consistent with `dapr/dapr`, but I think the `workflow_call` approach is better in this case.

JoshVanL (approved):

cicoyle (approved):

PR #1627, opened 2026-04-09
Add timer origin to the history information

acroca:
Timers have `origin` now. This PR displays the timer origin.


acroca:
Waiting for https://github.com/dapr/cli/pull/1629 to be merged

cicoyle (changes requested):
changes mostly lgtm but there are test failures still from this PR - please take a look and address it and copilots feedback:
```
        time="2026-05-12T16:21:38.006246853Z" level=info msg="Workflow Actor 'child-retry-origin-test:0002': workflow completed with status 'ORCHESTRATION_STATUS_FAILED' workflowName 'FailingChildWorkflow'" app_id=test-workflow instance=runnervmeorf1 scope=dapr.runtime.actors.targets.orchestrator type=log ver=1.18.0-rc.3

```

nelson-parente:
Hi @acroca — I pushed [d9317ad](https://github.com/dapr/cli/pull/1627/commits/d9317ad91de6cb214c9a1763c5f73bd597215d06) directly to this branch (`maintainerCanModify=true`) to address both halves of @cicoyle's review. Happy to revert if you'd rather drive this yourself.

### Root cause of the 4 failing subtests

The `TestWorkflowHistory/timer_origin_*` subtests were timing out with `Condition never satisfied`. After tracing the wire path end-to-end, the failure is on the **worker** side, not the CLI:

- `tests/apps/workflow/go.mod` still pinned `github.com/dapr/durabletask-go v0.10.0` (released 2025-09-10).
- `v0.10.0` predates the `CreateTimerAction.Origin` field — the worker's `createTimerInternal` builds a `CreateTimerAction` with **no Origin set**.
- The runtime's applier (`backend/runtimestate/applier.go`) does `switch o := timerAction.GetOrigin().(type) { … default: // stays nil }`, so it emits a `TimerCreatedEvent` with no Origin.
- The CLI then reads back a `TimerCreatedEvent` whose `Origin` oneof is `nil`, `timerOriginInfo` returns `nil`, no `origin=` attr is added, and the polling assertion never finds the substring.

So the CLI code in this PR is correct; the worker just wasn't producing the field.

### Fix

1. **`tests/apps/workflow/go.mod` + `go.sum`** — bump `durabletask-go` from `v0.10.0` → `v0.11.4-0.20260413145313-c4b7279b6a8e` (same commit `cli/go.mod` already uses, pre-rename so the API surface is still compatible with `go-sdk v1.13.0`). At this commit, the worker emits Origin for `createTimer`, `externalEvent`, `activityRetry`, and `childWorkflowRetry`.

2. **`tests/e2e/standalone/workflow_test.go`** — addresses Copilot's outstanding comment about long-running instances leaking. `WTimer` creates a 10h timer and `EventWorkflow` waits on an external event for 1h, so without cleanup they stay `RUNNING` in the state store and pollute later E2E runs. Each subtest now has a `t.Cleanup` that terminates + purges its own instance, and the instance ID is hoisted into a local `const` for readability.

### Verification done locally

- `go build ./...` for both `cli` and `tests/apps/workflow` — pass
- `go vet -tags=e2e ./tests/e2e/standalone/...` — clean
- `go test ./pkg/workflow/...` — 20 tests pass

E2E will run on CI for the real verification.

cicoyle (dismissed):

acroca:
@nelson-parente all good, thanks for doing it :)
I just resolved the merge conflict.

cicoyle (approved):

PR #1643, opened 2026-05-12
wf: faster list, IDs-only output, and quieter port-forward teardown

JoshVanL:
- workflow list: fetch per-instance metadata concurrently via errgroup (cap 32 in-flight) instead of serially, which was the bottleneck for apps with many instances.
- workflow list: add `-o ids` output, which skips the metadata fetch entirely and prints one instance ID per line. Errors out if combined with filters since those require metadata.
- portforward: filter out the noisy klog "Unhandled Error" lines emitted
  by client-go during teardown for benign "broken pipe", "connection reset by peer", and "use of closed network connection" cases, while letting all other errors fall through to the default handlers.


pkg/workflow/list.go
```diff
@@ -110,6 +118,31 @@ func ListShort(ctx context.Context, opts ListOptions) ([]*ListOutputShort, error
...
+		return nil, fmt.Errorf("failed to create Dapr client: %w", err)
+	}
+	defer dclient.Cancel()
+
+	ids, err := dclient.InstanceIDs(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	sort.Strings(ids)
```
acroca: Is it the default order by creation time? I think I'd keep it in whatever order we receive, the user can always sort if they need it sorted for any reason.

JoshVanL: Makes sense- I have updated :+1:

cicoyle (approved):

PR #1656, opened 2026-06-12
Use an unrestricted url to check latest dapr release

acroca:
# Description

The install script could fail with a confusing error when run from environments with shared IPs (e.g. GitHub Actions runners):

```
Downloading https://github.com/dapr/cli/releases/download//dapr_linux_amd64.tar.gz ...
tar: This does not look like a tar archive
```

**Root cause:** `getLatestRelease()` queries `api.github.com`, whose unauthenticated quota (60 req/hour/IP) is routinely exhausted on shared CI runner IPs. On a 403, `curl -s` exits 0 and the grep pipeline silently yields an empty version, so the script proceeded to build a malformed `download//` URL and failed much later in `tar`.

This PR makes two changes:

1. **Fetch the latest version from the releases Atom feed** (`https://github.com/dapr/cli/releases.atom`) instead of the REST API. The feed is served from GitHub's web tier and is not subject to the API rate limit. The selection logic is unchanged: newest release first, `rc` releases filtered out, first match wins.
2. **Fail fast if version detection still fails** (network error, throttling, unparseable response): the script now exits immediately with a clear message — `Failed to determine the latest Dapr CLI release. Retry later, or pass a specific version to this script as an argument.` — instead of attempting a download with an empty version.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -109,13 +109,15 @@ checkExistingDapr() {
 }
 
 getLatestRelease() {
-    local daprReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases"
+    # Use the releases Atom feed instead of the REST API, which is rate-limited for unauthenticated callers.
+    # Entry ids look like "tag:github.com,2008:Repository/192994336/v1.18.0", newest release first.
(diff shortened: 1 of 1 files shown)


CasperGN (approved):
Thank you @acroca - this should fix the issue I saw

javier-aliaga (approved):

JoshVanL (approved):

PR #1676, opened 2026-07-19
`dapr mcpservers`: skip cert-expiry check in self-hosted mode, validate `--output`

### LINKED ISSUE #1675: `dapr mcpservers`: cert-expiry check runs in self-hosted mode, `--output` unvalidated, `-n` shorthand clashes with other commands
## Expected Behavior

`dapr mcpservers` in self-hosted mode is a purely local operation (it scans a resources directory for MCPServer YAML), so it should not touch kubeconfig or the cluster. Its `--output` flag should reject unsupported values, consistent with `dapr list`.

## Actual Behavior

Three issues in the new `cmd/mcpservers.go` (checked at `96dd71a`):

1. **PostRun runs `kubernetes.CheckForCertExpiry()` unconditionally**, even in self-hosted mode ([cmd/mcpservers.go#L61-L63](https://github.com/dapr/cli/blob/96dd71a32f0e6648c29c554a3aeff50a38a070f8/cmd/mcpservers.go#L61-L63)). `dapr list` guards the same call with `if kubernetesMode` ([cmd/list.go#L105-L109](https://github.com/dapr/cli/blob/96dd71a32f0e6648c29c554a3aeff50a38a070f8/cmd/list.go#L105-L109)). A local listing command shouldn't be reaching for kubeconfig/cluster access.
2. **`--output` accepts any value without validation** — `dapr mcpservers -o table` silently behaves like the default instead of erroring, while `dapr list -o bogus` properly fails in PreRun ([cmd/list.go#L73-L78](https://github.com/dapr/cli/blob/96dd71a32f0e6648c29c554a3aeff50a38a070f8/cmd/list.go#L73-L78)).
3. **Flag-shorthand inconsistency worth a maintainer decision:** here `-n` means `--name` ([cmd/mcpservers.go#L90](https://github.com/dapr/cli/blob/96dd71a32f0e6648c29c554a3aeff50a38a070f8/cmd/mcpservers.go#L90)), while in `dapr init`/`uninstall`/`workflow`/`scheduler` `-n` means `--namespace`. So `dapr mcpservers -k -n default` filters by the *name* "default" and quietly returns nothing. Since changing a shorthand is breaking, I've left this out of the fix PR — flagging it while the command is still new and the blast radius is small.

1. Self-hosted machine, no cluster access: run `dapr mcpservers` — observe kubeconfig/cluster access from the PostRun cert-expiry check.
2. `dapr mcpservers -o table` — accepted silently instead of erroring.

I have a fix ready for 1 and 2 and will open a PR right after filing this.

## Release Note

RELEASE NOTE: **FIX** `dapr mcpservers` no longer performs Kubernetes cert-expiry checks in self-hosted mode and now validates its `--output` flag.

Mukuwul:
## Description

Two fixes for the new `dapr mcpservers` command:

- **Self-hosted runs no longer touch the cluster.** PostRun called `kubernetes.CheckForCertExpiry()` unconditionally, so a purely local resources-directory listing reached for kubeconfig/cluster access. It's now guarded behind `kubernetesMode`, matching what `dapr list` does.
- **`--output` is validated.** `dapr mcpservers -o table` used to be silently accepted and behave like the default; it now fails in PreRun with the valid values (json, yaml, list), matching `dapr list`'s behavior.

The `-n` shorthand clash described in the issue (name here vs namespace in `init`/`uninstall`/`workflow`/`scheduler`) is deliberately not touched - changing a shorthand is breaking, so that part needs a maintainer call first.

## Issue reference

Fixes #1675

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests (no e2e harness exercises mcpservers yet; happy to add coverage if wanted)
* [ ] Extended the documentation (n/a - behavior fix)

### DIFF
--- a/cmd/mcpservers.go
+++ b/cmd/mcpservers.go
@@ -35,6 +35,12 @@ var (
 var MCPServersCmd = &cobra.Command{
 	Use:   "mcpservers",
 	Short: "List all Dapr MCPServer resources. Supported platforms: Kubernetes and self-hosted",
+	PreRun: func(cmd *cobra.Command, args []string) {
+		if mcpServersOutputFormat != "list" && mcpServersOutputFormat != "json" && mcpServersOutputFormat != "yaml" {
+			print.FailureStatusEvent(os.Stderr, "An invalid output format was specified. Valid values are: json, yaml, or list")
+			os.Exit(1)
+		}
+	},
 	Run: func(cmd *cobra.Command, args []string) {
 		if kubernetesMode {
 			if allNamespaces || resourceNamespace == "" {
@@ -59,7 +65,9 @@ var MCPServersCmd = &cobra.Command{
 		}
 	},
 	PostRun: func(cmd *cobra.Command, args []string) {
-		kubernetes.CheckForCertExpiry()
+		if kubernetesMode {
+			kubernetes.CheckForCertExpiry()
+		}
 	},
 	Example: `
 # List all Dapr MCPServer resources in self-hosted mode (reads from ~/.dapr/components/ by default)


JoshVanL (approved):

Mukuwul:
The red check here is `darwin_amd64_complete` failing on `TestWorkflowChildCalls/fan_out_workflow` - the test-scheduler app times out waiting for actor in-flight lock claims during rebalancing. That's unrelated to this change (which only adds `--output` validation and guards the PostRun cert-expiry check behind `kubernetesMode`); it's the same flaky workflow test that's been going red intermittently on the macOS runner. A re-run should clear it.

marcduiker:
@holopin-bot @Mukuwul cli-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #1688, opened 2026-07-28
mtls: add --namespace to renew-certificate

MyMirelHub:
## Description

`dapr mtls renew-certificate` auto-detects the control-plane namespace from the first Dapr sentry pod found (`status[0].Namespace`) and runs a single Helm upgrade there, with no way to choose the namespace. If the control plane is installed in a namespace other than `dapr-system`, or more than one control plane is present in the cluster, the command can't be reliably pointed at the intended one.

This PR adds an optional `--namespace`/`-n` flag:

```
dapr mtls renew-certificate -k --namespace custom-namespace --valid-until 365 --restart
```

When set, certificate generation, the Helm upgrade, the control-plane restart, and the post-renewal expiry check all act on that namespace. When unset, behaviour is unchanged; if more than one control plane is detected, the command now errors and asks for `--namespace` rather than silently picking one.

## Changes

- Resolve the target namespace once in `RenewCertificate` and thread it through cert generation, `renewCertificate`, `restartControlPlaneService`, and the expiry check.
- Add namespace-aware helpers (`ExpiryInNamespace`, `getSystemConfigInNamespace`, `getTrustChainSecretInNamespace`, `GetDaprControlPlaneCurrentConfigInNamespace`) that default to the previous all-namespace behaviour when passed an empty namespace, so existing callers are unaffected.

## Testing

- New unit tests in `pkg/kubernetes/renew_certificate_test.go` (explicit-namespace path, none/single/multiple detection, namespace enumeration via fake clientset).
- `go build ./...`, `go test ./pkg/kubernetes/...`, and `gofmt` pass.

## Notes

- `pkg/kubernetes/upgrade.go` uses the same auto-detect pattern, so `dapr upgrade` shares this limitation — left out here to keep the change focused; happy to follow up if preferred.

## Release Note
RELEASE NOTE: ADD `--namespace` flag to `dapr mtls renew-certificate`.


MyMirelHub:
Closing as not planned.

After testing against a real multi-control-plane install, the `--namespace` flag works for targeting, but it's only part of the problem: making `renew-certificate` complete end-to-end across non-standard / OCI-hosted or air-gapped distributions turns out to need a larger redesign of how the chart is sourced during the helm upgrade (the command re-pulls the chart from the public Dapr Helm repo, and reusing the release's stored chart is unsafe for umbrella charts with subcharts). Given a documented manual per-namespace workaround exists (per-namespace `helm upgrade` with the new cert PEMs, or clearing the trust bundle to regenerate self-signed certs), this isn't worth the added complexity right now.

PR #1692, opened 2026-08-24
Placement -> Scheduler CLI Changes

cicoyle:
Placement -> Scheduler CLI Changes

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -47,6 +47,7 @@ var (
 	imageVariant                       string
 	schedulerVolume                    string
 	schedulerOverrideBroadcastHostPort string
+	schedulerPlacement                 bool
 	redisStack                         bool
 )
 
@@ -195,6 +196,7 @@ dapr init --redis-stack
 				DaprInstallPath:                    runtime.GetDaprRuntimePath(),
 				SchedulerVolume:                    &schedulerVolume,
 				SchedulerOverrideBroadcastHostPort: schedulerHostPort,
+				SchedulerPlacement:                 schedulerPlacement,
 				RedisStack:                         redisStack,
 			})
 			if err != nil {
@@ -245,6 +247,7 @@ func init() {
 	InitCmd.Flags().StringVarP(&imageVariant, "image-variant", "", "", "The image variant to use for the Dapr runtime, for example: mariner")
 	InitCmd.Flags().StringVarP(&schedulerVolume, "scheduler-volume", "", "dapr_scheduler", "Self-hosted only. Specify a volume for the scheduler service data directory.")
 	InitCmd.Flags().StringVarP(&schedulerOverrideBroadcastHostPort, "scheduler-override-broadcast-host-port", "", "", "Self-hosted only. Specify the scheduler broadcast host and port, for example: 192.168.42.42:50006. If not specified, it uses localhost:50006 (6060 for Windows).")
+	InitCmd.Flags().BoolVarP(&schedulerPlacement, "scheduler-placement", "", false, "Self-hosted only. Serve actor placement from the scheduler service instead of running the placement service. Requires Dapr 1.19 or later.")
 	InitCmd.Flags().BoolVarP(&redisStack, "redis-stack", "", false, "Self-hosted only. Use redis-stack-server image instead of standard Redis for RediSearch support")
 	InitCmd.Flags().BoolP("help", "h", false, "Print this help message")
 	InitCmd.Flags().StringArrayVar(&values, "set", []string{}, "set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)")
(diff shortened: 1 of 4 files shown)


nelson-parente (dismissed):

pkg/standalone/standalone.go
```diff
@@ -866,7 +922,7 @@ func installDaprRuntime(wg *sync.WaitGroup, errorChan chan<- error, info initInf
 func installPlacement(wg *sync.WaitGroup, errorChan chan<- error, info initInfo) {
 	defer wg.Done()
 
-	if !info.slimMode {
+	if !info.slimMode || info.schedulerPlacement {
```
JoshVanL, line 925: `dapr init --slim --scheduler-placement` finishes with `Success! Dapr is up and running.` but leaves no actor-placement authority installed or running.

Two early returns compose badly:

- Here, `installPlacement` returns before installing the binary whenever `info.schedulerPlacement` is set, including in slim mode.
- `runSchedulerService` returns at `pkg/standalone/standalone.go:682` when `info.slimMode`, so slim mode never starts a scheduler either.

So the user gets no placement binary, and no scheduler process that could serve placement instead. They are expected to launch the scheduler themselves in slim mode, and the runtime default is `--placement-enabled=false`, so even then placement is off unless they happen to pass the flag.

Nothing rejects the combination: `schedulerPlacement` is read at `cmd/init.go:199` and never validated against `slimMode`.

Two ways out, either is fine:

```go
// keep the binary in slim mode, since nothing else can serve placement there
if !info.slimMode {
	return
}
```

or reject the combination up front in `cmd/init.go`, the way `--redis-stack` does at `cmd/init.go:119`:

```go
if slimMode && schedulerPlacement {
	print.FailureStatusEvent(os.Stderr, "--scheduler-placement cannot be used with --slim: slim mode does not run the scheduler")
	os.Exit(1)
}
```

tests/e2e/standalone/init_test.go
```diff
@@ -223,6 +223,43 @@ func TestStandaloneInit(t *testing.T) {
 		verifyTCPLocalhost(t, schedulerPort)
 	})
 
+	t.Run("init with scheduler placement", func(t *testing.T) {
```
JoshVanL, line 226: This subtest needs an `isSlimMode()` skip. It is the only container-dependent subtest in the file without one, and the very next subtest at line 264 has exactly the guard it is missing:

```go
	t.Run("init with scheduler placement", func(t *testing.T) {
		if isSlimMode() {
			t.Skip("Skipping scheduler placement test because of slim installation")
		}
```

On the slim CI legs (`DAPR_E2E_INIT_SLIM` is true for `windows-latest` and the `slim` matrix entry in `.github/workflows/self_hosted_e2e.yaml`) it fails twice over:

- `verifyBinaries` requires `binaries["placement"]` when `isSlimMode()` (`init_test.go:395-397`), but `installPlacement` returns early for `info.schedulerPlacement` (`pkg/standalone/standalone.go:925`), so that binary is never installed.
- `verifySchedulerPlacementContainers(t)` looks for containers, and slim mode runs none.

Right now this is dormant only because the `>= 1.19.x` gate skips the subtest everywhere. It turns red as soon as `DAPR_RUNTIME_LATEST_STABLE_VERSION` reaches 1.19, which is easy to miss because it will look like an unrelated release-bump failure.

Worth noting this is the same bug as the `--slim --scheduler-placement` one above, just seen from the test side. If you fix that by installing the placement binary in slim mode, the `verifyBinaries` half goes away on its own, but the container half still needs the skip.

cmd/init.go
```diff
@@ -245,6 +247,7 @@ func init() {
 	InitCmd.Flags().StringVarP(&imageVariant, "image-variant", "", "", "The image variant to use for the Dapr runtime, for example: mariner")
 	InitCmd.Flags().StringVarP(&schedulerVolume, "scheduler-volume", "", "dapr_scheduler", "Self-hosted only. Specify a volume for the scheduler service data directory.")
 	InitCmd.Flags().StringVarP(&schedulerOverrideBroadcastHostPort, "scheduler-override-broadcast-host-port", "", "", "Self-hosted only. Specify the scheduler broadcast host and port, for example: 192.168.42.42:50006. If not specified, it uses localhost:50006 (6060 for Windows).")
+	InitCmd.Flags().BoolVarP(&schedulerPlacement, "scheduler-placement", "", false, "Self-hosted only. Serve actor placement from the scheduler service instead of running the placement service. Requires Dapr 1.19 or later.")
```
JoshVanL, line 258: The help text says "Self-hosted only", but `dapr init -k --scheduler-placement` does not say so: the flag is read into `standalone.Config` at `cmd/init.go:199` and never reaches `kubernetes.InitConfiguration`, so the user gets an ordinary Kubernetes install with a standalone placement statefulset and no indication the flag did nothing.

`--redis-stack` handles the mirror-image case explicitly four lines above the Kubernetes branch:

```go
if redisStack {
	print.FailureStatusEvent(os.Stderr, "--redis-stack is only valid for self-hosted mode")
	os.Exit(1)
}
```

I know `--scheduler-volume` and `--scheduler-override-broadcast-host-port` already behave this way, so this is a pre-existing convention rather than something you introduced. Raising it because this flag changes which control-plane component owns placement, so silently ignoring it is a bigger surprise than ignoring a volume path.

JoshVanL (changes requested):

JoshVanL (approved):

